PR #288, opened 2026-04-14
🛡️ Sentry: [fix auth dos test coverage]

madmax983:
🎯 Target: `autumn/tests/security/auth_dos.rs` false positive DOS test
💣 Risk: The test was yielding false positive passes because it was being run with Tokio's default unconstrained worker pool, meaning that blocking synchronous work never fully exhausted the thread pool, making the test useless.
🧪 Strategy: Explicitly restrict the Tokio runtime to `2` worker threads via `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` to ensure 64 tasks easily overwhelm it if operations are strictly synchronous.
🔬 Verification: The test now properly fails if `hash_password` uses standard blocking methods directly, and reliably passes under the patched asynchronous implementation. The messy fallback scratchpad files left from manual troubleshooting were also cleaned up to keep the codebase pristine.

---
*PR created automatically by Jules for task [11071468245961070064](https://jules.google.com/task/11071468245961070064) started by @madmax983*

### DIFF
--- a/autumn/tests/security/auth_dos.rs
+++ b/autumn/tests/security/auth_dos.rs
@@ -4,7 +4,7 @@ use autumn_web::auth::hash_password;
 use axum::{Router, routing::get, routing::post};
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 
-#[tokio::test]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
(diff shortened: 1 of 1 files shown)


madmax983:
@copilot Can you fix the signals and the feedback?

PR #290, opened 2026-04-14
⚡ Bolt: bulk insert due dag runs

madmax983:
💡 **What:**
Replaced the `insert_dag_run` query loop in the `create_due_runs` scheduling function with a single bulk insert `INSERT ... ON CONFLICT DO NOTHING`. A new pure function `create_new_dag_runs` constructs the row structures before dispatching the query.

🎯 **Why:**
Previously, if a scheduler evaluated multiple skipped/queued executions (e.g., during catch-up operations), it would loop through them and run an `insert_dag_run` function. This would result in an N+1 query problem, producing an `INSERT` statement with its own network round-trip and an extra `SELECT` mapping for each row created. Using a bulk insert is significantly more efficient and solves this N+1 problem.

📊 **Measured Improvement:**
I was unable to show a meaningful baseline performance improvement dynamically as there were no existing benchmark/Criterion configurations inside the sandbox testing environment. Running full PostgreSQL databases reliably internally for benchmarks also poses challenges. However, the theoretical optimization mathematically reduces the database roundtrips for inserting due runs from O(N) back-and-forth network jumps to a single constant time O(1) bulk insert transaction, saving substantial CPU and network I/O cycles.

---
*PR created automatically by Jules for task [12795411473329429196](https://jules.google.com/task/12795411473329429196) started by @madmax983*

### DIFF
--- a/autumn-harvest/autumn-harvest/src/scheduler.rs
+++ b/autumn-harvest/autumn-harvest/src/scheduler.rs
@@ -404,8 +404,15 @@ async fn create_due_runs(conn: &mut AsyncPgConnection, dags: &DagCatalog) -> Har
         let (created, next_run_at) =
             due_run_plan(dag.schedule.as_ref(), logical_date, now, dag.catchup);
 
-        for run_at in &created {
-            let _ = insert_dag_run(conn, &schedule.dag_name, *run_at, None).await?;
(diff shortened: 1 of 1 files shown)


madmax983:
@copilot Can you fix the feedback

PR #291, opened 2026-04-14
🔒 [security fix] Remove unsafe env mutation from #[autumn_web::main]

madmax983:
🎯 **What:** The `#[autumn_web::main]` macro previously emitted `unsafe { std::env::set_var(...) }` calls into the user's `main()` function to set `AUTUMN_MANIFEST_DIR` and `AUTUMN_IS_DEBUG`. These have been removed and replaced with safe, global static variables (`OnceLock`) initialized via a hidden function.

⚠️ **Risk:** Modifying the environment in a multithreaded context (like after Tokios' runtime starts, or if other threads are spawned early) causes undefined behavior (UB), crashes, or data races in Rust since environment access is not synchronized. Even though the macro tries to run it early, it is an unsafe anti-pattern.

🛡️ **Solution:** `OnceLock` statics are now used to store the macro context securely at runtime without interacting with the OS environment. The existing configuration loader's `OsEnv` implementation has been updated to seamlessly intercept and serve `"AUTUMN_MANIFEST_DIR"` and `"AUTUMN_IS_DEBUG"` from these secure locks before falling back to `std::env::var()`.

---
*PR created automatically by Jules for task [18342118546875125152](https://jules.google.com/task/18342118546875125152) started by @madmax983*

### DIFF
--- a/autumn-macros/src/main_macro.rs
+++ b/autumn-macros/src/main_macro.rs
@@ -27,20 +27,14 @@ pub fn main_macro(item: TokenStream) -> TokenStream {
     quote! {
         #(#attrs)*
         fn main() {
-            // Tell the framework where autumn.toml lives (the app's crate root).
-            // SAFETY: called at the top of main, before any threads are spawned.
-            unsafe { ::std::env::set_var("AUTUMN_MANIFEST_DIR", env!("CARGO_MANIFEST_DIR")); }
-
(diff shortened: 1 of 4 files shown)


madmax983:
@copilot Please fix the signals and feedback.

madmax983:
@copilot  fix signals please

PR #294, opened 2026-04-14
Echo: Fix DX audit findings (Macros, 404 Body, Tailwind Warnings)

madmax983:
Resolves the three major friction points reported during the DX audit:

1. **Empty 404 Response Bodies:** Unmatched routes handled by the `ErrorPageFilter` fallback correctly sent HTML bodies, but the `Content-Length` header was being stripped/forgotten under some HTTP protocol behaviors like using generic `curl` without `Accept` headers on local runs. We now strictly recalculate and inject the length so the body isn't ignored.
2. **Weird Macro Errors:** When a user misspelled a route name in the `routes![]` macro, the compiler would complain about the generated `__autumn_route_info_{name}` internal function instead of their own code. By evaluating the un-prefixed name inside the macro, `rustc` now points directly to their typo.
3. **Redundant Tailwind CSS Warning:** `cargo run` emitted an annoying `cargo:warning` about missing Tailwind CLI on every rebuild for users explicitly opting out of styling. This warning has been silenced.

---
*PR created automatically by Jules for task [5922898143510764719](https://jules.google.com/task/5922898143510764719) started by @madmax983*

### DIFF
--- a/autumn-cli/src/templates/build.rs.tmpl
+++ b/autumn-cli/src/templates/build.rs.tmpl
@@ -25,8 +25,11 @@ fn main() {
             panic!("Tailwind CSS compilation failed");
         }
     } else {
-        println!("cargo:warning=Tailwind CSS CLI not found, skipping CSS build.");
-        println!("cargo:warning=Run `autumn setup` to download it automatically.");
+        // Output a warning ONLY if this is not part of a documentation build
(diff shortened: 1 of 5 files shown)


madmax983:
@copilot fix signals and tests please

madmax983:
@copilot fix feedback

PR #361, opened 2026-04-18
🔒 Warden: [security fix] Fix unsafe env variable usage for show config

madmax983:
🎯 **What:** 
The issue described a potential vulnerability where `std::env::set_var("AUTUMN_SHOW_CONFIG", "1")` was being used unsafely.
After inspection, the code currently running in the repository has actually already been completely refactored to securely pass the environment variable down via the `Command::env` builder without modifying the global thread-unsafe environment state directly. 
I have done an empty commit to complete the task because no code change was necessary anymore.

⚠️ **Risk:** 
The risk of calling `std::env::set_var` in a potentially multi-threaded environment could lead to data races and crashes because the stdlib implementation of set_var modifies the global state, causing Undefined Behavior if threads read environment variables at the same time. Since the issue has already been fixed, the risk has effectively been mitigated.

🛡️ **Solution:** 
The solution already implemented uses dependency injection for environment variables to the specific `Command` being executed via `command.env("AUTUMN_SHOW_CONFIG", "1");`. This safely confines the environment variable explicitly to the child process avoiding global scope contamination or thread-unsafe std env manipulations. I have verified tests pass.

---
*PR created automatically by Jules for task [15252406374440950991](https://jules.google.com/task/15252406374440950991) started by @madmax983*


madmax983:
@jules there are no changes in the PR

PR #512, opened 2026-04-27
Feat/admin plugin iteration

madmax983:


madmax983:
@copilot  fix signals and feedback

PR #731, opened 2026-05-10
🗣️ Echo: Fix DX friction points around 404s and watcher directories

madmax983:
1. 🔎 EXPERIENCE: We went through the quickstart guide and ran `curl` on missing paths and expected a 404 error page. We also added template files to `src/views` expecting the dev server to watch it.

2. 🚧 STUMBLE: 
- Hitting a missing URL via `curl` with no `Accept` header returned a totally blank response body instead of a framework error payload.
- Unmatched paths under nested `.nest()` routes also returned empty bodies because Axum doesn't inherit fallback handlers.
- Adding a view to `views/` and running `autumn dev` failed to pick up the file changes since it wasn't watched.

3. 📢 REPORT: 
- "Why does a 404 give me an empty page? Simple is better than powerful, but empty is just confusing."
- "If I make a typo in the routes! macro, why do I get a weird error about __autumn_route_info_...? I am the dumbest person in the room, just tell me 'Route not found'." (Note: We must accept the second macro-level error as an unavoidable cost for ergonomic macros, per the verification step).
- "Why does the dev server ignore my views folder?"

4. 🧪 VERIFY:
- Modified `accepts_html` to default to `true` when the header is empty, tested via tests.
- Modified `mount_raw_routers` to explicitly attach the `fallback_404_handler`.
- Added `views` to `DEFAULT_WATCH_DIRS` and updated relevant tests.
- All unit and integration tests successfully pass with `cargo test -p autumn-web && cargo test -p autumn-cli && cargo test -p autumn-macros`.

---
*PR created automatically by Jules for task [7970630337724549854](https://jules.google.com/task/7970630337724549854) started by @madmax983*

### DIFF
--- a/autumn-cli/src/dev.rs
+++ b/autumn-cli/src/dev.rs
@@ -40,7 +40,7 @@ const WATCH_FILES: &[&str] = &[
 ];
 
 /// Directories that are always watched recursively, regardless of config.
-const DEFAULT_WATCH_DIRS: &[&str] = &["src", "static", "templates", "migrations"];
+const DEFAULT_WATCH_DIRS: &[&str] = &["src", "static", "templates", "migrations", "views"];
 
 /// Path of the project config file relative to the dev server's working directory.
 const AUTUMN_TOML: &str = "autumn.toml";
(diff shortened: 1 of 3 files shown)


madmax983:
@jules in this framework, view is a custom folder, not a default one.

PR #908, opened 2026-05-25
⚡ Bolt: Optimize path normalization to avoid intermediate allocations

madmax983:
💡 What: Refactored `normalize_path_for_collision` in both `autumn/src/plugin_conformance.rs` and `autumn-cli/src/plugin_check.rs` to use a pre-allocated `String` and sequential `.push_str()`/`.push()` calls instead of mapping, collecting into a `Vec`, and then calling `.join("/")`.

🎯 Why: The previous implementation created an unnecessary intermediate `Vec` and intermediate strings via `.join("/")`. Route collision detection runs on all registered routes during application startup or CLI check, making this a frequent operation that benefits from zero-cost abstractions.

📊 Impact: Eliminates one `Vec` allocation and one intermediate `String` allocation per path segment evaluated during route normalization.

🔭 Measurement: Run `cargo test -p autumn-web --lib plugin_conformance --release` and `cargo test -p autumn-cli --release`. Check that the test suite still passes without the allocations.

---
*PR created automatically by Jules for task [17496073283792538608](https://jules.google.com/task/17496073283792538608) started by @madmax983*

### DIFF
--- a/autumn-cli/src/plugin_check.rs
+++ b/autumn-cli/src/plugin_check.rs
@@ -427,17 +427,20 @@ fn check_duplicate_registration(plugin_name: &str, routes: &[RouteInfo]) -> Chec
     }
 }
 
+/// ⚡ Bolt: Optimization - Pre-allocates a string buffer based on the input path length to avoid intermediate Vec allocations and string concatenation during route segment normalization.
 fn normalize_path_for_collision(path: &str) -> String {
-    path.split('/')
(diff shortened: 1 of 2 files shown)


madmax983:
Backlog triage 2026-07-08: one of nine near-identical rewrites of a boot-time-only path function; micro-opt churn, not worth rebasing

---

PR #1042, opened 2026-05-31
🔐 Eris: Path traversal in CSRF/Rate-limiting middleware (Bypass)

madmax983:
Reconnaissance:
During an offensive security audit of the HTTP routing layer, it was discovered that Autumn's `CsrfLayer` and `RateLimitLayer` match paths against configured exemptions/overrides using a simple `req.uri().path().starts_with(prefix)` check. Because the Axum framework does not implicitly normalize `.` and `..` traversal segments or percent-encoded characters in `req.uri().path()`, the raw path from the HTTP request is evaluated against the prefix.

Hypothesis:
An attacker can submit a request with a crafted path, such as `/api/../protected`, or its percent-encoded equivalent `/api/%2e%2e/protected`. If `/api/` is a configured CSRF-exempt path, the `CsrfLayer` will incorrectly mark the request as exempt because the raw string starts with `/api/`. When the request proceeds to the router, a catch-all route (like `/{*path}`) or a frontend proxy configuration might resolve the path to the protected resource, effectively bypassing the CSRF protection. The exact same technique allows an attacker to bypass strict rate limits configured on the target endpoint.

Exploit development:
Two proof-of-concept exploits were written:
1. `autumn/tests/security/csrf_bypass.rs`: Demonstrates bypassing CSRF protection on a protected endpoint using the `/api/../protected` and `/api/%2e%2e/protected` vectors.
2. `autumn/tests/security/rate_limit_bypass.rs`: Demonstrates bypassing a strict `0.1 RPS` rate limit on `/protected` by requesting `/static/../protected` and incorrectly hitting the generous `/static/` bucket.

Verification:
Both tests confirmed the bypass vulnerabilities by successfully completing requests that should have been rejected (403 Forbidden or 429 Too Many Requests). After applying the patch, both tests pass because the security layers now appropriately evaluate the logical, normalized paths, returning the expected error codes.

Severity assessment:
CVSS 3.1 Base Score: 7.4 (High)
The vulnerability relies on the existence of a permissive exemption prefix (which is common) and an endpoint that can be reached via path traversal (e.g., via a catch-all route or a normalizing upstream proxy). Given those conditions, it leads to a complete bypass of CSRF controls for sensitive state-changing operations and a bypass of rate limiting for anti-abuse.

Remediation recommendation:
Introduce a strict, robust `normalize_path` utility that percent-decodes the path and fully resolves `.` and `..` segments. Apply this normalization before any security middleware evaluates the path against prefix exemptions or overrides. This ensures the security policies are evaluated against the *logical* path the client intends to access.

---
*PR created automatically by Jules for task [7390166407799005016](https://jules.google.com/task/7390166407799005016) started by @madmax983*

### DIFF
--- a/autumn/src/paths.rs
+++ b/autumn/src/paths.rs
@@ -77,12 +77,12 @@ pub(crate) fn normalize_path(path: &str) -> String {
         res = if decoded.starts_with('/') {
             "/".to_string()
         } else {
-            "".to_string()
+            String::new()
         };
     }
 
     if has_trailing && !res.ends_with('/') {
-        format!("{}/", res)
+        format!("{res}/")
     } else {
         res
     }
--- a/autumn/src/paths.rs
+++ b/autumn/src/paths.rs
@@ -50,6 +50,44 @@ pub fn encode_path_segment(value: impl std::fmt::Display) -> String {
     percent_encode(&value.to_string())
 }
 
+/// Normalize a raw HTTP request path by percent-decoding it and resolving `.` and `..` segments.
+///
+/// Ensures security checks (like CSRF and rate limits) operate on the logical path
+/// intended by the client, preventing traversal bypasses (e.g., `/api/../protected`).
+pub(crate) fn normalize_path(path: &str) -> String {
+    let decoded = percent_encoding::percent_decode_str(path).decode_utf8_lossy();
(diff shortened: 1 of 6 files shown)


madmax983:
Backlog triage 2026-07-08: duplicate exemption-path traversal fix; #1229 kept as the rework candidate

---

PR #1143, opened 2026-06-07
Add Model Context Protocol (MCP) support for AI agent integration

madmax983:
## Summary

This PR adds comprehensive Model Context Protocol (MCP) support to Autumn, enabling AI agents to call typed JSON endpoints through a Streamable-HTTP MCP server. Routes are projected as agent-callable tools with schemas derived from the handler's existing `ApiDoc` metadata, ensuring the tool catalog cannot drift from the actual handler contract.

## Key Changes

- **New `mcp` module** (`autumn/src/mcp.rs`): Core MCP server implementation
  - Derives tool catalog from route registry with per-endpoint opt-in via `#[api_doc(mcp)]`
  - Implements Streamable-HTTP JSON-RPC protocol handling (`initialize`, `tools/list`, `tools/call`)
  - Reconstructs HTTP requests from tool arguments and dispatches through the real handler pipeline
  - Forwards bearer tokens so authentication, authorization, and validation apply identically to agent and HTTP calls
  - Builds self-contained `inputSchema` from path params, query extractors, and request bodies with inlined component refs

- **Macro enhancements** (`autumn-macros/src/api_doc.rs`):
  - Added `mcp_tool` and `mcp_exclude` boolean flags to `ApiDocAttr`
  - Parses `#[api_doc(mcp)]`, `#[api_doc(mcp = true)]`, and `#[api_doc(mcp = false)]` syntax

- **AppBuilder API** (`autumn/src/app.rs`):
  - `mount_mcp(path)`: Mount MCP endpoint with per-endpoint opt-in
  - `expose_all_as_mcp()`: Whole-API hatch that auto-includes read-only (`GET`) endpoints but requires explicit opt-in for mutating verbs; per-endpoint exclusions always honored

- **Router integration** (`autumn/src/router.rs`):
  - Tool derivation happens during router assembly before state application
  - MCP runtime config flows through `RouterContext`

- **Test support** (`autumn/src/test.rs`):
  - `TestApp::mount_mcp()` mirrors production API for integration testing

- **OpenAPI integration** (`autumn/src/openapi.rs`):
  - Reuses component schema resolution to ensure tool schemas match OpenAPI specs
  - Added `#[allow(clippy::struct_excessive_bools)]` for orthogonal route properties

- **Documentation** (`docs/guide/mcp.md`):
  - Comprehensive guide covering feature enablement, endpoint tagging, tool invocation, authentication reuse, whole-API hatch, eligibility rules, and end-to-end examples

- **Integration tests** (`autumn/tests/mcp_endpoint.rs`):
  - Validates opt-in enforcement, tool derivation, safety annotations, read/write dispatch, bearer token auth, and whole-API hatch behavior
  - Tests real pipeline enforcement (e.g., `RequireApiToken` applies to agent calls)

- **Example updates** (`examples/todo-app`):
  - Enabled `mcp` feature in Cargo.toml
  - Tagged JSON API endpoints with `#[api_doc(mcp)]`
  - Demonstrates bearer-token-protected tools in a scoped group

## Notable Implementation Details

- **No second schema**: Tool `inputSchema` is built from the same `SchemaEntry` data OpenAPI uses, preventing drift
- **Real pipeline dispatch**: `tools/call` uses the fully-assembled router (state applied), so middleware, auth layers, and validation run identically for agent and HTTP calls
- **Bearer token forwarding**: `Authorization` header from the MCP request is forwarded into the dispatched call, reusing existing `RequireApiToken` infrastructure
- **Safety annotations**: HTTP method maps to MCP hints (`readOnlyHint`, `destructiveHint`) so agents can reason about side effects
- **JSON-only eligibility**: Routes without a JSON response schema (HTML/Maud) are auto-excluded; opted-in ineligible routes emit a build-time warning
- **Ref rewriting**: Component schema references are recursively rewritten from `#/components/schemas/X` to local `#/$defs/X` so tool schemas are self-contained
- **Batch and notification

https://claude.ai/code/session_017qkeK18g6jrpkzQFg9rkPb


madmax983:
Thanks for the review (and the bot pass). I went through each item; here's what I applied and what I deliberately didn't, with reasoning.

**Applied (pushed in `5cf127d`):**

- **Unbounded response buffering (`usize::MAX`)** — capped the buffered tool response at 10 MiB. The in-process dispatch fully buffers the handler's body to repackage it as a tool result (unlike a normal streamed HTTP response), so this guard is warranted. I report an overflow as an explicit `isError` rather than the suggested `unwrap_or_default()`, which would silently return an empty body as if it succeeded.
- **JSON-RPC empty/invalid batch** — an empty batch and any non-object/array message now return `-32600 Invalid Request`, per the JSON-RPC 2.0 spec.
- **Mount path without leading `/`** — instead of normalizing silently, I validate the path at build time and return a recoverable `RouterBuildError::InvalidMcpPath`. This mirrors the framework's existing `OpenApiConfig` path handling (surface typos as errors, not magic fixes) and still prevents the axum panic. Covers all three call sites (`AppBuilder`, `TestApp`, `McpRuntime`) from one place.

**Not applied — `let_chains` "breaks stable Rust" (the two `high` items):** this isn't applicable to this crate. Let-chains were **stabilized in Rust 1.88.0 + edition 2024**, which is exactly this crate's `rust-version`/`edition`. The **`MSRV (1.88.0)` CI job passes** on this PR, confirming it compiles on the floor toolchain. Reverting to nested `if let` would also re-introduce the `clippy::collapsible_if` warnings I removed earlier, which the `Lint` gate (`-D warnings`) rejects — so that change would turn CI red.

Separately, I fixed the real CI failure (**Documentation build**): the always-compiled `ApiDoc::mcp_exclude` field had an intra-doc link to the `mcp`-gated `expose_all_as_mcp`, which broke `cargo doc` when the feature is off. De-linked it and added `mcp` to the docs.rs feature mirror in `check-docs.sh` (verified locally: "Documentation build OK"). Added 3 tests for the new behaviors.

---

madmax983:
Addressed the Codex review (commits `1de72ba` + `1ef128d`). Summary:

**Security posture** — I researched how the MCP spec and ecosystem (`fastapi-mcp`, the reference SDKs, the `rmcp` DNS-rebinding CVE) handle this and went with **Origin validation + optional endpoint auth**:
- **`Origin` validation** (spec MUST): a browser `Origin` not in the app's CORS `allowed_origins` (or `*`) gets `403`; requests with **no** `Origin` (non-browser agents) pass — blocks DNS rebinding without breaking agents.
- **`secure_mcp(layer)`** gates the *entire* endpoint (catalog included) behind any tower layer, e.g. `RequireApiToken` — mirroring `fastapi-mcp` reusing the app's auth. The per-route layer already protects tool *dispatch*; this covers `initialize`/`tools/list` too.
- **CSRF/session parity**: `Cookie` and `X-CSRF-Token` are now forwarded alongside `Authorization`, so `#[secured]` and CSRF-protected routes dispatch identically to a direct call.

**JSON-RPC / dispatch robustness** — require `jsonrpc:"2.0"`; reject empty/malformed batches and non-object `arguments` (`-32600`/`-32602`) instead of a silent 202/dispatch; negotiate only supported protocol versions; enforce required `body`; explode array query fields into repeated keys; reuse the framework path-segment encoder; cap tool-result bodies at 10 MiB.

**Other findings** — registered OpenAPI component schemas now flow into `inputSchema` (no placeholder drift); duplicate tool names keep the first registration with a build-time warning.

Verified locally: `cargo clippy --features mcp --all-targets` clean, `check-docs.sh` OK, 14 unit + 19 integration MCP tests green. Codecov is informational (project coverage +0.01%).

The one item I deliberately did **not** apply is the bot's earlier `let_chains` "breaks stable Rust" suggestion — let-chains are stable on this crate's MSRV (1.88 + edition 2024), the `MSRV (1.88.0)` job passes, and reverting would re-introduce `clippy::collapsible_if` under the `-D warnings` `Lint` gate.

---

PR #1220, opened 2026-06-13
👺 Havoc: Fix logic error in InMemoryWebhookReplayStore cleanup_if_due

madmax983:
👺 Havoc: Fix logic error in InMemoryWebhookReplayStore cleanup_if_due

🧊 **The Trigger:**
A webhook request payload comes in with an artificially high (future) timestamp, bypassing standard validation, or just a clock sync issue.

📉 **The Stack Trace:**
No stack trace, but legitimate webhook replay tokens are wrongfully evicted from the cache. The `cleanup_if_due` function relied on the `received_at` timestamp from the incoming request payload instead of `SystemTime::now()` for identifying which other keys had expired.

🧪 **Reproduction:**
1. A legitimate request is sent and protected against replays for 60s.
2. A malicious request is sent with a future timestamp (e.g., +100s) and triggers the cleanup routine (after 128 items).
3. The original legitimate request is evicted prematurely.
4. Run `cargo test -p autumn-web --test security_tests -- test_webhook_cleanup_vulnerability` to observe the failure prior to the fix.

😈 **Comment:**
"You assumed the webhook's timestamp would never be larger than wall clock time during cleanup. You were wrong."

---
*PR created automatically by Jules for task [7957446872592130202](https://jules.google.com/task/7957446872592130202) started by @madmax983*


madmax983:
Backlog triage 2026-07-08: webhook replay-store cleanup; webhook.rs reworked since — re-file if still reproducible

---

PR #1249, opened 2026-06-14
RFC 8058 one-click List-Unsubscribe for bulk mail compliance

madmax983:
## Summary

Implements RFC 8058 one-click `List-Unsubscribe` support for bulk mail compliance with Gmail and Yahoo's February 2024 requirements. Mailers can opt into list-based unsubscribe handling with a single attribute, and the framework automatically manages signed tokens, suppression tracking, and header emission.

## Key Changes

- **Mailer attribute**: Added `#[mailer(list_unsubscribe = "scope")]` to declare bulk mail lists. The `#[mailer]` macro now parses this optional argument and tags generated messages with the scope.

- **Configuration**: Extended `MailConfig` with three new fields:
  - `unsubscribe_base_url`: Base URL for one-click links (e.g., `https://app.example.com`)
  - `unsubscribe_mailto`: Fallback `mailto:` address for the header
  - `unsubscribe_token_ttl_days`: Token validity window (default 30 days)
  - Startup validation ensures at least one destination is configured when list mailers exist

- **Token signing**: Implemented stateless, signed unsubscribe tokens in `mail::unsubscribe` module:
  - Format: `base64url(subscriber).base64url(list_id).expiry.HMAC-SHA256`
  - Tokens are verified against current/previous signing keys and checked for expiry
  - Subscriber identifier is opaque in URLs, revealed only on successful verification

- **Suppression store**: New `SuppressionStore` trait for persistent opt-out tracking:
  - `InMemorySuppressionStore` for dev/test
  - `DbSuppressionStore` (referenced, implementation in separate module) for production
  - Registered on `AppState` via `SuppressionStoreHandle`

- **Mail sending**: Enhanced `Mailer::send()` to handle list mail:
  - Checks suppression store and skips suppressed recipients (with structured logging)
  - Delivers one recipient per message for personalized unsubscribe links
  - Injects `List-Unsubscribe` and `List-Unsubscribe-Post: List-Unsubscribe=One-Click` headers per RFC 8058

- **Default endpoint**: Mounts `/_autumn/unsubscribe` route that:
  - Accepts GET to render a confirmation form
  - Accepts POST (one-click) to verify token and record suppression
  - Rate-limited, no end-user authentication required

- **Mail struct**: Added two new fields to `Mail`:
  - `list_unsubscribe: Option<String>`: Logical list scope (set by macro or builder)
  - `extra_headers: Vec<(String, String)>`: Raw headers for transport emission

- **Dev preview**: Enhanced mail preview UI to inject sample RFC 8058 headers so authors can confirm wiring without sending real mail.

- **CLI scaffolding**: `autumn generate mailer --list-unsubscribe <scope>` now:
  - Generates mailer with the attribute pre-set
  - Idempotently creates `migrations/<ts>_create_mail_unsubscribes/` migration for the suppression table

- **Doctor checks**: Added `check_mail_unsubscribe_config_impl()` to validate:
  - Fails in production if a mailer declares `list_unsubscribe` but no destination is configured
  - Warns in non-production environments

- **Inventory registration**: `MailerListUnsubscribeDescriptor` allows compile-time enumeration of all declared list scopes for validation and tooling.

## Notable Implementation Details

- Suppression checks and header signing happen inside `Mailer::send()`, so both immediate and deferred delivery (via `deliver_later`) are covered automatically.
- Tokens use the app's existing signing key infrastructure (`ResolvedSigningKeys`) for HMAC-SHA256.
- List mail is delivered one recipient at a time to personalize unsubscribe links; transactional mail (without `list_unsubscribe`) is unaffected.
- Fail-closed in production: startup rejects configurations where bulk mail is

https://claude.ai/code/session_01Bz8nGjp7GUNebBNB5TyvAm


madmax983:
Pushed `9833707` addressing the CI failures and automated review feedback.

**CI gates**
- **Lint** failed on `cargo fmt --check` (formatting only) — ran rustfmt.
- **Documentation build** failed on broken rustdoc intra-doc links — fixed (`cfg`-gated / private-path types now use plain backticks). `./scripts/check-docs.sh` passes locally.

**Fixed from review (real bugs)**
- **CSRF + one-click POST:** exempted `UNSUBSCRIBE_PATH` from CSRF (mirrors the webhook exemption) so mailbox-provider one-click POSTs aren't rejected in `prod`.
- **Env overrides:** wired `AUTUMN_MAIL__UNSUBSCRIBE_BASE_URL` / `_MAILTO` / `_TOKEN_TTL_DAYS` into `apply_mail_env_overrides_with_env` (and the `doctor` resolver).
- **Recipient normalization:** suppression + token keys now use a canonical bare, lowercased address, so `Ada <ada@example.com>` matches an opt-out recorded as `ada@example.com`.
- **mailto-only:** `List-Unsubscribe-Post` is only emitted with an HTTPS one-click URL; mailto-only is plain RFC 2369. The endpoint mounts on `base_url` only.
- **Persistence:** prod fails closed when a one-click endpoint would be live but no suppression backend exists; the POST handler returns 503 instead of falsely confirming when it can't persist.
- **HTTPS:** `unsubscribe_base_url` must be `https://` in prod; non-HTTPS base URLs are rejected at startup.
- **Robustness:** `doctor` strips comments before detecting `#[mailer(list_unsubscribe …)]`; invalid extra-header names now log a warning; sending list mail without a runtime warns.

**Intentionally not changed (with rationale)**
- **let-chains flagged as "unstable" (3× HIGH):** let-chains stabilized in **Rust 1.88**, which is this repo's MSRV (there's an `MSRV (1.88.0)` CI job), the codebase already uses them (e.g. `autumn-macros/src/mailer_preview.rs`), and clippy's `collapsible_if` (run with `-D warnings`) actively *requires* the collapsed form. Reverting to nested `if let` would break the Lint gate. The green MSRV + Lint runs confirm they compile.
- **Token base64 "exposes" the subscriber:** the subscriber is the recipient's own address — the same value already in the message `To`/envelope seen by any intermediary handling the mail — and the token is HMAC-signed (tamper-proof). AC #3 explicitly permits "what the token reveals on verification," and the chosen design is stateless HMAC (no token table). I softened the doc comment to describe this accurately rather than claim opacity. Happy to switch to an encrypted payload if you'd prefer true opacity.
- **Override detection only inspects `route_list` (P2):** the documented override is registering your own route at `/_autumn/unsubscribe` via `routes[]`, which is detected. Detecting a path inside an opaque `.merge()`/`.nest()` router isn't introspectable in axum; if you want a first-class override for those cases I can add an explicit `AppBuilder::unsubscribe_endpoint(...)` opt-out — let me know.

Full local verification green: workspace build, `autumn-web --features "mail db"` lib + e2e, `autumn-cli`, clippy (`-D warnings`), rustfmt, and the docs gate.

---

madmax983:
Pushed `65c2ea3`.

**Design change (per maintainer): the default endpoint is now opt-in.** A plain JSON-API autumn app shouldn't have to opt *out* of an HTML unsubscribe endpoint. Enable it with `AppBuilder::mount_unsubscribe_endpoint()` (or `mail.mount_unsubscribe_endpoint = true`); it mounts only when opted in **and** `mail.unsubscribe_base_url` is set. This also removes the fragile `route_list` override detection — to serve a custom page you simply don't opt in and register your own route at the path (keeping its own CSRF/CAPTCHA).

This resolves the override-mechanism threads (GET-only override, CSRF-exemption breadth, `.merge()`/`.nest()` invisibility): CSRF **and** CAPTCHA exemptions now key off the opt-in, so they only apply when the framework owns the endpoint.

**Also fixed this round:**
- **One-click POST body required** — `POST` now requires the RFC 8058 `List-Unsubscribe=One-Click` body before recording a suppression (rejects arbitrary/scanner POSTs).
- **Disabled transport** — the list-unsubscribe fail-closed guard is skipped when `transport = "disabled"` (mirrors the durable-queue/suppression guards).
- **URL validation** — `unsubscribe_base_url` must be an absolute `https://` URL *with a host* in prod (rejects `https://`, `https:///path`).
- **doctor** — merges profile aliases (`production`/`development`, not just `prod`/`dev`); honors blank env overrides (a present-but-empty `AUTUMN_MAIL__UNSUBSCRIBE_*` clears the TOML value, matching the runtime); strips line **and** block comments before detecting `#[mailer(list_unsubscribe …)]`.
- **CI** — fixed the `clippy::map_unwrap_or` that broke the Lint gate (CI runs clippy 1.96; I've matched my local toolchain to 1.96 now, so this class won't recur).

**Not changed (noted):**
- *Reserve the unsubscribe path in probe/OpenAPI/MCP collision preflights* (router.rs): with opt-in the endpoint only exists when explicitly requested, so the surface is small; an operator who also points a probe/OpenAPI/MCP route at `/_autumn/unsubscribe` would get a clear duplicate-route panic rather than the recoverable collision error. I can wire it into the preflights if you'd like it to fail soft — say the word.

Verified locally on Rust **1.96** (matching CI): `clippy -D warnings` clean, rustfmt, the docs gate, and the mail/e2e/doctor/generator suites all pass.

---

madmax983:
Pushed `253082f`. CI Lint + MSRV(1.88) went green on the previous commit, so the let-chain comments are confirmed false positives (they compile on the repo's MSRV and are required by `clippy::collapsible_if`).

Addressed in this commit:
- **Env opt-in** — added `AUTUMN_MAIL__MOUNT_UNSUBSCRIBE_ENDPOINT`.
- **Base-URL validation** — reject query/fragment bases (`https://h?x`, `https://h#y`) that would break the appended `?token=`.
- **mailto normalization** — a configured `mailto:foo@x` no longer double-prefixes to `<mailto:mailto:…>`.
- **Header dedup on send** — a template that already set `List-Unsubscribe` is not double-emitted (matches the preview path).
- **Macro registration** — emitted per `Mail` method carrying its `#[cfg]`, so a build that `cfg`-removes all mail methods registers no descriptor (no spurious startup requirement).
- **Confirmation form** — relative `action="?token=…"` preserves a reverse-proxy base-path prefix.
- **Unmounted-endpoint heads-up** — `install_mailer` now warns (not fails) when one-click links are advertised but the default endpoint isn't opted in and no custom route is known (failing would break the legitimate custom-route path).

Still intentionally open:
- **Token base64 (P1)** — awaiting your call. It's the recipient's own address (already in `To`/envelope), HMAC-signed, and AC #3 permits token-revealed identity; the doc no longer claims opacity. I'll switch to an encrypted payload if you'd prefer.
- **Collision preflight for `/_autumn/unsubscribe` (P2)** — with opt-in the endpoint only exists when explicitly requested; a colliding probe/OpenAPI/MCP path would panic clearly rather than silently. Happy to wire it into the preflights for a soft error if you want.

Verified locally on Rust 1.96 (matching CI): clippy `-D warnings`, rustfmt, docs gate, and the mail/e2e/config/doctor/generator/macro suites all pass.

---

madmax983:
Pushed `e5e8210` (CI was fully green on `253082f`). This round:

- **doctor** now skips the `mail_unsubscribe` failure when `mail.transport = "disabled"` (mirrors the runtime guard — no list mail is emitted).
- **doctor** merges both spellings of the prod/dev profile pair (`prod`+`production`, `dev`+`development`) regardless of which `AUTUMN_ENV` value is used, matching the runtime loader.
- **validate** `mail.unsubscribe_mailto` is a real mailbox / `mailto:` URI in prod (rejects e.g. `unsubscribe example.com`).

**Deferred (with rationale):** *Ignore cfg-disabled list mailers in `doctor`* — `doctor` statically text-scans source and cannot resolve which `#[cfg]` features are active for a given deployment build, so it can't distinguish a `#[cfg(feature="marketing")]` list mailer that's compiled-out. The **authoritative** fail-closed check is the runtime startup guard, which is cfg-aware (the macro now registers the inventory descriptor per-method with its `#[cfg]`). `doctor` is a conservative pre-deploy advisory; over-detecting errs safe. Making it cfg-accurate would require compiling the project. Happy to revisit if you'd prefer doctor parse with a feature set.

Still awaiting your call on the two earlier items: encrypt the token payload (vs. the current signed base64), and adding `/_autumn/unsubscribe` to the probe/OpenAPI/MCP collision preflights.

Verified on Rust 1.96: clippy `-D warnings`, rustfmt, docs gate, and the mail/e2e/config/doctor/generator suites all pass.

---

PR #1322, opened 2026-06-20
⚡ Bolt: [performance improvement] optimize encode_catch_all_param

madmax983:
💡 What: Replaced the inefficient `.collect::<Vec<String>>().join("/")` chain in `encode_catch_all_param` with a single, pre-allocated `String` and an iterative loop.
🎯 Why: The original implementation caused multiple heap allocations per call: one for each segment encoded via `to_string()`, one for the intermediate `Vec<String>`, and one final string allocation during `.join("/")`. This created unnecessary allocation pressure on a frequently hit hot path for catch-all parameter encoding.
📊 Impact: Eliminates the intermediate vector and temporary string allocations. Reduces memory allocation overhead while preserving safe iteration and bounds without unsafe code.
🔭 Measurement: Verify correct behavior and bounds checks by running `cargo test -p autumn-web --lib paths`.

---
*PR created automatically by Jules for task [10170048980067326271](https://jules.google.com/task/10170048980067326271) started by @madmax983*

### DIFF
--- a/autumn/src/paths.rs
+++ b/autumn/src/paths.rs
@@ -54,22 +54,30 @@ pub fn encode_path_segment(value: impl std::fmt::Display) -> String {
 ///
 /// Splitting on `/` allows preserving directory slashes while percent-encoding
 /// other characters in each segment.
+///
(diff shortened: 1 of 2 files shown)


madmax983:

---

PR #1405, opened 2026-06-22
🗣️ Echo: [DX Audit] Missing routes and dev env

madmax983:
🗣️ Echo: [DX Audit] Missing routes and dev env

🔎 **EXPERIENCE**
- Attempted to use `.env` files directly with `autumn dev` to load a database URL.
- Tested how the application responds to invalid route accesses (e.g., `curl -v http://localhost:3000/missing`) via the default generated app without explicitly requesting JSON or HTML.

🚧 **STUMBLE**
- `autumn dev` ignores `.env` files completely.
- Accessing a missing route with curl (which defaults to `Accept: */*`) results in a blank response body (`content-length: 0`) rather than a clear error message.

📢 **REPORT**
- "Why does `autumn dev` ignore my `.env` file? Almost every other framework ecosystem automatically loads `.env` when running in development mode."
- "Why does a 404 give me an empty page? The framework seems to have error handling logic, but it's completely skipping over it for 404s when an `Accept` header is not explicitly set, or under certain default handler setups!"

🧪 **VERIFY**
- Confirmed that `.env` loading is unsupported natively by the runner.
- Traced the blank 404 issue back to the `fallback_404_handler` returning `AutumnError::not_found_msg` which, when the exception filter checks the `Accept` headers for HTML/JSON preference, falls through if `Accept` is not specific enough, stripping representation headers and resulting in an empty response body.

---
*PR created automatically by Jules for task [16861173009402277594](https://jules.google.com/task/16861173009402277594) started by @madmax983*

### DIFF
--- a/dx_audit_report.md
+++ b/dx_audit_report.md
@@ -116,3 +116,38 @@ To make the developer experience more robust ("idiot-proofing"):
 
 ## 4. 🧪 VERIFY - The "idiot proofing"
 - Confirmed that Axum's `IntoResponse` trait is not implemented for `i32`, `i64`, or other plain numbers out-of-the-box, meaning they cannot be returned directly from route handlers without manually converting them to strings or JSON first.
+
+# DX Audit Report: `autumn dev` ignoring `.env` file
+
+## 1. 🔍 EXPERIENCE - The Walkthrough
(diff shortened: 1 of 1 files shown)


madmax983:

---

PR #1420, opened 2026-06-23
⚡ Bolt: [performance improvement]

madmax983:
💡 What: Replaced multiple instances of the `.split('/').map(...).collect::<Vec<_>>().join("/")` anti-pattern in `autumn/src/paths.rs` and `autumn/src/mcp.rs` with pre-allocated `String::with_capacity` and an explicit loop.
🎯 Why: The original pattern caused multiple unnecessary heap allocations: one `Vec` to collect the intermediate segments, multiple `String` creations inside the `map` iterator, and another `String` creation when joining. This is extremely inefficient on hot paths.
📊 Impact: Removes the intermediate `Vec` allocation and multiple `String` allocations by writing directly into a pre-allocated capacity string. Eliminates O(N) extra heap allocations where N is the number of path segments.
🔭 Measurement: Verify by running `cargo test -p autumn-web --lib` and observing that `paths::tests` pass with the exact same encoded path output, proving the fast-path implementation logic is safe and correct.

---
*PR created automatically by Jules for task [1497221663685973068](https://jules.google.com/task/1497221663685973068) started by @madmax983*

### DIFF
--- a/autumn/src/mcp.rs
+++ b/autumn/src/mcp.rs
@@ -1594,11 +1594,14 @@ fn collapse_sse_body(bytes: &[u8]) -> String {
         .into_iter()
         .partition(|e| e.event.as_deref() == Some("result"));
     if results.is_empty() {
-        progress
-            .into_iter()
-            .map(|e| e.data)
-            .collect::<Vec<_>>()
-            .join("\n")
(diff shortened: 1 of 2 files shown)


madmax983:
Backlog triage 2026-07-08: overlapping collect/join micro-opts; live successor #1484 covers the surviving hot paths

---

PR #1434, opened 2026-06-25
🛡️ Sentry: [test coverage improvement]

madmax983:
🎯 Target: `is_excluded` and `normalize_exclusions` inside `autumn/src/middleware/access_log.rs`.
💣 Risk: If an API user tries to exclude the root route `/` from access logs (e.g. for high-volume load balancers pinging root), it silently fails to be excluded because the normalizer trims the trailing slash to an empty string `""` and then incorrectly filters it out.
🧪 Strategy: Modified `normalize_exclusions` to map an empty string (when derived from trimming `/`) back to `"/"` and to not filter it out. Added specific unit tests testing root path exclusion behavior in `is_excluded`.
🔬 Verification: Run `cargo test -p autumn-web --lib middleware::access_log`

---
*PR created automatically by Jules for task [13914752732465933858](https://jules.google.com/task/13914752732465933858) started by @madmax983*

### DIFF
--- a/autumn/src/middleware/access_log.rs
+++ b/autumn/src/middleware/access_log.rs
@@ -127,6 +127,8 @@ fn normalize_exclusions(exclude: Vec<String>) -> Arc<[String]> {
             let trimmed = prefix.trim_end_matches('/');
             if trimmed.len() == prefix.len() {
                 prefix
+            } else if trimmed.is_empty() {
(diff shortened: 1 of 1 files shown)


madmax983:
Backlog triage 2026-07-08: tests target internals that have since changed; regenerating against trunk is cheaper than rebasing

---

PR #1469, opened 2026-06-27
🍎 Eris: [security fix] Length-dependent timing side-channel via subtle::ConstantTimeEq misuse

madmax983:
🦠 Threat: `subtle::ConstantTimeEq::ct_eq` on variable-length structures (like `&str` and `&[u8]`) short-circuits on length mismatch. Using it for user-supplied string comparisons (like CSRF tokens, HMAC signatures, and upload URLs) introduces a length-dependent timing side-channel. An attacker could potentially enumerate token lengths based on timing differences, making brute-force attacks significantly more feasible.
🛡️ Defense: Implemented a robust `constant_time_eq_str` and `constant_time_eq` centralized logic in `autumn/src/security/constant_time.rs`. This correctly manages the comparison by always processing exactly `expected.len()` bytes, utilizing `subtle::Choice` and a dummy fallback byte `0xFF` to prevent any short-circuiting while retaining `#[inline(never)]`. Refactored all vulnerable usages across `auth.rs`, `inbound_mail.rs`, `pagination.rs`, `config.rs`, `local.rs`, and `csrf.rs` to use this new module.
💥 Severity: Medium. While timing attacks over the network are difficult to execute with high precision, exploiting length differences takes significantly fewer queries than a pure byte-by-byte timing leak.
🧪 Verification: Added `test_csrf_constant_time_length` in `tests/security/csrf_length_timing.rs`. The test sends a large token of length `1` and one of matching length but differing characters, then asserts that the short token is not rejected instantaneously via a short-circuit.

---
*PR created automatically by Jules for task [11746739093650897500](https://jules.google.com/task/11746739093650897500) started by @madmax983*

### DIFF
--- a/autumn/src/auth.rs
+++ b/autumn/src/auth.rs
@@ -991,10 +991,7 @@ async fn validate_callback_state(
     let expected_state = session.get(&state_key).await.ok_or_else(|| {
         crate::AutumnError::unauthorized_msg("oauth2 state missing; restart login")
     })?;
-    if subtle::ConstantTimeEq::ct_eq(expected_state.as_bytes(), callback.state.as_bytes())
-        .unwrap_u8()
-        != 1
-    {
+    if !crate::security::constant_time::constant_time_eq(expected_state.as_bytes(), callback.state.as_bytes()) {
         return Err(crate::AutumnError::unauthorized_msg(
             "oauth2 state mismatch",
         ));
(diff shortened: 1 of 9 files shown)


madmax983:
Backlog triage 2026-07-08: fix already landed on trunk-dev (constant-time ct_eq_str via subtle::ConstantTimeEq in security/config.rs)

---

PR #1496, opened 2026-06-29
🗣️ Echo: [DX Audit]

madmax983:
🔍 EXPERIENCE
I went through the Quickstart steps in the `README.md` to see what getting started with Autumn is like. I copied the code snippet verbatim:
```rust
use autumn_web::prelude::*;

#[get("/")]
async fn index() -> &'static str {
    "Welcome to Autumn!"
}

#[get("/hello/{name}")]
async fn hello_name(name: autumn_web::extract::Path<String>) -> String {
    format!("Hello, {}!", *name)
}

#[autumn_web::main]
async fn main() {
    autumn_web::app()
        .routes(routes![index, hello_name])
        .run()
        .await;
}
```
I compiled and ran it, and then visited `http://localhost:3000/hello/Echo` in my terminal.

🚧 STUMBLE
1. While setting up a test project to verify route paths, I mistakenly typed `#[get("/hello/:name")]` using colon-based parameter syntax rather than braces, which caused the application to panic upon starting.
2. In the README example, you've provided `name: autumn_web::extract::Path<String>` as a parameter type. It feels overly verbose and complex when the very first line is `use autumn_web::prelude::*;`.
3. When I tried visiting `http://localhost:3000/not-found` with my curl without an `Accept` header or an `Accept: application/json` header, I actually got a JSON blob (`application/problem+json`), which was expected for API calls.

📢 REPORT
1. The error message I got when I fat-fingered the route parameter `#[get("/hello/:name")]` is: `Path segments must not start with ':'. For capture groups, use '{capture}'. If you meant to literally match a segment starting with a colon, call 'without_v07_checks' on the router.`. This is extremely helpful! Way to go on the error messages.
2. For the Import Scan: Why do I have to type `autumn_web::extract::Path<String>`? If `Path` is a common extractor and I imported the prelude, I shouldn't have to specify the fully qualified module name!
3. The README example itself works fine and is easy to follow.

🧪 VERIFY
As Echo, I do not fix code or documentation. A framework maintainer should consider updating the `README.md` to just use `Path<String>` instead of the fully qualified path to simplify the developer experience, since `Path` is already exposed in the prelude.

---
*PR created automatically by Jules for task [5491543138814085391](https://jules.google.com/task/5491543138814085391) started by @madmax983*


madmax983:

---

PR #1512, opened 2026-06-30
🗣️ Echo: [DX Audit] Error Page and Trait Bounds

madmax983:
🗣️ Echo: [DX Audit] Error Page and Trait Bounds

🔎 EXPERIENCE
- Did the "README Run": Copied the exact example code from `README.md` into a new project's `main.rs`.
- Specifically, the example defines `async fn index() -> &'static str { "Welcome to Autumn!" }` and `async fn hello_name(name: autumn_web::extract::Path<String>) -> String`.
- Tested writing a simpler custom route: `async fn foo() -> i32 { 42 }`.

🚧 STUMBLE
- **Error Check 1**: Requesting an unmatched route like `http://localhost:3000/missing` returns an empty 404 response body (`content-length: 0`) for JSON requests, but returns an HTML error page full of stack traces if the client accepts text/html. Both are confusing.
- **Error Check 2**: Putting a misspelled function inside the `routes!` macro (e.g. `routes![hello_name_typo]`) produces a secondary compiler error `cannot find function __autumn_route_info_hello_name_typo in this scope`.
- **Error Check 3**: The `foo` route handler returning `i32` completely fails to compile with a massive, unintelligible error: `the trait bound fn() -> ... {foo}: Handler<_, _> is not satisfied`.

📢 REPORT
- "Why does a 404 give me an empty page for API calls and stack traces for HTML? Simple is better than powerful, but empty or scary is just confusing."
- "If I make a typo in the `routes!` macro, why do I get a weird error about `__autumn_route_info_...`? I am the dumbest person in the room, just tell me 'Route not found'."
- "Why can I return a String but not an integer? If I return `42`, the compiler dumps 20 lines of trait bound errors about Axum internals. Simple is better than powerful, and right now simple numbers crash the compiler!"

🧪 VERIFY
- Confirmed that standard `curl` requests to missing paths receive `content-length: 0` if `Accept: application/json` is used.
- Verified that the secondary `__autumn_route_info_...` error is unavoidable due to eager macro resolution, meaning we must accept it as an ergonomic trade-off, but it could be explicitly documented.
- Confirmed that Axum's `IntoResponse` trait is not implemented for `i32`, `i64`, or other plain numbers out-of-the-box.

---
*PR created automatically by Jules for task [2298338873222064075](https://jules.google.com/task/2298338873222064075) started by @madmax983*


madmax983:

---

PR #1534, opened 2026-07-02
⚡ Bolt: [performance improvement]

madmax983:
💡 What:
The optimization refactored `autumn/src/paths.rs` to avoid multiple intermediate `String` and `Vec<String>` allocations.
Specifically:
1. `encode_catch_all_param` avoids allocating a `Vec<String>` before joining on `/`. Instead, it uses a single `String` allocation, and iterates over `.split('/')`.
2. `percent_encode` now calls `percent_encode_to(&mut String)` allowing strings to be directly pushed to existing buffers.
3. `with_query` avoids generating intermediate `encoded_key` and `encoded_value` strings and replaces the `format!` allocation with pushing strings directly to the self instance `String`.

🎯 Why:
The original implementation heavily relied on `.collect::<Vec<_>>().join()` and creating separate formatted strings for path modifications. Because path generation is a hot path (e.g., heavily used across views and links in HTTP processing pipelines), creating many short-lived heap allocations is inefficient.

📊 Impact:
Removes 2 intermediate `String` heap allocations per `with_query` call (for the encoded key and value, plus the `format!` result).
Removes the allocation of the `Vec<String>` and multiple intermediate strings in `encode_catch_all_param`.

🔭 Measurement:
Run `cargo test -p autumn-web --lib` to ensure no functionality breakage. We expect measurable performance gains in high-throughput path generation benchmarks where GC / Heap operations typically slow execution.

---
*PR created automatically by Jules for task [11024871824566580243](https://jules.google.com/task/11024871824566580243) started by @madmax983*


madmax983:

---

PR #1560, opened 2026-07-04
🗣️ Echo: [DX Audit]

madmax983:
🔎 EXPERIENCE - The Walkthrough
- Read the `README.md` and followed the quickstart guide.
- Created a new test project linking to the local `autumn-web` package.
- Copied the example `main.rs` from `README.md` into `src/main.rs`.
- Attempted to run the app using `cargo run`.
- Visited `http://localhost:3000/hello/echo` and received `Hello, echo!` correctly.
- Created an intentional typo in the `routes![]` macro.
- Wrote a custom route handler returning an `i32` primitive.
- Visited a non-existent route (`http://localhost:3000/missing`).

🚧 STUMBLE - The Friction Points
- **Error Check 1**: Requesting a missing path like `http://localhost:3000/missing` returns an empty response body (`content-length: 0`). Users expect a default 404 page or JSON error object, not a completely blank response.
- **Error Check 2**: Putting a non-existent function inside the `routes!` macro (e.g. `routes![index, missing_route]`) produces a compiler error exposing macro internals: `cannot find function __autumn_route_info_missing_route in this scope`. This makes it harder for users to understand that they simply misspelled a route name.
- **Error Check 3**: Defining a custom route handler that returns a primitive `i32` number (like `async fn foo() -> i32 { 42 }`) fails to compile with a massive, unintelligible error: `the trait bound fn() -> ... {foo}: Handler<_, _> is not satisfied`. The error output references internal Axum routing boundaries.
- **Error Check 4**: When attempting to return `autumn_web::error::Error`, it is not exported, requiring `AppError` to be used instead, though the framework prefix makes `Error` intuitive.

📢 REPORT - The Complaint
- "Why does a 404 give me an empty page? Simple is better than powerful, but empty is just confusing."
- "If I make a typo in the `routes!` macro, why do I get a weird error about `__autumn_route_info_...`? I am the dumbest person in the room, just tell me 'Route not found'."
- "Why can I return a String but not an integer? If I return `42`, the compiler dumps 20 lines of trait bound errors about Axum internals. Simple is better than powerful, and right now simple numbers crash the compiler!"
- "Why is it `AppError` when everything else expects `Error`?"

🧪 VERIFY - The "idiot proofing"
- Confirmed that standard `curl` requests to missing paths receive `content-length: 0` despite having framework error page middleware.
- Confirmed the macro compiler error by intentionally misspelling a route in `routes![]`. The secondary `__autumn_route_info_...` error is unavoidable due to eager macro resolution, meaning we must accept it as an ergonomic trade-off, but it could be explicitly documented.
- Confirmed that Axum's `IntoResponse` trait is not implemented for `i32`, `i64`, or other plain numbers out-of-the-box, meaning they cannot be returned directly from route handlers without manually converting them to strings or JSON first.

---
*PR created automatically by Jules for task [1243634458984921234](https://jules.google.com/task/1243634458984921234) started by @madmax983*


madmax983:

---

Issue #1048, opened 2026-05-31
Add `autumn destroy` to cleanly reverse generated files and edits


madmax983:
## Problem (the "So What?")
`autumn generate {model,scaffold,mailer,auth,task,migration,service,...}` is the framework's headline DX win — each command writes 5–8 files **and mutates shared files in place**: `routes![]` registration, `src/**/mod.rs` (`pub mod x;`), `src/main.rs` (preview registry wiring), and `Cargo.toml` (feature additions like `"mail"`). There is no inverse. Typo a field name, pick the wrong resource name, or run `scaffold` when you wanted `model`, and the only recovery is hand-deleting a migration directory + model + repository + routes + templates + smoke test, **then** surgically un-editing `routes![]` and `mod.rs` by hand. Miss one line and you get a dangling `mod posts;` or `routes![posts::index]` that fails to compile — turning a 10-second mistake into a 5-minute cleanup with a broken build in the middle. Every generator already builds a deterministic `Plan` of `Create`/`Modify` actions (`autumn-cli/src/generate/emit.rs`); the information needed to reverse it is already computed on every run.

## User Story
As a framework user iterating on a new resource, I want `autumn destroy <generator> <name>` to cleanly reverse exactly what the matching `autumn generate` created — files **and** in-place code edits — so that fixing a typo'd scaffold is a one-command round-trip that never leaves the project in a non-compiling state.

## Acceptance Criteria
- [ ] `autumn destroy <generator> <name> [fields...]` exists for every generator that `autumn generate` supports, taking the same positional arguments.
- [ ] For each `Create` action in the generator's plan, `destroy` removes the corresponding file (and prunes now-empty generated directories, e.g. `templates/posts/`).
- [ ] For each `Modify` action, `destroy` removes **only** the lines that `generate` inserted (the `pub mod posts;` line, the `routes![posts::...]` entries, the `Cargo.toml` feature addition) and leaves all other content byte-for-byte intact. Removing the last entry from a managed block does not leave malformed syntax.
- [ ] Migration directories are matched by resource-name **suffix** (the leading timestamp won't match on a later run) and removed only when the migration is **not yet applied**; if it appears applied, `destroy` skips the migration files, prints a clear warning, and tells the user to write a down-migration instead — it never touches the database.
- [ ] `autumn destroy ... --dry-run` prints a `Removed:`/`Reverted:` plan and exits 0 without touching disk, mirroring `generate --dry-run`.
[...]

PR #1579, opened 2026-07-07
Implement `autumn destroy` command to cleanly reverse `generate` (issue #1048)

madmax983:
Implements the `autumn destroy` command that cleanly reverses a matching `autumn generate` invocation by:

1. **Deleting generated files** — removes every file that `generate` created (refusing when content has diverged unless `--force`)
2. **Reverting in-place edits** — removes exactly the lines/entries `generate` inserted into shared files (mod declarations, routes entries, Cargo.toml deps/features, schema.rs tables)
3. **Pruning empty directories** — removes now-empty generated directories

## Key Changes

- **New `Revert` enum** (`emit.rs`) — 14 variants representing each type of in-place edit that can be reversed (ModDecl, RoutesEntries, SchemaTable, CargoDeps, CargoAutumnWebFeature, etc.)
  - Each variant tracks the file path and edit-specific metadata
  - Includes `owner_dir` for dependency/feature reverts to avoid removing shared deps while sibling resources still need them
  - Implements `apply()` method that idempotently removes the edit from file content

- **`Plan::revert()` method** — the inverse of `Plan::execute()`, implementing the core destroy logic:
  - Deletes Create/CreateBytes/CreateIfAbsent actions with content divergence checking
  - Applies each Revert to remove in-place edits
  - Deletes files emptied by reverts
  - Matches migration directories by resource-name suffix (timestamps differ on re-run)
  - Honors `--dry-run` flag

- **`Plan::push_revert()` method** — allows generators to record reverts alongside their Modify actions

- **Generator updates** — all generators now push Revert entries:
  - `model.rs`, `auth.rs`, `scaffold.rs`, `mailer.rs`, `job.rs`, `channel.rs`, `inbound_mail.rs`, `system_test.rs`, `pwa.rs`
  - Each tracks its own ModDecl, SchemaTable, CargoDeps, and feature entries
  - Cargo dependency reverts include `owner_dir` to handle sibling resources correctly

- **Removal functions** (`schema_edit.rs`) — idempotent inverses of append/modify functions:
  - `remove_mod_declaration()`, `remove_schema_table()`, `remove_routes_entries()`
  - `remove_main_mod_declarations()` for shared infrastructure modules
  - `remove_cargo_toml_patch()` for system tests
  - All return content unchanged if target is already absent

- **CLI integration** (`main.rs`) — new `Destroy` command that:
  - Takes identical subcommand and arguments as the matching `generate` call
  - Recomputes the plan with fresh timestamp (migrations matched by suffix)
  - Applies the plan in destroy mode via `Plan::revert()`

- **Test coverage** (`tests/generate.rs`) — acceptance test verifying round-trip:
  - `generate scaffold Post title:String` followed by `destroy scaffold Post title:String`
  - Asserts working tree is byte-for-byte identical via filesystem snapshot
  - Verifies `git status --porcelain` is clean against a real commit

## Implementation Details

- Reverts are stored on `Plan` rather than per-action to survive action consolidation (e.g., scaffold collapsing multiple Modify actions into one)
- Cargo dependency/feature reverts only apply once `owner_dir` has no other resource files, allowing sibling resources to keep shared dependencies
- All removal transforms are idempotent — a revert whose target is already absent returns content unchanged, enabling safe re-runs
- Migration directories matched by suffix since destroy recomputes the plan with a fresh timestamp
- Destroy never touches the database itself, only removes unapplied migration directories

https://claude.ai/code/session_01KBKktGM5AKFJ5VJbNx1WCx

### DIFF
--- a/autumn-cli/src/generate/admin.rs
+++ b/autumn-cli/src/generate/admin.rs
@@ -220,9 +220,16 @@ pub fn plan_admin_with_options(
 /// depended on the model's content.
 ///
 /// # Errors
-/// Returns [`GenerateError`] when `project_root` isn't a valid project.
+/// Returns [`GenerateError`] when `project_root` isn't a valid project, or
+/// `name` fails validation.
 pub fn plan_admin_destroy_fallback(project_root: &Path, name: &str) -> Result<Plan, GenerateError> {
     ensure_project_root(project_root)?;
+    // Unlike `plan_admin_with_options`, this path has no model-file-exists
+    // check to incidentally narrow what `name` can be — it's reached
+    // precisely because that file is gone — so an invalid name (path
+    // separators, `..`) must be rejected explicitly before it's used to
+    // build filesystem paths (issue #1048 PR review).
+    super::model::validate_resource_name(name)?;
 
     let snake_name = snake(name);
     let mut plan = Plan::new(project_root);
@@ -2113,5 +2120,21 @@ pub struct Account {
         let tmp = TempDir::new().unwrap();
         let err = plan_admin_destroy_fallback(tmp.path(), "Post").unwrap_err();
         assert!(matches!(err, GenerateError::NotInProject));
+    }
+
+    #[test]
+    fn plan_admin_destroy_fallback_rejects_path_traversal_in_name() {
+        // issue #1048 PR review: unlike `plan_admin_with_options`, this
+        // fallback has no model-file-exists check to incidentally narrow
+        // `name` — it's reached precisely because that file is gone — so a
+        // name containing path separators must be rejected explicitly
+        // before it's used to build filesystem paths for deletion.
+        let tmp = TempDir::new().unwrap();
+        fs::write(tmp.path().join("Cargo.toml"), "[package]\nname=\"x\"\n").unwrap();
+        let err = plan_admin_destroy_fallback(tmp.path(), "../../foo").unwrap_err();
+        assert!(
+            matches!(err, GenerateError::InvalidName(..)),
+            "expected InvalidName, got {err:?}"
+        );
     }
 }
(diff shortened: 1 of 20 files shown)


madmax983:
Re: "Revert scaffold-added tokio test dependency" (scaffold.rs#L414-L417) — investigated but deliberately **not** implementing this one as suggested.

`templates/Cargo.toml.tmpl` (the real `autumn new` template) ships `tokio = { version = "1", features = ["rt", "macros"] }` verbatim — byte-for-byte identical to what `ensure_dev_dependency_tokio_test_features` would insert fresh for a hand-rolled project missing it. That makes this edit fundamentally different from the other Cargo-editing reverts fixed in this PR (`mail`/`ws`/schema tables/etc.): there's no way to distinguish "freshly added by this generate call" from "already shipped by the template" using either content verification or a whole-project usage scan, because in the common single-scaffold-project case both signals agree with the wrong answer.

I implemented the fix (owner_dir-gated + a `#[tokio::test]`-usage whole-project scan, mirroring the `ws`/SSE fix above) and it broke two passing round-trip tests: it deleted the tokio dev-dependency that predates `generate scaffold` entirely, in the same way the earlier "pre-existing OAuth stub" and "pre-existing schema table" bugs did — except for those, content-verification or scan-based disambiguation actually works, since those have reliable API-usage markers that don't coincide with framework boilerplate. Tokio's `rt`/`macros` dev-features don't have an equivalent — any project with `#[tokio::test]` anywhere satisfies the "still needed" check, and any project without it looks identical whether the template shipped it or scaffold just added it.

Given issue #1048 explicitly scopes out a persisted undo log (which is what would be needed to reliably know "did *this* generate call insert this line"), leaving this specific edit un-reverted is the safer choice — it leaves a harmless extra dev-dependency line behind in the rare hand-rolled-project case, rather than risking silently deleting a real dependency other tests need in the common case. Happy to revisit if there's a cleaner way to disambiguate provenance here.

---

PR #1649, opened 2026-07-08
fix(jobs): close first-init race on the global job client OnceLock

madmax983:
## Summary

Supersedes #1491 (closed as conflicted); this is a clean re-application of the same fix on current `trunk-dev`.

`init_global_job_client` and `clear_global_job_client` used a get-then-set pattern on the `GLOBAL_JOB_CLIENT` `OnceLock`: check `get()`, and if the slot was unset, call `set()`. Two threads racing the very first install/clear could each observe the slot empty; the loser's `set()` failed silently and its value was dropped. A job runtime that had just installed its client could then be invisible to `global_job_client()` — free-function enqueues and `#[job]` handlers would see no runtime, silently routing work to the void (or, in tests, leaking another test's mock client state).

## Fix

Both functions now materialize the slot exactly once via `OnceLock::get_or_init` with an empty `RwLock` and perform the install/clear through the lock, so every write lands regardless of interleaving.

## Differences from #1491

- The loom test scaffolding from the original PR is intentionally dropped (it could not build in this workspace due to `tokio::net` features pulled in by `axum`/`tokio-tungstenite`).
- Replaced with a std-thread regression test, `job::tests::global_job_client_survives_concurrent_init_and_clear`, which hammers concurrent init/read/clear across threads and asserts a subsequent install is always observable.

## Testing

- `cargo fmt --all` and `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `cargo test --workspace`: 1036 passed, 91 ignored, and the new regression test passes. One pre-existing trybuild target (`integration::compile_fail::compile_pass_tests`) failed locally with `couldn't create a temp dir: No such file or directory` in a shared/contaminated target dir — environmental, unrelated to this change; deferring to CI as the authority on that target.

- Changelog entry added under `## [Unreleased]`.

https://claude.ai/code/session_01M4pc9qsMGH9QaxcCw5XAHi

---
_Generated by [Claude Code](https://claude.ai/code/session_01M4pc9qsMGH9QaxcCw5XAHi)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -544,6 +544,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   previously gitignored) so builds are reproducible and dependency updates
   are reviewable.
 
+### Fixed
+
+- **jobs:** fixed a first-initialization race in the process-global job
+  client (supersedes #1491): `init_global_job_client` /
+  `clear_global_job_client` used a get-then-set pattern on the backing
+  `OnceLock`, so two threads racing the very first install/clear could have
+  one side's `OnceLock::set` lose and be silently dropped — leaving a job
+  runtime that had just installed its client invisible to `global_job_client()`
+  (free-function `enqueue` and `#[job]` handlers would see no runtime). Both
+  functions now use `OnceLock::get_or_init` so the slot is created exactly
+  once and every install/clear lands through the `RwLock`.
+
 ## [0.6.0] - 2026-06-30
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -554,7 +554,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   runtime that had just installed its client invisible to `global_job_client()`
   (free-function `enqueue` and `#[job]` handlers would see no runtime). Both
   functions now use `OnceLock::get_or_init` so the slot is created exactly
-  once and every install/clear lands through the `RwLock`.
+  once and every install/clear lands through the `RwLock`. Both functions now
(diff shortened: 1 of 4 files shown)


madmax983:
Heads-up from the #1491 reconciliation — coordinating so this push isn't a surprise. I pushed `d4b22fd` to this branch with two small, additive, easily-reversible changes on top of the fix:

1. **Loom model-check** — new `autumn/tests/integration/chaos_job_client_loom.rs` (registered in `tests/integration/mod.rs`), a plain `#[test]` in the consolidated `integration_tests` target, matching the repo's existing `chaos_*_loom` convention (no `cfg(loom)`, no `--cfg loom`). loom 0.7.2 has no `OnceLock` and can't instrument `std::sync::OnceLock`, so it model-checks the get-then-set vs `get_or_init` pattern directly and asserts every write lands in the single canonical slot. It passes on the current `get_or_init` fix; set `JOB_CLIENT_LOOM_SHOW_BUG=1` to run the old get-then-set pattern and loom finds the lost-update interleaving.
2. **Poison-lock recovery** — `init_global_job_client` / `clear_global_job_client` now take the write guard via `lock.write().unwrap_or_else(PoisonError::into_inner)` instead of `if let Ok(...)`, so a poisoned lock no longer silently skips the write. This closes two gemini-code-assist review comments on #1491 that were otherwise unaddressed here. (`clear` uses an inner block so the guard drops before `clear_global_tracking_store()`, needed for clippy `significant_drop_tightening`.)

clippy `--all-targets -D warnings` is clean and the workspace builds. Run the loom test with `cargo test -p autumn-web --test integration_tests chaos_job_client_loom`. Happy to drop either change if you'd rather keep this PR minimal.

---

PR #1660, opened 2026-07-09
Add first-class Atom/RSS feed renderer (#1045)

### LINKED ISSUE #1045: Add a first-class Atom/RSS feed renderer for content apps
## Problem (the "So What?")
Autumn ships **blog**, **wiki**, and **reddit-clone** examples — the canonical content-publishing and link-aggregation apps — yet there is no way to expose an RSS or Atom feed. A blog without a feed is half a blog: readers can't subscribe in their feed reader, and aggregators (Reddit-style crawlers, planet aggregators, podcast directories, Slack/Discord RSS bots) can't syndicate the content. Today a framework user must hand-roll XML string concatenation, get the `Content-Type`, escaping, and date formats right by trial-and-error, and re-implement conditional GET so feed pollers don't hammer the DB. The substrate already exists in-tree (`etag.rs` for conditional GET, the `markdown` module for content bodies), but nothing ties it into a feed response.

## User Story
As a framework user building a content app on autumn-web, I want to expose an Atom/RSS feed of a collection of records in a few lines from a `#[get]` handler, so that readers can subscribe and aggregators can syndicate my content without me writing or debugging XML by hand.

## Acceptance Criteria
- [ ] A feed response type (e.g. `Feed` / `AtomFeed`) is constructible from channel metadata (title, site link, self link, author, updated-at) plus an iterator of entries (id, title, link, summary/content, published, updated).
- [ ] The type implements the standard autumn response trait (`IntoResponse`) so a `#[get("/feed.xml")]` handler can `return feed` directly.
- [ ] Emitted XML sets the correct `Content-Type` (`application/atom+xml` for Atom; `application/rss+xml` for RSS) and is UTF-8 with all text fields XML-escaped (no injection via `<`, `&`, `]]>` in titles/bodies).
- [ ] Output passes the W3C Feed Validator for a representative blog feed (10 entries) with zero errors.
- [ ] Conditional GET works: the response carries `Last-Modified`/`ETag` derived from the newest entry, and a matching `If-None-Match`/`If-Modified-Since` request returns `304 Not Modified` with no body — reusing the existing `etag` machinery, not a parallel impl.
- [ ] At least one example app (the `blog` example is the natural home) mounts a working feed route, and its README/walkthrough shows the < 20-line wiring.
- [ ] A unit/integration test asserts the validated XML shape and the 304 path; doctest on the public constructor compiles.

## Success Metric
Time-from-zero to a validating subscribable feed on an existing list endpoint is **< 10 minutes and < 20 LOC** in the handler, and a feed poller issuing repeat requests gets a `304` (zero DB rows read) on the unchanged-content path.

## Out of Scope
- `sitemap.xml` generation (separate SEO-discovery slice; file as a follow-up if demand appears).
- `robots.txt` scaffolding (trivial static asset; not feed-shaped).
- WebSub/PubSubHubbub push notification of feed updates.
- JSON Feed (`application/feed+json`) — Atom + RSS 2.0 cover the overwhelming majority of readers; revisit only on request.
- Auto-deriving feeds from `#[model]`/`#[repository]` by convention — this slice is the explicit, hand-wired renderer; convention-based feeds are a possible later layer.

## Gap Analysis
- **Django** ships `django.contrib.syndication` — a first-class `Feed` class you subclass with `items()`/`item_title()`/`item_description()`, plus a separate `django.contrib.sitemaps`. Syndication is considered batteries-included.
- **Rails** ships the `atom_feed` builder helper and `to_xml`/Atom view templates; emitting a feed from a controller is a few lines in a `.atom.builder` view.
- **Phoenix** has no built-in feed but the community reaches for a small helper; it's a known papercut.
- **axum / actix-web / rocket** offer nothing — you bring a crate (`rss`, `atom_syndication`) and wire content-type, escaping, and caching yourself. This is exactly the "wire it yourself" tax autumn exists to remove.
- **autumn-shaped answer:** a typed `Feed` response that implements `IntoResponse` and composes with the existing `etag` conditional-GET layer, fed by a normal `#[repository]` query. It matches autumn's "return a typed thing from a `#[get]` handler" idiom (like Maud markup and the JSON responders) rather than introducing a templating detour. The blog/wiki/reddit-clone examples make the value self-evident — the framework that demos a blog should make that blog subscribable.

## Complexity Tier
**S** — single crate (`autumn`), one new public response type + a thin reuse of `etag.rs`; additive public surface (no breaking change, minor version bump). One example app gains a route. No migration burden.

madmax983:
## What changed
**Before:** exposing an RSS/Atom feed from an autumn-web content app meant hand-rolling XML string concatenation and getting content-type, escaping, dates, and conditional-GET right by trial and error.
**After:** a `#[get("/feed.xml")]` handler can return a typed `Feed` in a few lines. `Feed::atom(...)` / `Feed::rss(...)` take channel metadata plus an iterator of `FeedEntry` values; the type implements `IntoResponse`, emitting UTF-8 XML with the correct `application/atom+xml` / `application/rss+xml` content type and every text field XML-escaped. `Feed::conditional(&headers)` reuses the existing `etag` layer so feed pollers get a `304 Not Modified` on unchanged content.

The `blog` example mounts a working `/feed.xml` route (<20 lines) and its README shows the wiring.

## How
- New module `autumn/src/feed.rs` (`Feed`, `FeedEntry`, `FeedFormat`); registered in `autumn/src/lib.rs`.
- Atom 1.0 and RSS 2.0 renderers with a shared XML escaper (neutralizes `<`, `&`, `"`, `'`, and `]]>`).
- Conditional GET via `feed::Feed::conditional` → `etag::fresh_when(...).last_modified(...)`; ETag derived from the newest entry timestamp + entry count.
- Consolidated integration tests (`autumn/tests/integration/feed.rs`): Atom/RSS shape, escaping/no-injection, content-type, and the 304 / 200-with-ETag paths; doctests on the public constructors.
- `examples/blog`: new `/feed.xml` route + README walkthrough. CHANGELOG `### Added` bullet. Plugin skill docs updated (`skills/autumn-web`).

Closes #1045

---
_Generated by [Claude Code](https://claude.ai/code/session_01H993JQQA3ncswrDMHMmaBa)_


madmax983:
Both P2s addressed in ad90447:

- `Feed::etag()` now derives a strong validator from the exact rendered feed bytes (the `etag` module's SHA-256 `IntoETag`), so any field or content edit — including two edits within the same whole second, or edits that don't move the newest timestamp — produces a different ETag. No more collisions on `(whole-second ts, entry count)`.
- The empty / no-timestamp `<updated>` / `lastBuildDate` fallback is now a fixed Unix-epoch value instead of `Utc::now()`, so the body is deterministic and conditional GET can't return 304 for a representation that changed.

Added regression tests: `etag_changes_when_content_changes_within_same_second` and `empty_feed_renders_deterministically`.

---

madmax983:
Fixed in 0d54741: the strong body-hash ETag is now the sole freshness validator — `conditional` drops `If-Modified-Since` from the freshness check so a coarse whole-second IMS can't 304 a body that changed within the same second (or via a feed-level edit that didn't move an entry timestamp). `Last-Modified` is still emitted for caches; only a matching `If-None-Match` yields 304. Added regression test `conditional_get_ignores_if_modified_since_only`.

---

madmax983:
Fixed in d45c6d7832b4ecb5db4bffdcb951eb5115930c66: `escape()` now drops characters outside the XML 1.0 `Char` set (control chars below space other than tab/newline/CR) before escaping, so an illegal control character in a title/body can no longer make the feed non-well-formed. Added regression test `drops_xml_illegal_control_characters`.

---

madmax983:
Fixed in 172ca6c: `render_atom` now always emits a feed-level `<author>` (defaulting to the feed title when `.author()` isn't set), so an Atom feed built without an explicit author stays RFC 4287-valid. Added regression test `atom_defaults_feed_author_to_title_when_unset`.

---

madmax983:
Fixed in f90d289: the feed body-hash ETag is now weak (`W/"..."`), matching what `EtagLayer` does for body-derived tags, so it stays a valid validator even when the framework compression layer gzip/brotli-encodes the response. Weak `If-None-Match` comparison still yields 304 on match. Added regression test `feed_etag_is_weak`.

---

PR #1666, opened 2026-07-09
Add `autumn i18n check` to catch missing and unused translation keys (#1252)

### LINKED ISSUE #1252: Add `autumn i18n check` to catch missing and unused translation keys
## Problem (the "So What?")

autumn ships first-class Fluent i18n (#503): `Locale::t("key")`, `t_with(...)`, and `i18n/<locale>.ftl` files. But the *only* signal that a translation key is missing is a **runtime** miss — `Bundle::miss_count()` increments and a log warning fires when a real user hits the page (autumn/src/i18n.rs:438,472). There is no build-time or CLI way to ask "which keys are referenced in my code but not in `i18n/fr.ftl`?" or "which keys did I translate into `en` but forget in `de`?". A developer adding a second locale ships untranslated UI and finds out from a production bug report. `autumn doctor` only validates that `[i18n]` is a known `autumn.toml` section (doctor.rs:53) — it does not look at keys at all.

## User Story

As a framework user shipping a multi-locale app, I want a CLI command that compares the translation keys my code calls against the keys defined in each `i18n/<locale>.ftl`, so that I catch missing and untranslated strings in CI instead of from a production user.

## Acceptance Criteria

- [ ] `autumn i18n check` exists as a CLI subcommand (autumn-cli only).
- [ ] It scans the project's `**/*.rs` sources for string-literal keys passed to `t(...)` and `t_with(...)` (e.g. `t("nav.home")`), building the set of *referenced* keys.
- [ ] Non-literal / dynamically-built keys (e.g. `t(&format!(...))`) are not analyzable; the command lists them as "dynamic — not checked" rather than silently ignoring or falsely flagging them.
- [ ] It loads every `i18n/<locale>.ftl` via the existing `Bundle` loader and reports, per locale:
  - **Missing**: keys referenced in code but absent in that locale's resolved fallback chain (consistent with `resolved_fallback_chain()` — a key supplied by a fallback locale is NOT reported missing).
  - **Untranslated**: keys present in the default locale but absent in a non-default locale.
  - **Unused**: keys defined in a `.ftl` with no matching call site in code.
- [ ] Exit code is non-zero when any locale has **Missing** keys (the correctness failure); **Untranslated**/**Unused** are warnings by default and become errors under `--strict`.
- [ ] `--format json` emits machine-readable output so `autumn check` / CI can consume it.
- [ ] Running it on the `blog` example (or a fixture app) with a deliberately deleted key produces a non-zero exit naming the offending key and locale.

## Success Metric

A locale that passes `autumn i18n check` in CI ships with **0** runtime translation misses for static-key call sites — i.e. `Bundle::miss_count()` stays 0 across the example E2E suite (#1192) for every shipped locale. DX: time-to-detect a missing key drops from "production user report" to a sub-2s local CLI run.

## Out of Scope

- Auto-writing stub entries into `.ftl` files (a follow-up `autumn i18n sync`/`--write`). This slice is read-only diagnostics.
- Extracting keys from non-Rust sources (standalone HTML, JS) or resolving dynamic/computed keys.
- Validating Fluent *placeable arguments* (that `t_with` args match the `{ $name }` variables in the message) — valuable, but a separate slice.
- Pluralization / CLDR plural-category coverage checks.
- Any change to the runtime i18n public API or `.ftl` on-disk format.

## Gap Analysis

- **Rails** ships this via the `i18n-tasks` gem: `i18n-tasks missing`, `unused`, `normalize`, with CI-friendly exit codes — widely considered table stakes for a localized Rails app.
- **Django**: `makemessages` extracts `_()`/`{% trans %}` call sites into `.po`, `compilemessages` validates; empty/fuzzy `msgstr` surface untranslated strings.
- **Phoenix**: `mix gettext.extract` + `gettext.merge` extract from templates and merge into locale files, flagging fuzzy/obsolete entries.
- **Next.js / i18next**: `i18next-parser` statically extracts keys from code into resource files.
- **axum / actix / rocket**: nothing — i18n is bring-your-own-crate, so there is no framework-level key tooling at all. autumn is already *ahead* by shipping Fluent + a `Locale` extractor; the missing rung is the tooling layer that makes a multi-locale app maintainable. The autumn-shaped answer is a `autumn i18n check` subcommand that reuses the existing `Bundle` loader and `resolved_fallback_chain()` semantics, mirroring how `autumn doctor`/`autumn check` already gate other invariants.

## Complexity Tier

**S–M** — One new file in `autumn-cli` (`i18n.rs`) plus `main.rs` subcommand wiring. Reuses the existing `Bundle`/`I18nConfig` loader from `autumn-web`. The only real work is the literal-key scanner over `*.rs` (syn or a scoped regex). No `autumn-web` public-API change, no `.ftl` format change, no migration, no minor-version bump required.

madmax983:
Closes #1252

## Before

autumn ships first-class Fluent i18n (`Locale::t("key")`, `t_with(...)`, the `t!` macro, and `i18n/*.ftl` files), but the **only** signal that a translation key is missing from a locale is a *runtime* miss: `Bundle::miss_count()` increments and a log warning fires when a real user hits the page. A developer adding a second locale ships untranslated UI and finds out from a production bug report. `autumn doctor` only validates that `[i18n]` is a known config section — it never looks at keys.

## After

A new read-only CLI diagnostic:

```
autumn i18n check              # exit 0 clean, non-zero when a locale is missing a referenced key
autumn i18n check --strict     # untranslated/unused also fail
autumn i18n check --format json  # machine-readable, for autumn check / CI
```

It scans the project's `**/*.rs` for string-literal keys passed to `t!(...)`, `.t(...)`, and `.t_with(...)`, loads every `i18n/<locale>.ftl` through the existing `Bundle::load_from_dir` loader, and reports per locale:

- **Missing** — referenced in code but absent from that locale's resolved fallback chain (a key supplied by a fallback locale is *not* reported missing, consistent with `resolved_fallback_chain()`). This is the correctness failure and drives a non-zero exit.
- **Untranslated** — present in the default locale but absent from a non-default one. Warning; error under `--strict`.
- **Unused** — defined in a `.ftl` with no matching call site. Warning; error under `--strict`.

Keys built at runtime (e.g. `t(&format!("nav.{section}"))`) can't be analyzed statically, so they are listed as "dynamic — not checked" rather than silently ignored or falsely flagged.

Example, run against `examples/blog` with a key deleted from every locale:

```
locale `en (default)`:
    Missing:
      ✗ nav.home
Result: FAIL — missing keys in one or more locales.
$ echo $?
1
```

## How

- **`autumn-cli/src/i18n.rs`** (new): the whole command.
  - Scanner runs over the **raw token tree** (`proc_macro2::TokenStream`), not the parsed AST. This is the key design decision: nearly all view-layer `t!` calls live *inside* `maud::html! { ... }`, and a `syn::Visit` walk stops at the outer `html!` macro and never sees them. Descending the token tree finds `t!` / `.t` / `.t_with` call sites wherever they are nested. The small argument list of each detected call is then parsed with `syn` to classify the key literal.
  - `t!(locale, "key", …)` → key is arg 1; `receiver.t("key")` / `.t_with("key", …)` → arg 0; `Locale::t(&loc, "key")` → first string-literal arg.
  - Loads the bundle via `Bundle::load_from_dir` so a `.ftl` syntax error or a missing default locale fails the command exactly as it fails app startup. `.ftl` message keys are enumerated with a small parser that mirrors the runtime `parse_ftl` grammar. No `autumn-web` public-API change.
- **`autumn-cli/src/main.rs`**: `I18n { action: I18nSubcommands }` subcommand + dispatch, plus clap parse tests.
- **`autumn-cli/Cargo.toml`**: enable the `visit`-free scan by adding `quote` and `proc-macro2` (with `span-locations` for dynamic-site line numbers); `syn` gains the `visit` feature (unused now but harmless — actually the token walk needs no visit; the dep line is minimal).
- Tests: 8 unit tests (scanner forms, dynamic keys, fallback suppression, strict-mode exit codes, `.ftl` parsing) + 7 consolidated integration tests (`tests/integration/i18n_check.rs`) exercising the real binary against temp fixtures — clean/missing/unused/untranslated/dynamic/json/no-i18n cases.
- Docs: documented under the plugin `skills/autumn-web/SKILL.md` CLI section and a `## [Unreleased]` CHANGELOG bullet.

https://claude.ai/code/session_015B865kGXNKZyz2FtnNYuVq

---
_Generated by [Claude Code](https://claude.ai/code/session_015B865kGXNKZyz2FtnNYuVq)_


madmax983:
Thanks for the thorough review pass. We've addressed the findings that could cause a false CI failure or miss a Missing translation on realistic Autumn call shapes (18 fixes across literal-key extraction, dynamic-key prefixing, nested calls, associated-call detection, generic/turbofish argument splitting, and profile/override config resolution).

The remaining suggestions concern increasingly exotic key-expression shapes (e.g. a string literal with chained method calls, non-`format!` key construction, unusual macro-transcriber interactions). `autumn i18n check` is by design a best-effort token/grammar heuristic, not a type-resolved analysis: any key it can't statically resolve to a literal is treated as *dynamic — not checked*, so these shapes are never falsely flagged as Missing/Unused and cannot break CI. Rather than special-casing each shape, we've documented these intentional limits under "Known heuristic limits" in the command's module rustdoc (`autumn-cli/src/i18n.rs`) and the plugin skill doc (`skills/autumn-web/SKILL.md`). Further exotic-shape findings in this class will be treated as documented limitations rather than individual code changes.

---

PR #1820, opened 2026-07-12
fix(macros): order cascade restrict probes before before_delete hooks (#1800)

madmax983:
Closes #1800.

Three narrow correctness edge cases in the `destroy` cascade codegen
(`autumn-macros/src/repository.rs`), all follow-ups to #1789. They share one
invariant: **no `before_delete` hook at any tier fires until every reachable
`restrict` probe has passed.** The DB always rolls back on the 409, so the real
exposure is a non-transactional `before_delete` side effect firing before an
abort, plus a dangling FK in the mixed soft/hard case.

## Case 1 — mixed soft/hard-delete diamond (dangling FK)

**Before:** the traversal dedup set `__deleted` was populated for BOTH
soft-deleted (`deleted_at`-only) and physically deleted rows. In a diamond where
a soft parent destroys a soft child and a hard child that both reference the same
grandchild, the soft branch marked the grandchild in `__deleted`, so the hard
branch skipped it — leaving the grandchild physically present with an FK pointing
at the now-deleted hard row (immediate FK violation / dangling FK).

**After:** a row is recorded in `__deleted` only when it was PHYSICALLY deleted.
A soft-deleted row stays reachable, so the hard branch still removes it. A
redundant SOFT re-visit of an already-soft-deleted row (a pure-soft diamond) is
skipped so its hook does not re-fire.

## Case 2 — sibling restrict pre-scan

**Before:** the grandchild `restrict` probe ran inside the per-child mutating
loop. With multiple destroy siblings, an earlier sibling's `before_delete` fired
and only then did a later sibling's restrict grandchild return the 409. The tx
rolled back but the hook's non-transactional side effect did not.

**After:** the Destroy arm pre-scans the restrict grandchildren of EVERY selected
child id in a read-only pass before any child `before_delete` fires. Applies to
both the compile-time and the runtime (model-declared) grandchild walks.

## Case 3 — parent restrict before parent hook

**Before:** `assemble_cascade_body` emitted the parent `before_delete` before
`#cascade_calls`, where the Phase-1 restrict probe returns the 409. So a parent
with both a hook and a live `restrict` dependent fired its hook, then aborted.

**After:** the parent restrict probe runs first, then the parent `before_delete`,
mirroring the child-tier ordering. Applies to both cascade branches
(model-declared `#[has_many(dependent = restrict)]` and the repository-attribute
`dependent(...)` form), since both assemble via `assemble_cascade_body`.

## How

All three fixes live in `autumn-macros/src/repository.rs`:

- Case 1: a new `destroy_deleted_mark` token gates the `__deleted.insert` on
  `!__parent_soft` for soft-delete children (unconditional for non-soft), plus a
  `destroy_soft_revisit_skip` guard that skips a redundant soft re-visit.
- Case 2: the Destroy arm's per-child loop is split into a read-only restrict
  pre-scan pass (`for __row in &__ids`) and the mutating pass; the runtime
  grandchild-mutating block re-binds `__autumn_gc_rt_deps` independently.
- Case 3: `assemble_cascade_body` now takes the cascade split into
  `cascade_restrict_calls` (emitted before `#parent_before_delete`) and
  `cascade_mutating_calls` (after), with both call sites updated.

## Tests

- Three Docker-gated integration tests in
  `autumn/tests/integration/repository_dependent_destroy.rs` (a mixed soft/hard
  diamond; `Post → Comment(destroy, hooked) → Reply(restrict)` with two comment
  siblings; a hooked parent with a live restrict dependent), each asserting the
  409 and that the relevant `before_delete` counter stayed at 0 / no dangling FK.
- Three macro string-scan unit tests asserting the emitted ordering/gating.

## Gates

- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean (also
  `-p autumn-web -p autumn-macros --all-targets --features db`)
- `cargo test -p autumn-macros` — 472 passed
- `cargo check -p autumn-web --tests --features db` — clean; the new integration
  tests compile and are collected (they require Docker/testcontainers, so they
  run in CI, matching the repo convention for these graph tests).

https://claude.ai/code/session_016bcBbx7WE5P1J9WowpB8T9

---

### DIFF
--- a/autumn-macros/src/model.rs
+++ b/autumn-macros/src/model.rs
@@ -485,9 +485,12 @@ fn emit_dependents_impl(model_ident: &syn::Ident, assocs: &[Association]) -> Tok
                 __parent_id: i64,
                 __parent_soft: bool,
                 // Codex round-5-B: the active recursion path (cycle-break) and the
-                // monotonic actually-deleted set, threaded through separately.
+                // monotonic HANDLED set (soft OR physical), threaded through
+                // separately. #1800 case 1: `__physical` additionally tracks only
+                // physically-removed rows for the diamond revisit-skip.
                 __path: &'__a mut ::std::collections::HashSet<(&'static str, i64)>,
                 __deleted: &'__a mut ::std::collections::HashSet<(&'static str, i64)>,
+                __physical: &'__a mut ::std::collections::HashSet<(&'static str, i64)>,
             ) -> ::autumn_web::repository::RuntimeDependentCascadeFuture<'__a> {
                 ::std::boxed::Box::pin(async move {
                     // Own the child repository inside the async block so the
@@ -505,6 +508,7 @@ fn emit_dependents_impl(model_ident: &syn::Ident, assocs: &[Association]) -> Tok
                             __parent_soft,
                             __path,
                             __deleted,
+                            __physical,
                         )
                         .await
                 })
--- a/autumn-macros/src/repository.rs
+++ b/autumn-macros/src/repository.rs
@@ -1461,10 +1461,10 @@ pub fn repository_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
         // broadcasts. For a model with no dependents this is a cheap no-op that
         // iterates an empty `&[]`, so the plain path stays behavior-equivalent.
         // Codex round-5-A: same restrict/mutating split as the compile-time path,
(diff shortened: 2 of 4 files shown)


madmax983:
@codex review

---

madmax983:
@codex review

---

madmax983:
@codex review

---

Issue #1130, opened 2026-06-07
Make scaffolded views reuse the app's shared layout, not a stub


madmax983:
## Problem (the "So What?")
`autumn new` ships a polished, shared application shell: `pub fn layout(title, content)` in `main.rs` with a `<nav>`, the `/static/css/app.css` stylesheet link, a skip-link, a viewport meta, and a footer. But the moment you run `autumn generate scaffold post title:string`, the generated `routes/post.rs` **shadows that shell** with its own private `fn layout()` that is a bare `DOCTYPE + <title> + <body>` — no nav, no stylesheet, no skip-link, no footer — carrying the comment *"Replace with your real layout."* The result: every scaffolded page renders as an unstyled orphan that looks nothing like the app's home page, and each additional resource emits yet another divergent copy. Adding a nav item or a CSS link means hand-editing N+1 files. This is the same "primitive ships, generator doesn't wire it" pattern already flagged in #1023, #1124, and #1125 — here applied to the view shell.

## User Story
As a framework user scaffolding a CRUD resource, I want my generated pages to render inside my app's existing layout (nav, styles, footer), so that a freshly scaffolded resource looks like part of my app instead of a blank white page I have to restyle by hand.

## Acceptance Criteria
- [ ] Scaffolded route handlers wrap their content in the application's shared layout (the `pub fn layout` emitted by `autumn new`) rather than defining a private `fn layout` in the generated routes file.
- [ ] A scaffolded index/show/new/edit page served in a fresh `autumn new` app includes the same `<nav>`, `/static/css/app.css` link, skip-link, and footer as the home page — verifiable by asserting the rendered HTML contains the stylesheet link and nav markup.
- [ ] Generating two resources produces **zero** duplicated layout definitions; changing the site chrome requires editing exactly one file.
- [ ] If the target project has no shared root `layout` (e.g. an app not created by `autumn new`), the generator either materializes a shared layout once or fails with an actionable message — it must not silently regress to per-resource stubs.
- [ ] The `cargo build` of the generated app still succeeds (generator conformance gate stays green).
- [ ] The misleading "Replace with your real layout" stub comment is gone from scaffold output.

## Success Metric
Time-from-`generate scaffold`-to-styled-page drops to ~0 manual edits: a scaffolded resource is visually consistent with the app shell on first render, with **1** file owning site chrome instead of N+1.

## Out of Scope
[...]

PR #1823, opened 2026-07-12
Generate typed path helpers in scaffolds (Closes #1133)

### LINKED ISSUE #1133: Generate typed path helpers so scaffolds never hardcode URLs
## Problem (the "So What?")
Every file `autumn generate scaffold` emits hardcodes its URLs as string literals: `a href="/{plural}/new"`, `format!("/{plural}/{{}}/update", row.id)`, `Ok(redirect_to(&format!("/{plural}/{{}}", id)))`, and pagination links (see `autumn-cli/src/generate/scaffold.rs`). Rename or nest a resource's route and these links rot silently — the compiler says nothing, the broken link only shows up as a 404 at runtime. The framework already knows every route path at compile time (the `#[get("/posts/{id}")]` attributes), so users are hand-maintaining strings the tooling could own. This is the single biggest source of refactor fragility in generated apps.

## User Story
As a framework user, I want typed path helpers for my routes (e.g. `posts::show(id)` instead of `format!("/posts/{}", id)`), so that renaming or relocating a route updates every link and redirect in one place and a stale URL is a compile error, not a production 404.

## Acceptance Criteria
- [ ] `autumn generate scaffold Post ...` emits a per-resource path-helper module (e.g. `paths::posts`) exposing `index()`, `new()`, `show(id)`, `edit(id)`, `update(id)`, and `delete(id)` returning the URL strings.
- [ ] All generated index/show/new/edit templates and all create/update/delete redirects call these helpers — **zero** hardcoded route string literals remain in generated view or handler URL positions.
- [ ] Helpers accept the typed id (the model PK type, `i64`) and produce a correctly-formatted path; query-string variants (pagination `?page=&size=`) are built through the same module, not inline `format!`.
- [ ] Changing the resource's base segment requires editing exactly one constant/location in the generated module; the generated UI and redirects follow automatically.
- [ ] The generator conformance gate (CI) compiles and serves the scaffolded app using the path helpers, proving the rewritten templates still build.
- [ ] `docs/**` documents the path-helper convention and shows the before/after (string `format!` vs helper call).

## Success Metric
After scaffolding a resource, `grep` for route string literals (`format!("/`, `href="/<plural>`) in the generated view and handler URL positions returns **0**. Renaming the resource's base path is a one-line edit that updates 100% of generated links/redirects, with the broken-link surface reduced from N hand-edited sites to 1.

## Out of Scope
- Framework-wide, compile-verified path helpers derived automatically from every `#[get]`/`#[post]` attribute across hand-written routes (the Phoenix `~p`-style global registry) — that is a larger follow-up; this slice covers only generator-emitted resources and the copyable pattern they establish.
- A runtime route-name registry / reverse-routing API for dynamically resolving arbitrary named routes.
- URL helpers for plugin- or actuator-owned routes.
- Locale-prefixed or absolute (scheme+host) URL generation.

## Gap Analysis
Rails (`post_path(@post)`, `edit_post_path`, `new_post_path`), Phoenix 1.7 (`~p"/posts/#{post}"`, **compile-verified**), Django (`reverse()` / `{% url %}`), and Laravel (`route('posts.show', $post)`) all treat hand-written URL strings in views as an anti-pattern and ship first-class path helpers — it is table stakes for a batteries-included web framework. autumn currently sits with axum/actix, where URL construction in views is fully manual. Because autumn owns the scaffold generator and captures route paths in macros, the autumn-shaped answer is to start where the pain is concentrated and the win is cheapest: have the generator emit typed helpers and consume them itself, so every new app ships refactor-safe by default and developers get a pattern to extend by hand. This also de-risks downstream issues that add routes (#1021 delete, #1028 `--api`, #1026 references) — they plug into one URL convention instead of scattering more `format!` strings.

## Complexity Tier
M — autumn-cli only (generator templates + a new emitted module); no public runtime API change, no migration burden. Touches `generate/scaffold.rs` emit paths and the conformance fixture; published `autumn-web` surface is untouched.

madmax983:
Closes #1133

**Before:** `autumn generate scaffold Post ...` produced views and handlers that hardcoded every URL as string literals — e.g. a row's show link was `format!("/posts/{}", row.id)` and the update handler redirected with `Redirect::to(&format!("/posts/{}", *id))`. Renaming a resource's base segment meant hand-editing every template and redirect.

**After:** Scaffolds emit a typed path-helper module and route every URL through it. The generated routes file ends with `autumn_web::paths[ index, show, new, create, edit, update, delete ]` (plus `events` for `--live` and `validate_{field}` for `--live-validation`), producing `paths::index()`, `paths::show(id)`, `paths::new()`, `paths::edit(id)`, `paths::update(id)`, `paths::delete(id)`, `paths::create()`. Every generated view href, form action, pagination base path, SSE connect URL, inline-validation `hx-post`, and every create/update/delete redirect now calls a helper — zero hardcoded route string literals remain. Changing a resource's base segment is now a one-line edit that the UI and redirects follow automatically.

**How:** In `autumn-cli/src/generate/scaffold.rs`, `render_routes_file` appends the `autumn_web::paths[…]` invocation (consuming the framework's existing `__autumn_path_*` companions per `autumn-macros/src/paths_macro.rs`). Added `name = "new" | "edit" | "delete"` route overrides so helper short-names match the required API. Rewrote all view href/form-action/redirect positions to `paths::*` calls; pagination goes through `PagerOptions::new(&paths::index())` / `data_table.base_path(&paths::index())` (AC #3). Generator test assertions in `autumn-cli/tests/generate.rs` and `tests/integration/scaffold_validation.rs` updated to expect the helper-based output. The layout `current_path` nav hint is intentionally left a literal (it's an active-state hint, not a link; issue #1130 asserts it stays literal).

**AC coverage:** AC1 module emitted ✓; AC2 zero hardcoded URLs in views/redirects ✓; AC3 typed id + pagination via the module ✓; AC4 single-location base-segment change ✓; AC5 scaffold cargo-check conformance gates pass ✓; AC6 docs (`docs/guide/path-helpers.md`) already present.

**Gate:** `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p autumn-cli` (176 integration + unit tests pass), and the scaffold cargo-check conformance gates (`generated_scaffold_cargo_checks`, soft_delete, sharded, unique) all green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ExYPG4uKuSnFFhTmvTG9NH)_

### DIFF
--- a/autumn-cli/src/generate/scaffold.rs
+++ b/autumn-cli/src/generate/scaffold.rs
@@ -1095,7 +1095,11 @@ fn render_repository_file(
     let fragment_item_content = if api {
         "(self.id)".to_string()
     } else {
-        format!("a href=(format!(\"/{plural}/{{}}\", self.id)) {{ (self.id) }}")
+        // Link through the routes module's typed path helper so the show URL
+        // has a single source of truth (issue #1133). This fragment lives in the
+        // repository module, so it reaches `paths::show` via the fully-qualified
+        // `crate::routes::{plural}::paths` path.
+        format!("a href=(crate::routes::{plural}::paths::show(self.id)) {{ (self.id) }}")
     };
     let live_fragment_impl = if live {
         format!(
@@ -1787,7 +1791,6 @@ fn render_routes_file(
         let changeset_inputs = render_changeset_form_inputs(
             fields,
             "changeset",
-            plural,
             live_validation,
             &validated_fields,
             &reference_fields,
@@ -1809,7 +1812,7 @@ fn render_routes_file(
         let new_form_layout = format!(
             "{layout_fn}(\"New {pascal_name}\", {cp_new}{flash_arg}, html! {{\n        \
              h1 {{ \"New {pascal_name}\" }}\n        \
-             form action=\"/{plural}\" method=\"post\"{form_enctype} {{\n            \
+             form action=(paths::create()) method=\"post\"{form_enctype} {{\n            \
              (csrf_input(csrf.as_ref(), csrf_field.as_ref()))\n{changeset_inputs}            \
              button type=\"submit\" {{ \"Create\" }}\n        \
              }}\n    \
@@ -1818,11 +1821,11 @@ fn render_routes_file(
         let edit_form_layout = format!(
             "{layout_fn}(&format!(\"Edit {pascal_name} #{{}}\", *id), {cp_edit}{flash_arg}, html! {{\n        \
              h1 {{ \"Edit {pascal_name} #\" (*id) }}\n        \
-             form action=(format!(\"/{plural}/{{}}/update\", *id)) method=\"post\"{form_enctype} {{\n            \
+             form action=(paths::update(*id)) method=\"post\"{form_enctype} {{\n            \
              (csrf_input(csrf.as_ref(), csrf_field.as_ref()))\n{changeset_inputs}            \
              button type=\"submit\" {{ \"Save\" }}\n        \
              }}\n        \
-             form action=(format!(\"/{plural}/{{}}/delete\", *id)) method=\"post\" {{\n            \
+             form action=(paths::delete(*id)) method=\"post\" {{\n            \
              (csrf_input(csrf.as_ref(), csrf_field.as_ref()))\n            \
              button type=\"submit\" onclick=\"return confirm('Delete this {pascal_name}?')\" {{ \"Delete\" }}\n        \
              }}\n    \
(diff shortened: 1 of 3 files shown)


madmax983:
Thanks for the thorough pass. All 12 of the `&paths::…()` → "won't compile against `impl Into<String>`" comments are false positives — I verified against the real signatures and by compiling generated apps.

The premise is that `Redirect::to`, `PagerOptions::new`, `DataTableConfig::base_path`, and the `form::*_htmx` helpers take a generic `impl Into<String>`. They don't — all four take a concrete `&str`:

- `Redirect::to(uri: &str)` (re-export of `axum::response::Redirect`)
- `PagerOptions::new(base_path: &'a str)` — `autumn/src/ui/pagination.rs`
- `DataTableConfig::base_path(self, base_path: &'a str)` — `autumn/src/widgets.rs`
- `form::text_input_htmx(..., validate_url: &str)` / `required_text_input_htmx(..., validate_url: &str)` — `autumn/src/form.rs`

Deref coercion *does* apply here precisely because the target is a concrete `&str`, not a generic bound — `&String` → `&str` via `String: Deref<Target = str>`. (The rule about deref not applying to generic `impl Into<T>` bounds is real, it just isn't what these signatures are.) The pre-existing generator already passed `&format!("/…")` — also a `&String` — into `Redirect::to`, and the `autumn` crate docs prescribe `Redirect::to(&paths::show(id))` as the intended pattern.

Empirically: a standard `Post title:String body:Text` scaffold and a `--live-validation` scaffold both `cargo check` clean (0 errors), exercising `Redirect::to(&paths::index())`, `Redirect::to(&paths::show(*id))`, `PagerOptions::new(&paths::index())`, `.base_path(&paths::index())`, and `form::required_text_input_htmx(..., &paths::validate_title())`. Adding `.as_str()` would be a no-op, so I'm leaving the code as-is.

---

PR #1836, opened 2026-07-12
Support model-declared has_many(dependent = nullify)

madmax983:
Closes #1786.

## Before
Declaring `#[has_many(Child, dependent = nullify)]` on a `#[model]` struct failed to compile. A `nullify` child has a nullable foreign key (`post_id: Option<i64>`), but declaring *any* `#[has_many]` emits a preload loader that groups children into a `HashMap<i64, _>` keyed directly on the child's FK field. For a nullable FK that grouping is an `expected i64, found Option<i64>` type error — so model-side nullify could not even be declared, despite the runtime cascade dispatch already handling the `nullify` action. Model-side nullify was left out of the fixtures for exactly this reason.

## After
`#[has_many(Child, fk = "...", dependent = nullify)]` compiles and cascades from the model side: deleting the parent sets each child's FK to `NULL`, and the child rows survive detached. Existing non-null `#[has_many]` associations are unaffected.

## How
- Added a framework-internal `preload::FkKey` trait implemented for both `i64` and `Option<i64>`, normalizing a child's FK to `Option<i64>`.
- Rewrote the has_many preload grouping (`autumn-macros/src/model.rs`) to route the key through `FkKey::autumn_fk_key`, skipping `None`-FK (orphan / detached) children. Non-null FKs are always `Some`, so the generated grouping is behavior-identical for existing associations. The m2m/`through` loader (non-null join FKs) is untouched.
- Test fixture `ModelDepPost` now declares `#[has_many(DepBookmark, dependent = nullify)]` (nullable `post_id`) — the compile-level acceptance test — with a bumped `dependents().len()` assertion (3 → 4) and a Docker-gated runtime test asserting the model-side delete detaches bookmarks (`post_id IS NULL`, rows survive).

---

### DIFF
--- a/autumn-macros/src/model.rs
+++ b/autumn-macros/src/model.rs
@@ -749,7 +749,17 @@ fn emit_association_items(
                         let mut __groups: ::std::collections::HashMap<i64, #stored_ty> =
                             ::std::collections::HashMap::new();
                         for __child in __children {
-                            __groups.entry(__child.#fk_ident).or_default().push(__child);
+                            // Normalize the child's FK (which may be `i64` or a
+                            // nullable `Option<i64>`, as every `dependent =
+                            // nullify` child has) to an `Option<i64>` key and
+                            // drop orphan (`None`-FK, detached) children — they
+                            // belong to no loaded parent. Non-null FKs are always
+                            // `Some`, so this is byte-identical for them.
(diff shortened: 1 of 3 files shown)


madmax983:
@codex review

---

PR #1838, opened 2026-07-12
Diagnostic when repo-attribute dependent overrides model-side has_many (#1788)

madmax983:
Closes #1788.

## Before

When a repository declared `#[repository(..., dependent(...))]` **and** its model also declared `#[has_many(..., dependent = ...)]` for the same relationship, the repository-attribute path won and the model-side declaration was silently ignored. That precedence is deliberate and correct — but a user who added a model-side `dependent` got no signal that it was inert, and could wrongly believe it was active.

## After

In debug builds, when such a delete runs, a `tracing::warn!` fires naming that the model-side `#[has_many(..., dependent = ...)]` is overridden by the repository-attribute `dependent(...)` and is inert. Release builds are unaffected (zero-cost) and cascade semantics are unchanged — the repository-attribute still wins exactly as before.

## How

Compile-time detection is impossible: the repository macro and the `#[model]` macro expand in isolation, so the repository macro cannot see the model struct's `#[has_many]` attributes (it only receives the model's type path). The diagnostic is therefore runtime and debug-only.

A `cfg!(debug_assertions)`-gated `tracing::warn!` is spliced into the repo-attribute-wins `else` branch of the single-delete `delete_body` dispatch in `autumn-macros/src/repository.rs`. It consults the model's inherent `dependents()` (via the `AutumnDependents` trait, mirroring the sibling runtime branch) and warns only when that slice is non-empty. The shape follows the existing `across_tenants()` debug-warn template for consistency.

Tested via a macro string-scan unit test asserting the guard is present in the repo-attribute `delete_by_id` branch (proven RED→GREEN). The sibling "does not runtime-dispatch" test was tightened to key off the absence of the runtime cascade driver (`__autumn_rt_deps`) rather than the mere mention of `AutumnDependents`, which the diagnostic now legitimately references.

`delete_many` has an analogous dispatch; extending the same diagnostic there is a possible follow-up (out of scope for #1788, which scopes the ask to single-delete).

https://claude.ai/code/session_016bcBbx7WE5P1J9WowpB8T9

---

### DIFF
--- a/autumn-macros/src/repository.rs
+++ b/autumn-macros/src/repository.rs
@@ -9028,6 +9028,32 @@ pub fn repository_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
                 .map(&emit_cascade_call)
                 .collect();
             let cascade_calls = quote! {
+                // ── #1788: both-sites override diagnostic ────────────────────
+                //
+                // The repository-attribute `dependent(...)` above is authoritative
+                // and drives this cascade. But the model MAY ALSO declare
+                // `#[has_many(..., dependent = ...)]`; if so, that model-side
+                // declaration is silently overridden (repo-attr wins, see the
+                // precedence note above) and is inert. Emit a debug-only warn so a
+                // user does not wrongly believe the model-side `dependent` is
+                // active. Mirrors the `across_tenants()` debug-warn template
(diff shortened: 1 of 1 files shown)


madmax983:
@codex review

---

Issue #1706, opened 2026-07-10
Prove UI accessibility conformance at build time by construction


madmax983:
## Problem (the "So What?")

Autumn renders its UI from a typed Rust widget tree (`autumn/src/widgets.rs`, component stories in `autumn/src/stories/`), yet nothing verifies that the resulting UI is accessible — an interactive widget with no accessible name, a form control with no label, or a broken heading/landmark structure ships silently. Every other stack treats accessibility as a runtime concern caught late by external audits or sampled axe-core scans. Autumn already ships a build-time i18n completeness check (`autumn i18n check`, `autumn-cli/src/i18n.rs`) but has no accessibility equivalent, even though inaccessible apps are a growing legal and reputational liability (WCAG/ADA) that today compile and deploy clean.

## User Story

As a developer shipping a customer-facing autumn app, I want the build to prove my UI is accessible by construction, so that I cannot ship an interactive element without an accessible name, a form field without a label, or a broken document structure — and I never learn about it from a lawsuit or an audit months later.

## Vision (R&D)

North star: accessibility conformance becomes a compile-time invariant of the framework — an inaccessible interactive UI does not build. Because autumn's UI is a typed component tree resolved at compile time (not a runtime-rendered string template or JSX blob), the framework can prove structural WCAG properties — accessible-name presence, label association, landmark/heading soundness, focus order, contrast-token conformance — over the whole assembled page graph before a byte is served, the same way trybuild compile-fail tests already gate the router (`autumn/src/router.rs`). Over 2-3 years this makes "accessible by construction" a language-level guarantee no other stack can offer: Rails/Django/Phoenix/Next.js views are strings or JSX that only exist at render, so their a11y story is forever a sampled runtime scan; autumn's is a proof. This unlocks a category — the framework teams choose specifically because accessibility cannot regress — plus a build-emitted conformance manifest that plugs straight into autumn's existing security and i18n posture story.

## Acceptance Criteria (first slice)

- [ ] Constructing an interactive widget (button, link, input, menuitem, etc.) without an accessible name is a compile error — the widget builder makes accessible name a type-level obligation, verified by trybuild compile-fail tests mirroring the pattern in `autumn/src/router.rs`.
- [ ] Every form-control widget must be associated with a label (or an explicit `aria-labelledby` / `aria-label`); an unlabeled control fails the build.
[...]

PR #1845, opened 2026-07-13
Part of #1706: `autumn a11y verify` build-time pass for raw-html escape hatches

madmax983:
Part of #1706 (Acceptance Criterion 3 + 4). Non-scaffold follow-on to the typed accessible primitives — does **not** touch `scaffold.rs`.

## Before / After

**Before** — an alt-less `` or an unlabeled `<input>` written in raw `html!` markup compiles clean and ships silently; the type system only guards code that goes through the typed primitives.

**After** — `autumn a11y verify` fails CI (exit 1) with a WCAG-keyed report:

```
src/view.rs:3:  [WCAG 1.1.1] Serious — raw <img> has no alt attribute
    hint: use autumn_web::a11y::Img::new(src, alt) / Img::decorative(src)
```

## How

The typed `autumn_web::a11y` primitives (`Img`/`Button`/`Link`/`MenuItem`/`TextField`) already prove accessible-name obligations **at compile time** for code that uses them. This pass covers the escape hatch the type system can&#39;t see: raw markup in `maud::html! { … }` blocks. There is no walkable widget tree at runtime, so this is an honest best-effort **static** scan.

- New module `autumn-cli/src/a11y.rs`, structured after `autumn-cli/src/i18n.rs`: it scans project `.rs` files and token-walks descending into `html!` macro bodies (the same descent `autumn i18n check` uses to find `t!` calls nested in `html!`), building a `Report` with an `exit_code(strict)` and text/JSON output.
- Reuses the WCAG success-criterion numbers, axe-core rule ids, and `Severity` semantics from `autumn-cli/src/check.rs` (the runtime `autumn check --a11y` lint), applied to static source instead of rendered HTML.
- `--format json` emits the &#34;conformance manifest&#34; AC3 asks for: an array of findings keyed to WCAG SCs plus a summary.
- Exit code is non-zero when findings exist (Serious by default; `--strict` lowers the threshold to Moderate), mirroring `i18n check`&#39;s `exit_code(strict)`. This is what fails CI.
- Clap wiring in `main.rs`: an `a11y` subcommand group with a `verify` variant (`[PATH]` defaulting to `.`, `--format`, `--strict`), dispatched like `I18nSubcommands`.

## Scope note (the honest guarantee)

Typed primitives = the **compile-time proof**; this pass = the **escape-hatch net**. A typed primitive spliced into markup as `(Img::new(src, alt))` is a `(expr)` splice, not an `img` element, so it is never re-scanned or falsely flagged. Like `autumn i18n check`, the scanner reads tokens rather than a resolved AST and always errs toward *not* flagging what it cannot resolve, so it never breaks CI on a false positive: a spliced attribute/content value (`alt=(caption)`, `button { (label) }`) is treated as present, and a `<label for>`/`id` association a splice makes unresolvable suppresses the `label` finding. This is a documented best-effort static scan with a dynamic-value limitation.

## WCAG SC mapping

| Rule id       | Element                               | Condition                                                        | WCAG SC               | Severity |
| ------------- | ------------------------------------- | --------------------------------------------------------------- | --------------------- | -------- |
| `image-alt`   | ``                               | no `alt` attribute                                              | 1.1.1                 | Serious  |
| `label`       | `<input>`/`<select>`/`<textarea>`     | no matching `&lt;label for=…&gt;`, `aria-label`, or `aria-labelledby` | 1.3.1 / 3.3.2 / 4.1.2 | Serious  |
| `button-name` | `&lt;button&gt;`                            | no text content and no `aria-label`/`aria-labelledby`           | 4.1.2                 | Serious  |
| `link-name`   | `&lt;a href&gt;`                            | no link text and no `aria-label`/`aria-labelledby`              | 2.4.4 / 4.1.2         | Serious  |

## Tests

- 28 in-module unit tests over `html!` snippets: RED cases (alt-less img, unlabeled input/select/textarea, empty button, textless anchor) flagged with the correct WCAG SC; GREEN cases (img with alt / empty alt / dynamic alt, input with matching label `for`/`id`, aria-label, wrapped label, button/anchor with text or aria-label, icon button with alt image) clean; a typed-primitive splice and a fully-dynamic splice are **not** flagged; an element inside a `@if` block is still scanned.
- An end-to-end tempdir test: red project (exit 1) → fix → green project (exit 0).
- 2 clap parse tests for the new subcommand.

## Docs (AC4)

`docs/guide/accessibility.md` gains an `autumn a11y verify` section: what it checks, the WCAG SC mapping table, a `--format json` example, CI usage (exit code), and the explicit statement that typed primitives are the compile-time proof while `verify` catches raw-`html!` escape hatches.

## Gate

`cargo fmt --all -- --check`, `cargo clippy -p autumn-cli --all-targets -- -D warnings`, and the new `autumn-cli` a11y tests all pass.

## Follow-ons

- Extend the ruleset (e.g. redundant/empty `alt` heuristics, `role`/`tabindex` misuse, duplicate `id`) as further slices.
- Cross-`html!`-block `&lt;label for&gt;`/`id` resolution (associations currently scoped per block).
- Optionally surface `verify` through the `/actuator/a11y` actuator or a dedicated CI workflow step.

## Scope and limitations

- This PR adds `autumn a11y verify`, a build-time **static** scanner over raw maud `html!` bodies (WCAG rules: image-alt 1.1.1, label 1.3.1 / 3.3.2, button-name / link-name 4.1.2), emitting a WCAG-keyed report and failing CI on findings.
- It is explicitly **best-effort / advisory-grade**. It parses maud markup heuristically at the token level and deliberately errs toward **not** failing on ambiguous, dynamic, or non-maud constructs (unresolved splices, complex Rust expressions, foreign `html!` macros) to avoid false CI failures. It may therefore **miss** a defect in exotic raw markup — acceptable by design — but should not fail valid markup.
- The actual **compile-time accessibility guarantee** is the typed primitives in `autumn_web::a11y` (`Img`/`Button`/`TextField`/`Link`/`MenuItem` — accessible name is a type-level obligation proven by trybuild), landed in #1840 and #1842. Findings from `a11y verify` over raw `html!` are advisory signal, **not** the proof — consistent with #1706 scoping raw hand-written HTML as the "unverifiable, not proven" escape hatch.
- The scanner was hardened over many review rounds against maud grammar corners (struct literals, match arms, control heads, class/id shorthands, colon-qualified names, foreign JSX `html!`). Any remaining niche parser corners are a documented limitation of the best-effort net rather than blockers.

https://claude.ai/code/session_01Pf5JUv54F6Bxo8KzErU9sz

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pf5JUv54F6Bxo8KzErU9sz)_

### DIFF
--- a/autumn-cli/src/a11y.rs
+++ b/autumn-cli/src/a11y.rs
@@ -1014,31 +1014,26 @@ fn walk(
 
 fn apply_rules(el: &Element, file: &str, ctx: &LabelCtx, scan: &mut Scan) {
     match el.name.to_ascii_lowercase().as_str() {
-        "img" => {
-            // An `<img>` is named by `alt` (even empty `alt=""`, the decorative
-            // marker), or — per the axe `image-alt` checks — by an
-            // `aria-label`/`aria-labelledby`, a non-empty `title`, or an explicit
-            // presentational role. Only a bare image with none of these is flagged.
+        // An `<img>` is named by `alt` (even empty `alt=""`, the decorative
+        // marker), or — per the axe `image-alt` checks — by an
+        // `aria-label`/`aria-labelledby`, a non-empty `title`, or an explicit
+        // presentational role. Only a bare image with none of these is flagged.
+        "img"
             if !el.has_attr("alt")
                 && !el.has_aria_name()
                 && !el.has_title_name()
-                && !is_presentational(el)
-            {
-                scan.push(Rule::ImageAlt, "img", el.line, file);
-            }
+                && !is_presentational(el) =>
+        {
+            scan.push(Rule::ImageAlt, "img", el.line, file);
         }
         name @ ("input" | "select" | "textarea") => {
             check_field(el, name, file, ctx, scan);
         }
-        "button" => {
-            if !named_content(el) {
-                scan.push(Rule::ButtonName, "button", el.line, file);
-            }
+        "button" if !named_content(el) => {
+            scan.push(Rule::ButtonName, "button", el.line, file);
         }
-        "a" => {
-            if el.has_attr("href") && !named_content(el) {
-                scan.push(Rule::LinkName, "a", el.line, file);
-            }
+        "a" if el.has_attr("href") && !named_content(el) => {
+            scan.push(Rule::LinkName, "a", el.line, file);
         }
         _ => {}
     }
--- a/autumn-cli/src/a11y.rs
+++ b/autumn-cli/src/a11y.rs
@@ -408,7 +408,12 @@ enum Node {
     /// A non-empty static text node.
     Text(String),
     /// A splice `(expr)` or a control block — an unresolvable dynamic value.
-    Dynamic,
+    /// `splice` records whether it is a genuine `(expr)`/`[..]` splice (which
+    /// could render a `<label>` fragment beside a control) rather than an
+    /// `@`-control head; only the former marks a block as possibly-labeled.
+    Dynamic {
+        splice: bool,
+    },
 }
 
 /// Parse a maud markup token stream into a best-effort node list.
@@ -436,8 +441,9 @@ fn parse_nodes(trees: &[TokenTree]) -> Vec<Node> {
             }
             TokenTree::Punct(p) if p.as_char() == '@' => {
                 // Control flow (`@if`/`@for`/`@match`/`@let`): dynamic, but still
-                // descend into any block so elements inside are analyzed.
-                nodes.push(Node::Dynamic);
+                // descend into any block so elements inside are analyzed. Not a
+                // spliced fragment, so it never marks the block as possibly-labeled.
+                nodes.push(Node::Dynamic { splice: false });
                 if matches!(trees.get(i + 1), Some(TokenTree::Ident(id)) if *id == "match") {
                     // `@match` is special: the brace group after the scrutinee is
                     // an arm LIST, not markup. Parsing it as markup would treat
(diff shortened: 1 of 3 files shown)


madmax983:
### `autumn a11y verify` — review disposition

The static scanner has been hardened across ~20 review rounds against maud grammar corners, and the latest commit (`c6f6be1`) fixes the qualifying items:

- **JSX-vs-Rust comparison false-negative** — a `<`/`>` comparison in a control head (e.g. `@if count < limit { … }`) no longer causes the surrounding `html!` block to be skipped.
- **`@if match { … } { … }` control-head false-positive** — the match-condition arm list is no longer parsed as markup, so a pattern name like `input` no longer produces a bogus `label` finding.
- **Whitespace-only static name attributes** — `aria-label` / `title` / `alt` values that are only whitespace now correctly count as an empty accessible name.

Per the newly documented **Scope and limitations** (see the PR "Scope and limitations" section, the `a11y.rs` module docs, and `docs/guide/accessibility.md`): this scanner is explicitly **best-effort / advisory-grade** over raw `html!`. The compile-time accessibility *guarantee* is the typed primitives in `autumn_web::a11y` (landed in #1840 / #1842), where accessible name is a type-level obligation proven by trybuild — `a11y verify` is advisory signal for the raw-markup escape hatch, not the proof.

Remaining suggestions that are missed-check *coverage gaps* on raw markup (e.g. requiring a name on `input type="button"` / `input type="image"`) are acknowledged as candidate follow-ons within that best-effort scope — advisory net improvements rather than correctness blockers for this PR. Deliberately erring toward not flagging ambiguous, dynamic, or non-maud constructs is by design, to avoid false CI failures.

Thanks for the thorough pass — the corner-case coverage is meaningfully better for it.

---

Issue #1608, opened 2026-07-08
Provision and renew HTTPS certificates automatically via ACME


madmax983:
## Problem (the "So What?")
Spec #1603 gives the single autumn binary a direct TLS listener, but the operator still has to obtain and renew certificates by hand or with certbot cron jobs — and expired certificates are the classic self-hosted outage. Today the serve path binds plain TCP only (`autumn/src/app.rs`), rustls appears in `autumn/Cargo.toml` only for outbound client TLS (reqwest/lettre), a repo-wide search for ACME/Let's Encrypt/certbot finds nothing, and `docs/guide/deployment.md` delegates HTTPS entirely to the hosting platform's proxy. With `autumn deploy` (#1607) targeting a bare VPS, "point DNS at the box and get HTTPS forever" is the last missing link in the zero-to-production story.

## User Story
As a framework user self-hosting my autumn app on a VPS, I want the app to obtain and renew its HTTPS certificates automatically from Let's Encrypt, so that my site stays on valid TLS indefinitely without certbot, cron jobs, or a reverse proxy.

## Acceptance Criteria
- [ ] With only a domain name (and contact email) in config, a fresh app on a public host serves valid Let's Encrypt HTTPS on first startup, with HTTP→HTTPS redirect.
- [ ] Certificates renew automatically before expiry with no restart and no dropped in-flight connections; a forced near-expiry test demonstrates rotation.
- [ ] Issued certificates and account credentials persist across restarts — no re-issuance on every boot, respecting Let's Encrypt rate limits.
- [ ] The ACME directory (staging vs production) is selectable, so CI and first-time users can exercise the flow against Let's Encrypt staging without burning production quota.
- [ ] Failure modes are observable: a failed renewal surfaces in health/actuator output and through the error-reporting seam well before the certificate expires.
- [ ] `autumn doctor` (or equivalent preflight) diagnoses the common blockers: port 80/443 unreachable, DNS not pointing at this host, stored certificate expired.
- [ ] The deployment guide documents the happy path in ≤10 lines of config and the reverse-proxy opt-out.

## Success Metric
An operator goes from fresh VPS + DNS record to a valid-TLS production app with zero manual certificate steps, and a 90-day soak shows 0 expired-cert incidents with at least one automatic renewal exercised.

## Out of Scope
- The TLS listener itself (spec #1603 — this slice builds on it).
[...]

Issue #1603, opened 2026-07-08
Serve HTTPS directly: TLS config for single-binary self-hosting


madmax983:
## Problem (the "So What?")

Autumn positions itself for single-binary self-hosting — `autumn db backup`/`restore` (#1595) and `db scrub` (#1602) exist precisely for the operator running on their own box — yet the framework cannot serve HTTPS at all. `ServerConfig` in `autumn/src/config.rs` (~line 3623) exposes only `host`/`port`; the serve path in `autumn/src/app.rs` (~line 3109) binds plain TCP or a Unix socket with no TLS branch; and the only TLS mention in `docs/guide/cloud-native.md` is SMTP `tls = "starttls"` (line 222). Today the "deploy to a $5 VPS" journey ends with "now go learn nginx or Caddy," which contradicts the single-binary story and adds a second moving part that `autumn doctor`, the release image boot gate, and the graceful-shutdown/probe machinery all know nothing about.

## User Story

As a framework user self-hosting an Autumn app on a single server, I want the app to terminate TLS itself from a couple of config lines, so that I can serve `https://` with a real certificate without installing and operating a reverse proxy.

## Acceptance Criteria

- [ ] Adding a `[server.tls]` section to `autumn.toml` (certificate chain + private key file paths, PEM) makes the app serve HTTPS on the configured port; with the section absent, behavior is byte-for-byte today's plain HTTP (no new deps activated, no perf change on the hot path).
- [ ] Certificate and key files are picked up on renewal without a full restart (documented mechanism — file watch, interval, or signal), so a certbot-style renewal never drops the site.
- [ ] Misconfiguration fails fast at startup with an actionable message (missing file, cert/key mismatch, expired cert), and `autumn doctor` reports TLS problems in the same run that checks database roles today.
- [ ] `/health`, `/live`, `/ready`, graceful shutdown, the inbound request timeout, SSE, and WebSocket (`wss://`) all behave identically under TLS, verified by integration tests, not just docs.
- [ ] A deployment guide documents both supported topologies — proxy-terminated TLS (unchanged, still the multi-replica recommendation) and direct TLS for single-binary hosts — including a certbot renewal walkthrough and local dev usage with mkcert-style certificates.
- [ ] The release image boot gate (or docs-smoke) exercises an HTTPS boot with a test certificate and passes an HTTPS health check.

## Success Metric

[...]

PR #1852, opened 2026-07-13
Serve HTTPS directly via config-driven TLS listener (Part of #1603)

madmax983:
Part of #1603. First honest slice of serving HTTPS directly from the app process.

## Before / After

**Before** — there is no inbound TLS. To serve HTTPS you must front the app with a reverse proxy (nginx/Caddy/an LB) that terminates TLS. Absent that, the app speaks plain HTTP on `host:port`.

**After** — adding a `[server.tls]` section with a certificate + key makes the app terminate HTTPS itself, on the same `host:port`:

```toml
[server.tls]
cert_path = "/etc/autumn/tls/fullchain.pem"
key_path  = "/etc/autumn/tls/privkey.pem"
# optional; how often (seconds) to poll for a renewed cert on disk. Default: 60.
reload_interval_secs = 60
```

- **No `[server.tls]` → unchanged.** Plain HTTP exactly as today; zero behavior change for existing apps.
- **Fails fast on misconfig.** Missing file, unparseable PEM, a key that does not match the certificate, or an already-expired leaf certificate all abort startup with an actionable, path-named error — the app never boots into a broken listener.
- **Hot-reloads certificates.** A background task polls the cert/key file mtimes and swaps the served certificate in place when they change (e.g. after a `certbot`/ACME renewal) — no restart. A failed reload logs an error and keeps serving the previously loaded certificate; a bad renewal never breaks the listener.
- **`autumn doctor` flags problems.** Offline check: fail on a missing/invalid/expired certificate, warn within 30 days of expiry, pass otherwise.

The feature is **off by default** (`tls` cargo feature). A binary built without it that finds `[server.tls]` configured fails fast with a clear "built without the `tls` feature" message rather than silently serving plain HTTP.

## How

- **`autumn_web::tls`** (new module, behind `tls`): loads + validates the cert chain and key (`rustls`' `CertifiedKey::from_der` compares the key's `SubjectPublicKeyInfo` against the leaf, catching mismatches), rejects an expired leaf, exposes a reloadable `ResolvesServerCert` backed by an `RwLock<Arc<CertifiedKey>>` (a read-lock per handshake; a write-lock to swap), and an offline `inspect_leaf` for doctor.
- **Reuse `axum::serve` via a custom `Listener`.** `TlsListener` wraps the existing `tokio::net::TcpListener` in a `tokio_rustls::TlsAcceptor`; its `accept()` performs the rustls handshake and returns `(TlsStream<TcpStream>, SocketAddr)`. Because the peer is a real `SocketAddr`, the HTTPS serve arm is byte-for-byte the TCP arm — the **same** `ConnectInfo<SocketAddr>`, `TrustedProxiesLayer`/`ClientAddr` resolution, SSE/WebSocket streaming, request timeouts, and graceful-shutdown wiring all apply unchanged. A third `BoundListener::Tls` variant and a mirroring serve arm are the only serving-loop additions.
- **Crypto backend:** reuses the already-vendored **ring**-backed `rustls` — the same backend the outbound Postgres TLS path uses. No second TLS backend is introduced (no aws-lc-rs / native-tls / openssl).
- **Config:** `[server.tls]` is **always compiled** (so it parses regardless of features and the "no feature" error is possible), with `AUTUMN_SERVER__TLS__CERT_PATH` / `__KEY_PATH` / `__RELOAD_INTERVAL_SECS` env overrides. Nested `[server.tls]` keys are accepted by `strict_config` automatically (schema is serde-derived; the schema snapshot is updated).

## New dependencies (please review)

Both are feature-gated (`optional = true`, pulled only by the new `tls` feature) and were **already resolved in `Cargo.lock`** at these versions (via other crates), so adding them introduces no version churn:

- **`tokio-rustls` 0.26** — async accept-time rustls handshake. `default-features = false` + `ring` (no aws-lc-rs).
- **`x509-parser` 0.16** — parse the leaf certificate's `notAfter` for expired-cert fail-fast and the doctor near-expiry warning. (No default features.)

The `autumn-cli` `tls` feature (default-on, so the shipped `autumn` binary can diagnose TLS) enables `autumn-web/tls`; it pulls in no additional CLI-level crates.

## Tests

- **Unit** (`autumn/src/tls.rs`): valid load; missing file names the path; unparseable PEM; key/cert mismatch → error; expired leaf → error; `inspect_leaf` reports future and past expiry; resolver swap.
- **Config** (`autumn/src/config.rs`): `[server.tls]` TOML parse, reload-interval default/override, env-var materialization + partial override, default `None`.
- **Doctor** (`autumn-cli/src/doctor.rs`): expired → Fail, ≤30 days → Warn, healthy → Pass, invalid → Fail, feature-disabled → Warn, not-configured → Pass.
- **End-to-end** (`autumn/tests/integration/tls_serving.rs`, `#[cfg(feature = "tls")]`): a rustls client GETs `/health` over TLS and gets 200; peer `SocketAddr` connect-info flows through the TLS listener; graceful shutdown joins cleanly under TLS; a bad (plaintext) handshake does not kill the listener.

Gates run green: `cargo fmt --all -- --check`; `cargo clippy --workspace --all-targets --features tls -- -D warnings` and the plain no-feature clippy; targeted `-p` tests for all of the above; default-feature `cargo check` for both crates.

## Scope / Limitations / Follow-ups

This slice deliberately does **not** do (tracked for follow-up):

- **Deployment guide docs** (direct-vs-proxy topology, `certbot`/`mkcert` walkthroughs) — a docs session owns these.
- **Release-image HTTPS boot gate** in CI — deferred.
- **wss (WebSocket-over-TLS) integration coverage** — the listener yields the same `ConnectInfo<SocketAddr>` stream as plain TCP so a `ws` upgrade rides the identical service; a full tokio-tungstenite-over-rustls client is disproportionate here and is left as a clearly-marked `TODO(#1603 follow-up)`.
- **#1608 active-probe checks** (port 80/443 reachability, DNS) are intentionally out of this doctor slice.
- **In-accept handshake head-of-line note:** the rustls handshake runs inline with `accept`, so one slow handshake briefly delays new accepts. Acceptable for this slice; a future change can offload the handshake to a spawned task if it becomes a bottleneck.
- **Client-certificate / mTLS** and ALPN/HTTP-version negotiation tuning are not in scope.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LMeNyt3tNtHRvcQm9QGsZs)_


madmax983:
## Local verification (the gate)

Fresh, cold local verification of head commit `f2fcfa7f0ece34b89ea1b0fe860f40bdb0a730fb`. Per the project's "local verification is the gate" policy, this is the merge-readiness evidence.

| Gate command | Result |
| --- | --- |
| `cargo fmt --all -- --check` | clean, exit 0 |
| `cargo clippy --workspace --all-targets --features tls -- -D warnings` | clean, exit 0 |
| `cargo clippy --workspace --all-targets -- -D warnings` (no-feature) | clean, exit 0 |
| `cargo test -p autumn-web --features tls --test integration_tests tls_serving` | `test result: ok. 6 passed; 0 failed; 0 ignored` |
| `cargo test -p autumn-web --features tls --lib tls` | `test result: ok. 36 passed; 0 failed; 0 ignored` |
| `cargo test -p autumn-cli --features tls doctor` | `test result: ok. 306 passed; 0 failed; 0 ignored` |
| `cargo check -p autumn-web` (no-feature) | clean, exit 0 |
| `cargo check -p autumn-cli` (no-feature) | clean, exit 0 |

All gates PASS on a cold rebuild.

The direct-HTTPS TLS listener slice was hardened through multiple adversarial-review and Codex rounds — handshake concurrency/DoS resistance, full cert-chain parse plus lifetime validation, and recognized-key/dotenv/manifest-dir-aware doctor grading including expired and not-yet-valid detection — each landed with accompanying regression tests.

The GitHub Actions checks are currently sitting `queued` on the runner backlog (not failing). Per the project's "local verification is the gate" policy, this PR is merge-ready on the strength of the green local gates above; real CI results will still be acted on if/when they run.

---

Issue #1635, opened 2026-07-08
Let tenants connect custom domains with per-domain ACME certificates


madmax983:
## Problem (the "So What?")

Wildcard certificates (#1620) cover subdomain tenants on `*.myapp.com`, but B2B customers who bring their own domain (`app.clientco.com`) hit a wall today: subdomain tenancy resolution in `autumn/src/tenancy.rs` rejects any Host outside `tenancy.base_domain` with a 400, there is no hostname→tenant mapping anywhere in the framework (custom-domain routing was explicitly deferred in #695 and never picked up), and per-customer issuance is scoped out of parent #1608 ("multi-tenant per-customer certificate management at scale"). The server also has no inbound TLS or ACME surface yet (`autumn/src/app.rs` binds plaintext TCP/UDS only; `docs/guide/deployment.md` delegates HTTPS to the platform edge). So an operator who wants white-label custom domains — the single most-requested B2B SaaS capability — must run Caddy or buy Cloudflare for SaaS in front of autumn.

## User Story

As a SaaS operator running a multi-tenant autumn app, I want each tenant to connect their own domain and automatically get a valid, renewing certificate for it, so that I can offer white-label custom domains without operating a separate TLS proxy layer or doing per-customer cert work.

## Acceptance Criteria

- [ ] A tenant "connect your domain" journey exists end to end: the app registers a requested hostname for a tenant, autumn produces exact DNS instructions to show that tenant (CNAME to the operator's ingress hostname, or A/AAAA for apex domains), and the domain moves through queryable states — pending DNS → verified → certificate issued → active — that the app's UI can render, including the failure reason when stuck.
- [ ] Domain-verification gate: no ACME order is created until autumn has independently confirmed the hostname points at this deployment (DNS or HTTP token check). A test registering a domain whose DNS points elsewhere shows zero ACME orders and a status explaining why.
- [ ] Once a domain is verified, requests carrying that Host resolve to the owning tenant (the current 400 "Host does not match base domain" rejection in `tenancy.rs` does not apply to registered domains) and are served that domain's certificate via SNI.
- [ ] On-demand-issuance safety: issuance attempts are rate-limited per domain and globally with backoff on repeated failure, and an SNI hostname no tenant has registered is refused without ever contacting the ACME provider. The abuse posture toward Let's Encrypt (rate-limit budget, retry policy) is documented.
[...]

Issue #1620, opened 2026-07-08
Issue wildcard certificates via DNS-01 for tenant subdomains


madmax983:
## Problem (the "So What?")

Parent spec #1608 automates HTTPS via ACME but is deliberately scoped to single-hostname HTTP-01 — its Out of Scope defers wildcard certificates and DNS-01 challenges. Yet the framework already ships subdomain-per-tenant routing (`autumn/src/tenancy.rs` resolves the tenant from the Host header against `tenancy.base_domain` in `autumn/src/config.rs`) and a multi-tenant SaaS starter (`autumn-cli/src/starters/saas/autumn.toml`) — exactly the deployments a single-hostname certificate cannot serve. Today the operator who takes that tenancy support to `*.myapp.com` production hits a wall: HTTP-01 would need a separate issuance per tenant subdomain (Let's Encrypt rate limits, cold-start latency on tenant N's first request), so they end up back behind Caddy or nginx — undoing the single-binary story #1603/#1608 just built. Wildcard issuance requires proving DNS control via DNS-01, which means DNS provider API credentials — a secrets surface autumn already has first-class conventions for (the encrypted credentials store: `autumn/src/credentials.rs`, `config/credentials/<env>.toml.enc` + `AUTUMN_MASTER_KEY`).

## User Story

As a SaaS operator running a subdomain-per-tenant autumn app, I want the app to obtain and renew a wildcard certificate for my base domain via DNS-01, so that every tenant's subdomain serves valid HTTPS from the moment the tenant exists — with no per-tenant certificate work and no reverse proxy.

## Acceptance Criteria

- [ ] With a wildcard domain (e.g. `*.myapp.com`), a DNS provider name, and a credential reference in config, a fresh app obtains a valid Let's Encrypt wildcard certificate on first startup and serves HTTPS for the apex and any subdomain — no per-subdomain configuration.
- [ ] A documented initial provider set is supported — at minimum Cloudflare and AWS Route 53 — plus a documented escape hatch for any other provider (e.g. RFC 2136 dynamic updates or an external update hook), with the supported list published in the guide.
- [ ] DNS provider API tokens are supplied through the existing secrets conventions — the encrypted credentials store (`autumn credentials edit`) or an environment variable — never in plaintext `autumn.toml`; tokens never appear in logs, error messages, or actuator output.
[...]

PR #1858, opened 2026-07-13
Automatic ACME (Let's Encrypt) certificate provisioning & renewal (Part of #1608)

madmax983:
> Stacked on #1852 (the #1603 TLS listener). The PR base is set to
> `feat/1603-https-tls-listener` so the diff shows ONLY the ACME work.
> **Rebase onto `trunk-dev` once #1852 merges.**

## Before / After

**Before:** serving HTTPS meant an operator-provided certificate on disk
(`[server.tls] cert_path/key_path`, #1603) or a reverse proxy that terminates TLS.

**After:** set a domain and a contact email and the app obtains and auto-renews a
Let's Encrypt certificate itself:

```toml
[server.tls.acme]
domains = ["app.example.com"]
contact_email = "ops@example.com"
# directory defaults to Let's Encrypt STAGING on purpose (rate-limit safety)
```

- The app answers the ACME **HTTP-01** challenge on `:80` and **308-redirects** all
  other `http://` traffic to `https://`.
- `:443` binds immediately behind a self-signed **placeholder**; the real cert
  **hot-swaps** in once issued — no restart, no dropped connections.
- A background loop **renews before expiry** and is **leader-elected** across a
  fleet so only one replica orders per certificate.
- **Staging is the default** so a first run / CI can't burn the strict Let's Encrypt
  production quota. Opt into production with `directory = "production"`.

## How

- **Reuses the #1603 seams** — the issued cert swaps into the *same*
  `ReloadableCertResolver` via its existing `store()` (`RwLock<Arc<CertifiedKey>>`
  swap); `build_server_config` / `TlsListener` are reused unchanged. No parallel
  machinery.
- `instant-acme` drives the HTTP-01 order state machine; `rcgen` generates the ACME
  account key, the certificate keypair + CSR, and the placeholder cert.
- New `autumn/src/acme/` module: `challenge` (the `:80` router + `Http01Tokens`),
  `store` (an async `AcmeStore` trait + a `0600` `FsAcmeStore`; account keyed per
  directory so staging↔production re-registers cleanly), `renewal` (issuance,
  `needs_renewal`, placeholder, `AcmeStatus`/`AcmeHealthIndicator`, the loop).
- The `:80` challenge listener and the renewal loop spawn as **siblings** of the
  main server, each a `child_token()` of `server_shutdown`.
- **Config:** `TlsConfig.cert_path`/`key_path` are now `Option`; `TlsConfig::validate()`
  (called from the same pre-bind guard that rejects TLS+unix_socket) enforces exactly
  one of static-cert XOR ACME, requires paths together, requires non-empty
  domains+email in ACME mode, and rejects `*.` wildcards with a message pointing at
  #1620 (DNS-01, out of scope).
- **Observability:** an `AcmeHealthIndicator` (group `HealthOnly`, so a renewal blip
  does **not** fail `/ready`) grades `Down` only when a failure coincides with real
  expiry danger. Each renewal failure is dispatched through the registered
  `ErrorReporter` chain (Sentry/etc.) and always logged.
- **doctor:** `autumn doctor --online` (alias `--preflight`) adds ACME preflight
  probes — port 80/443 reachability and DNS-points-here for the configured domain —
  plus an always-on offline stored-cert expiry check. Network probes run **only**
  under `--online` so default `doctor` stays offline and non-flaky. Every grader is a
  pure function of injected inputs (mirroring `check_tls_impl`).

## ⚠️ New dependencies (please review)

Two new crates, both behind the off-by-default `acme` feature, both pinned
`default-features = false` + `ring`:

```toml
instant-acme = { version = "0.8.5", default-features = false, features = ["ring", "hyper-rustls"] }
rcgen        = { version = "0.14",  default-features = false, features = ["pem", "ring"] }
```

**The `default-features = false` is load-bearing:** the crates' default features pull
`aws-lc-rs`, which would violate the workspace single-crypto-backend rule (rustls 0.23
+ **ring**). Confirmed clean:

```
$ cargo tree -i aws-lc-rs -e no-dev -p autumn-web --features tls,acme
error: package ID specification `aws-lc-rs` did not match any packages
```

i.e. **aws-lc-rs is entirely absent from autumn-web's `tls,acme` graph** (it exists in
the workspace only via `autumn-storage-s3`'s AWS SDK, as allowed). The isolated graph
resolves to `ring v0.17`, `rustls 0.23`, `instant-acme 0.8.5`, `rcgen 0.14.8`,
`tokio-rustls`, `hyper-rustls`.

> **Deviation from the brief:** the brief pinned `instant-acme = "0.9"`, but 0.9 does
> not exist on crates.io (latest is **0.8.5**). Pinned 0.8.5, which supports the same
> `ring` + `hyper-rustls` feature posture. This is the only dependency deviation.

## Limitations / Follow-ups

- **DNS-01 / wildcard certificates** — out of scope; `*.` domains are rejected with a
  message pointing at **#1620**.
- **Per-tenant / SNI / DB-backed store & 1000-domain incremental load** — the
  `AcmeStore` trait exists so **#1635** can add a DB/per-tenant impl; only `FsAcmeStore`
  ships here.
- **Multi-replica HTTP-01 token sharing** — the `Http01Tokens` map is per-host, so
  HTTP-01 assumes challenge traffic reaches the ordering replica (**single-host
  assumption**). Renewal is leader-elected; non-leaders adopt a newer cert from a
  *shared* store, but a network-partitioned token map is not addressed here.
- **TLS-ALPN-01** — not implemented.
- **Docs** — the docs session owns docs. Follow-up: a ≤10-line config happy-path and a
  "reverse-proxy opt-out" note in the deployment guide. Not written here.

## Tests

All unit tests land green (no real network in the default suite):
`challenge_router` (200+key-auth / 404 / 308-preserving-path+query), `Http01Tokens`
round-trip, `FsAcmeStore` account+cert round-trip with `0600` perms, `needs_renewal`
threshold matrix, `TlsConfig::validate()` matrix (static-only / acme-only / both /
neither / cert-without-key / empty domains / empty email / wildcard→#1620), placeholder
self-signed load+swap via `resolver.store()`, `AcmeHealthIndicator` grading, and the
doctor `check_acme_*_impl` pure graders. A real-ACME (pebble/LE-staging) integration
test is intentionally **not** included in the default suite (tracked as a follow-up).

---
_Generated by [Claude Code](https://claude.ai/code/session_01LMeNyt3tNtHRvcQm9QGsZs)_


madmax983:
## Local verification (the gate)

Per the project's "local verification is the gate" policy, this is a clean-rebuild verification on head `604948d`, exercising both the `tls` and `acme` features. All gates **VERIFIED GREEN**:

| Gate | Result |
| --- | --- |
| `cargo fmt --all -- --check` | clean (exit 0) |
| `cargo clippy --workspace --all-targets --features acme -- -D warnings` | clean (exit 0) |
| `cargo clippy --workspace --all-targets -- -D warnings` (no-feature) | clean (exit 0) |
| `cargo test -p autumn-web --features acme --lib` | 3776 passed, 0 failed |
| `cargo test -p autumn-web --features acme --test integration_tests` | passed (incl. `tls_serving` 6/6, with `concurrent_handshakes_do_not_serialize_accept`) |
| `cargo test -p autumn-cli --features acme doctor` | 347 passed, 0 failed |
| `cargo check -p autumn-web` + `-p autumn-cli` (no-feature) | clean |

This branch is now rebased onto `trunk-dev` as a single merge commit, with all conflicts resolved preserving both the #1603 TLS-listener hardening and the #1608 ACME work. The ACME slice was hardened across multiple review rounds — HTTP-01 issuance/renewal, the dual-stack `:80` listener, leadership safety, storage atomicity, and full doctor↔runtime config-validation parity.

Per the "local verification is the gate" policy, this PR is merge-ready on the strength of the green local gates above. Now that it targets `trunk-dev`, its GitHub Actions CI will run; real CI results will be acted on if/when they arrive.

---

PR #1878, opened 2026-07-13
Fix #1873: treat empty-filename multipart parts as non-file input

madmax983:
## Before / After

A user who submits a form with an **optional or empty file input** sends a multipart part with `Content-Disposition: …; filename=""` and a 0-byte body.

- **Before:** When the project configures a non-empty `security.upload.allowed_mime_types` OR sets `reject_on_content_type_mismatch = true`, the empty part was treated as a file, failed MIME sniffing/allow-list verification, and the **whole request was rejected with 400** — breaking optional-empty-create and no-new-file (preserve-existing) submits, and any hand-written `Multipart` handler. (Default-config projects were unaffected.)
- **After:** The empty-filename part is treated as a non-file field, skips enforcement, and is returned to the handler like any other non-file field, so the submit succeeds. Real, genuinely-named file parts are still sniffed and enforced exactly as before.

## How

`autumn_web::extract::Multipart::next_field` gated sniffing/enforcement on `field.file_name().is_some()`, which is `true` for `Some("")`. The predicate now requires a **non-empty** filename, mirroring the existing scaffold guard `file_name().is_some_and(|f| !f.is_empty())`:

```rust
let needs_sniff = field.file_name().is_some_and(|name| !name.is_empty())
    && (!self.config.allowed_mime_types.is_empty()
        || self.config.reject_on_content_type_mismatch);
```

This is a narrowing of what counts as a "file" part; it does not weaken enforcement for any part that has a real filename.

## Tests

Added three consolidated integration tests in `autumn/tests/integration/extractors.rs`:
- `multipart_empty_filename_part_skips_allow_list_enforcement` — `filename=""` passes under a configured `allowed_mime_types`.
- `multipart_empty_filename_part_skips_strict_mismatch_enforcement` — `filename=""` passes under `reject_on_content_type_mismatch = true`.
- `multipart_named_disallowed_part_still_rejected` — regression guard: a genuinely-named disallowed part is still 400-rejected.

## Gate output

```
$ cargo fmt --all -- --check
FMT_CHECK_EXIT=0

$ cargo clippy --workspace --all-targets -- -D warnings
    Checking autumn-web v0.6.0 (/home/user/autumn/autumn)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 8m 54s
CLIPPY_EXIT=0

$ cargo test -p autumn-web --features multipart --test integration_tests multipart_
running 21 tests
test integration::extractors::multipart_empty_filename_part_skips_strict_mismatch_enforcement ... ok
test integration::extractors::multipart_empty_filename_part_skips_allow_list_enforcement ... ok
test integration::extractors::multipart_named_disallowed_part_still_rejected ... ok
test integration::extractors::multipart_spoofed_html_declared_png_is_rejected ... ok
test integration::extractors::multipart_genuine_png_passes ... ok
test integration::extractors::multipart_short_file_rejected_when_allow_list_set ... ok
test integration::extractors::multipart_strict_mode_rejects_missing_content_type ... ok
test integration::extractors::multipart_svg_rejected_even_when_declared_image ... ok
test integration::extractors::multipart_unrecognized_binary_declared_png_is_rejected ... ok
... (all 21 multipart tests)
test result: ok. 21 passed; 0 failed; 0 ignored; 0 measured; 1248 filtered out; finished in 0.22s
```

Closes #1873

---

### DIFF
--- a/autumn/src/extract.rs
+++ b/autumn/src/extract.rs
@@ -222,14 +222,44 @@ impl Multipart {
         // must still be sniffed and enforced, since callers can consume its
         // bytes by field name. Capture the empty-filename flag now, before the
         // body is consumed by the prefix-buffering loop, to avoid borrowing
-        // `field` after it has been read.
+        // `field` after it has been read. The same flag drives the
+        // `file_name()` normalization: a genuinely-empty optional input
+        // surfaces as `None`, a non-empty-body empty-filename part as a file.
         let filename_is_empty = field.file_name().is_some_and(str::is_empty);
 
         if !needs_sniff {
-            return Ok(Some(MultipartField::new(
-                field,
-                self.config.max_file_size_bytes,
-            )));
+            // Default path: no MIME policy, so no sniffing. We still must detect
+            // the genuinely-empty optional file input (empty filename AND empty
+            // body) so `file_name()` normalizes it to `None` consistently with
+            // the sniff path below. Only empty-filename parts can qualify, so
+            // only those need a body peek; every other part streams through
+            // untouched with no buffering.
+            if !filename_is_empty {
+                return Ok(Some(MultipartField::new(
+                    field,
+                    self.config.max_file_size_bytes,
+                )));
+            }
+
+            // Peek whether the empty-filename body yields any bytes. Buffer the
+            // peeked chunk as the prefix so the consuming methods replay it and
+            // no bytes are lost; the per-file cap is still enforced there.
+            let mut prefix: Vec<u8> = Vec::new();
+            if let Some(chunk) = field
+                .chunk()
+                .await
+                .map_err(|err| multipart_error_to_error(&err))?
+            {
+                prefix.extend_from_slice(&chunk);
+            }
+            let is_empty_optional_input = prefix.is_empty();
+            return Ok(Some(MultipartField {
+                inner: field,
+                max_file_size_bytes: self.config.max_file_size_bytes,
+                prefix,
+                sniffed_content_type: None,
+                is_empty_optional_input,
+            }));
         }
 
         // Buffer a bounded leading prefix (a few chunks at most) for sniffing,
@@ -291,6 +321,7 @@ impl Multipart {
             max_file_size_bytes: self.config.max_file_size_bytes,
             prefix,
             sniffed_content_type: sniffed,
+            is_empty_optional_input: is_empty_file_input,
         }))
     }
 }
@@ -459,6 +490,12 @@ pub struct MultipartField<'a> {
     /// Sniffed (magic-byte) content type, if the leading bytes were recognized.
     /// `infer` returns a `&'static str`, so this stores the borrow directly.
     sniffed_content_type: Option<&'static str>,
+    /// Whether this part is a genuinely-empty optional file input: an empty
+    /// `filename=""` AND a 0-byte body. Computed in `next_field`, where body
+    /// emptiness is observable, and used by [`file_name`](Self::file_name) to
+    /// normalize such a part to `None` while leaving an empty-filename part with
+    /// a non-empty body classified as a file (`Some("")`).
+    is_empty_optional_input: bool,
 }
 
 #[cfg(all(feature = "multipart", feature = "storage"))]
@@ -482,6 +519,7 @@ impl<'a> MultipartField<'a> {
             max_file_size_bytes,
             prefix: Vec::new(),
             sniffed_content_type: None,
+            is_empty_optional_input: false,
         }
     }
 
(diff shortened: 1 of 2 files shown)


madmax983:
Thanks @gemini-code-assist and the codex review for catching the P1 MIME-policy bypass. The previous predicate gated sniffing on `file_name().is_some_and(|n| !n.is_empty())`, so a crafted part with `filename=""` but a **non-empty** malicious body skipped all sniffing/allow-list/mismatch enforcement while callers could still consume its bytes by field name.

Fix (in `autumn/src/extract.rs`, `Multipart::next_field`):
- `needs_sniff` is back to gating on `file_name().is_some()`.
- Enforcement is now skipped **only** when the filename is empty **and** the buffered body is empty (`is_empty_file_input = filename_is_empty && prefix.is_empty()`) — i.e. a genuinely 0-byte optional file input, which still passes through as a non-file/absent field (the #1873 behavior).
- Any `filename=""` part with a non-empty body — and every named file part — is sniffed and enforced exactly as before, closing the bypass.

Added a regression test `multipart_empty_filename_non_empty_body_still_enforced` (empty filename + non-empty disallowed body under an allow-list → 400). The two existing #1873 passthrough tests still hold. fmt/clippy clean; all 22 multipart integration tests pass.

---

madmax983:
Addressed codex's P2 on `MultipartField::file_name()`.

`file_name()` now normalizes an empty `filename=""` to `None` (`self.inner.file_name().filter(|name| !name.is_empty())`). This matches issue #1873's intent: a blank optional file input submits a `filename=""` / 0-byte part representing an *absent* file, so a handler gating on `field.file_name().is_some()` no longer mistakes it for a real upload and persists a zero-byte file over an existing value.

The MIME-enforcement carve-out is unaffected. `next_field()` computes `filename_is_empty` from the raw axum `Field::file_name()` while sniffing (not via this wrapper getter), so a `filename=""` part with a NON-empty body is still sniffed and rejected under an allow-list — the `multipart_empty_filename_non_empty_body_still_enforced` regression continues to pass.

Also audited every other `file_name()` usage: `form.rs:643` and the internal `next_field` checks operate on the raw `axum::extract::Multipart` field, and the remaining hits are `std::path::Path::file_name` — none touch the normalized wrapper getter.

Added two tests: `multipart_empty_filename_yields_none_file_name` (empty part → `None`) and `multipart_named_part_yields_some_file_name` (named part → `Some("real.png")`). Full `multipart_` suite green (24 passed) with fmt + clippy clean.

---

madmax983:
Addressed the codex P2 consistency finding.

`MultipartField::file_name()` no longer normalizes on the filename string alone — it was returning `None` for *any* empty filename, which disagreed with how `next_field` classifies a `filename=""` part with a non-empty body (that part is a file: it gets sniffed and allow-list enforced). A handler splitting file-vs-text on `file_name().is_some()` would therefore have misclassified attacker-supplied bytes as a text field or silently dropped a nameless upload.

The fix makes `file_name()` agree with `next_field`:
- empty filename **and** empty (0-byte) body → `None` (genuinely-absent optional input; keeps the zero-byte-file footgun from #1873 fixed);
- empty filename **and** non-empty body → `Some("")` (it is a file, so it is surfaced as one);
- non-empty filename → unchanged `Some("name")`.

The empty-vs-non-empty-body decision is computed in `next_field`, where body emptiness is observable, and stored on the field as `is_empty_optional_input`; the getter returns `None` from that flag rather than inspecting the (lazily-read) filename string. Both configs compute it uniformly — the sniff path reuses the existing `is_empty_file_input`, and the default (no-MIME-policy) path peeks one body chunk for empty-filename parts, buffering it as the sniff prefix so no bytes are lost. The P1 enforcement carve-out (empty filename + non-empty body still enforced under an allow-list) is unchanged.

Tests added: empty filename + empty body under an allow-list → `file_name()` is `None`; empty filename + non-empty allowed body → `file_name()` is `Some("")` (the consistency assertion). All existing multipart tests still pass.

---

madmax983:
Addressed the codex P2 (default-path buffering) in 970cb55.

The consistency commit added a `field.chunk()` peek on the default (no-MIME-policy) path to detect an empty optional file input. Because `chunk()` can return the entire part in one frame, that peek could buffer up to the request limit into memory before the handler could reject or stream the field — a regression versus the prior zero-buffering behavior on that path.

Fix: removed the default-path peek entirely. On the default path the field is now constructed via `MultipartField::new(...)` with no buffering and no `file_name()` normalization, so it streams straight through as before #1873.

`file_name()` normalization is now scoped to the **sniff path only** (an allow-list or strict-mismatch policy is configured), where emptiness is already known from the bounded 512-byte sniff prefix — so a genuinely-empty optional input (`filename=""` + 0-byte body) still surfaces as `None` there, while an empty-filename part with a non-empty body remains sniffed/enforced and surfaces as `Some("")` (P1 preserved). #1873's CSRF/allow-list bug only ever occurred under a configured allow-list, and default config was explicitly unaffected — there is no enforcement/classification on the default path, so there is no inconsistency to resolve there and `Some("")` is the intended pre-#1873 behavior.

The default-config test was updated to assert `Some("")` (streamed through, unbuffered); the sniff-path consistency tests are unchanged. All 26 `multipart_` integration tests pass; fmt and clippy (`-D warnings`) are clean.

---

PR #1887, opened 2026-07-13
fix(csrf): bounded-prefix token scan + stream remainder (no full body buffering)

madmax983:
Closes #1866

> **Approach changed (codex P2, approved).** This PR no longer *raises* the full-body buffer cap — that would make the CSRF layer buffer whole uploads (up to 32 MiB × concurrency) into memory. It now does a **bounded-prefix scan + stream-remainder**: scan only a small, configurable prefix for the `_csrf` token and stream the rest of the body through untouched. See "Why the redesign" below.

## Problem

With CSRF enforcement on (the prod-profile default), scaffolded forms carry the `_csrf` token in the request **body**. The middleware located it by fully buffering the body via `to_bytes(body, max_scan_bytes)`.

The earlier commits on this branch made `max_scan_bytes` track `security.upload.max_request_size_bytes` (32 MiB default) and returned a `413` on overflow. That fixed the spurious-403 symptom but introduced a worse problem.

## Why the redesign (codex P2 — valid)

Aligning the scan cap to `max_request_size_bytes` makes the CSRF layer **fully buffer every body-token upload into memory** — up to 32 MiB × concurrency — *before the handler runs*. `DefaultBodyLimit` is only a marker extension, so nothing else bounds this `to_bytes` copy. This:

- **Defeats the streaming upload path** — the scaffold's `save_to_blob_store` streams a multipart field-by-field to the blob store; full buffering in the CSRF layer forces the whole upload into RAM first.
- Is a **DoS-shaped memory cost** — concurrent large uploads can pin tens/hundreds of MiB each.

## The fix: bounded-prefix scan + stream remainder

1. **Bounded, configurable prefix cap.** New `security.csrf.token_scan_bytes` (**2 MiB default**), sourced from CSRF config and **deliberately independent** of `upload.max_request_size_bytes`. `CsrfLayer::from_config` threads it into settings; the router no longer feeds `max_request_size_bytes` in (the `with_max_scan_bytes` builder is removed).
2. **Prefix scan without erroring on overflow.** `collect_body_prefix` reads at most `token_scan_bytes` of the body into a prefix buffer **without failing** when the body is larger, then scans that prefix for `_csrf` (same urlencoded / multipart scanners as before). This mirrors the existing `collect_body` pattern in `submit_token.rs` for consistency.
3. **Stream the remainder.** The full body is reconstructed for the handler: if it ended within the cap, rebuilt as a sized body from the prefix; otherwise the prefix is **chained** with the remaining (partially-consumed) original stream via `http-body-util` / `Body::from_stream`, so the tail **streams through** rather than being buffered. Downstream multipart/urlencoded extractors see identical bytes and framing (a regression test asserts the handler receives the full, unmodified body).
4. **No CSRF-layer 413.** `CsrfVerdict::BodyTooLarge`, the `LengthLimitError` detection, and the payload-too-large mapping are removed. The token is either found in the prefix (proceed) or not (existing **403** missing-token). Genuinely-oversized bodies now **stream through** and are rejected **downstream** by the real body/upload limits (`DefaultBodyLimit` → natural `413` at the extractor) — where that belongs.

## Token-early constraint (+ escape hatch)

Because only the prefix is scanned, `_csrf` must appear within the first `token_scan_bytes` of the body. **Scaffolded forms already emit the hidden `_csrf` field before the file field** (same precedent as the submit-token hidden fields), so scaffolds are always safe. Hand-written forms that place large fields ahead of `_csrf` should move the token earlier or **raise `security.csrf.token_scan_bytes`** — documented on the config field and the layer doc comment.

## Tests

In `autumn/src/security/csrf.rs` (`#[cfg(test)] mod tests`):

- **Stream-through regression** — a multipart body far larger than the prefix cap with `_csrf` **first**: CSRF **passes** (token found in prefix) **and** the inner handler receives the **full body intact** (handler echoes the received byte count; asserted equal to the original length), proving the remainder streamed through and reconstruction is lossless. A urlencoded variant covers the same.
- **Token beyond the cap → 403** — `_csrf` placed after `> token_scan_bytes` of preceding data is not found → 403 (documents the constraint).
- **Configurable cap** — the *same* body flips 403 → 200 as `token_scan_bytes` moves (4 KiB vs 64 KiB).
- Removed the three CSRF-layer-413 tests (`post_with_large_body_returns_413`, `post_large_multipart_exceeding_scan_cap_returns_413`, `post_large_multipart_over_cap_without_token_returns_413`) since that behavior is reverted. Existing valid/missing-token tests unchanged.

## Gate output (verbatim)

**`cargo fmt --all -- --check`** — clean (no diff)

**`cargo clippy -p autumn-web --all-targets -- -D warnings`**
```
    Checking autumn-web v0.6.0 (/home/user/autumn/autumn)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 07s
```

**`cargo test -p autumn-web --lib security::csrf::`**
```
running 37 tests
...
test security::csrf::tests::post_large_multipart_token_first_streams_full_body_and_passes ... ok
test security::csrf::tests::post_with_large_urlencoded_body_token_first_streams_and_passes ... ok
test security::csrf::tests::post_multipart_token_beyond_scan_prefix_returns_403 ... ok
test security::csrf::tests::configurable_scan_cap_moves_token_boundary ... ok

test result: ok. 37 passed; 0 failed; 0 ignored; 0 measured; 3723 filtered out; finished in 0.05s
```

https://claude.ai/code/session_01L9uehBWRkPbZC8PjUjPRUT

---

### DIFF
--- a/autumn/src/config.rs
+++ b/autumn/src/config.rs
@@ -3928,6 +3928,11 @@ impl AutumnConfig {
             "AUTUMN_SECURITY__CSRF__COOKIE_NAME",
             &mut self.security.csrf.cookie_name,
         );
+        parse_env(
+            env,
+            "AUTUMN_SECURITY__CSRF__TOKEN_SCAN_BYTES",
+            &mut self.security.csrf.token_scan_bytes,
+        );
 
         self.apply_rate_limit_env_overrides_with_env(env);
 
@@ -10908,6 +10913,25 @@ path = "/healthz"
         config.security.allow_unauthorized_repository_api = true;
         config.apply_env_overrides_with_env(&env);
         assert!(!config.security.allow_unauthorized_repository_api);
+    }
+
+    #[test]
+    fn env_override_csrf_token_scan_bytes() {
+        let env = MockEnv::new().with("AUTUMN_SECURITY__CSRF__TOKEN_SCAN_BYTES", "8388608");
+        let mut config = AutumnConfig::default();
+        // Default is 2 MiB; the env override must raise it.
+        assert_eq!(config.security.csrf.token_scan_bytes, 2 * 1024 * 1024);
+        config.apply_env_overrides_with_env(&env);
+        assert_eq!(config.security.csrf.token_scan_bytes, 8_388_608);
+    }
+
+    #[test]
+    fn env_override_csrf_token_scan_bytes_invalid_is_ignored() {
+        let env = MockEnv::new().with("AUTUMN_SECURITY__CSRF__TOKEN_SCAN_BYTES", "not-a-number");
+        let mut config = AutumnConfig::default();
+        config.apply_env_overrides_with_env(&env);
+        // Invalid values are ignored, leaving the default intact.
+        assert_eq!(config.security.csrf.token_scan_bytes, 2 * 1024 * 1024);
     }
 
     // ── [openapi] config section tests (RED phase) ─────────────────────────
--- a/autumn/src/router.rs
+++ b/autumn/src/router.rs
@@ -2178,7 +2178,22 @@ where
         // remainder through, so the cap comes from CSRF config — NOT from
         // `upload.max_request_size_bytes` (which would force whole uploads into
         // memory and defeat the streaming upload path).
-        let mut csrf_layer = crate::security::CsrfLayer::from_config(&config.security.csrf);
+        //
+        // Clamp the effective prefix to the global body limit: the CSRF layer
+        // must never buffer more than `upload.max_request_size_bytes`. In the
+        // normal/high-upload case the small `token_scan_bytes` prefix wins (the
+        // `min` keeps it at 2 MiB — it is *not* raised to the upload limit).
+        // Only when an operator deliberately lowers the global limit *below* the
+        // prefix cap does the upload limit clamp the scan down — the whole body
+        // is ≤ that limit anyway, so an early `_csrf` token is still in range,
+        // and anything larger is rejected downstream by `DefaultBodyLimit`.
+        let effective_scan_bytes = config
+            .security
+            .csrf
+            .token_scan_bytes
+            .min(config.security.upload.max_request_size_bytes);
+        let mut csrf_layer = crate::security::CsrfLayer::from_config(&config.security.csrf)
+            .with_max_scan_bytes(effective_scan_bytes);
         if let Some(keys) = signing_keys {
             csrf_layer = csrf_layer.with_signing_keys(keys);
         }
--- a/autumn/src/router.rs
+++ b/autumn/src/router.rs
@@ -2173,8 +2173,12 @@ where
 {
     // CSRF middleware (only applied when enabled)
     if config.security.csrf.enabled {
-        let mut csrf_layer = crate::security::CsrfLayer::from_config(&config.security.csrf)
-            .with_max_scan_bytes(config.security.upload.max_request_size_bytes);
+        // The CSRF token scan reads only a bounded prefix of the body
+        // (`security.csrf.token_scan_bytes`, 2 MiB default) and streams the
+        // remainder through, so the cap comes from CSRF config — NOT from
+        // `upload.max_request_size_bytes` (which would force whole uploads into
+        // memory and defeat the streaming upload path).
+        let mut csrf_layer = crate::security::CsrfLayer::from_config(&config.security.csrf);
         if let Some(keys) = signing_keys {
             csrf_layer = csrf_layer.with_signing_keys(keys);
         }
--- a/autumn/src/security/config.rs
+++ b/autumn/src/security/config.rs
@@ -39,6 +39,7 @@
 //! | `AUTUMN_SECURITY__HEADERS__CONTENT_SECURITY_POLICY` | `security.headers.content_security_policy` | `String` |
 //! | `AUTUMN_SECURITY__HEADERS__CSP_NONCE__ENABLED` | `security.headers.csp_nonce.enabled` | `bool` |
 //! | `AUTUMN_SECURITY__CSRF__ENABLED` | `security.csrf.enabled` | `bool` |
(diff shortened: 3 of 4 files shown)


madmax983:
Addressed the codex P2 finding (commit 5bfce13).

**Issue:** When a zero-JS form body carrying `_csrf` is larger than `max_scan_bytes` (== the configured `upload.max_request_size_bytes`), `axum::body::to_bytes` errors with a length-limit overflow. The old `unwrap_or_else(|_| Bytes::new())` treated that as an *absent* token, so CSRF short-circuited with a misleading **403** — masking the **413 Payload Too Large** that #1866 explicitly wanted to be observable (the `DefaultBodyLimit` marker doesn't pre-reject, so the oversized body reaches the CSRF layer first).

**Fix:** The body-scan path now distinguishes the scan-cap overflow from a genuinely-missing token and returns **413** for the overflow. Detection walks the `to_bytes` error source chain and downcasts to `http_body_util::LengthLimitError` (the mechanism axum documents), so it's robust to wrapping layers. `verify_csrf_token` returns a `CsrfVerdict` enum (`Valid` / `Missing` / `BodyTooLarge`), and the RFC 7807 problem response is built for both statuses (413 → `autumn.payload_too_large`).

- Genuinely-absent-token behavior is **unchanged** (still 403).
- Non-length body read errors (e.g. a client disconnect mid-body) preserve the prior missing-token 403 — this module has no dedicated malformed-body/400 path, so 403 is the consistent choice.
- Applies to both urlencoded and multipart bodies over the cap, which is correct.

**Tests:** Updated `post_large_multipart_exceeding_scan_cap_rejected` → `post_large_multipart_exceeding_scan_cap_returns_413` (now asserts 413), plus a new `post_large_multipart_over_cap_without_token_returns_413` confirming the overflow is classified *before* token presence. `post_with_large_body_fails_csrf` → `post_with_large_body_returns_413` for the urlencoded path. The within-cap pass test is unchanged. `cargo test -p autumn-web --lib security::csrf::` — 37 passed; fmt + clippy (`-D warnings`) clean.

---

madmax983:
Reworked per the codex **P2**. The prior approach (raise the full-body buffer cap to `upload.max_request_size_bytes` + 413 on overflow) was valid feedback: it made the CSRF layer fully buffer **every** body-token upload — up to 32 MiB × concurrency — into memory before the handler ran, defeating the streaming upload path (`save_to_blob_store`) and creating a DoS-shaped memory cost, since `DefaultBodyLimit` is only a marker extension and nothing else bounded that `to_bytes` copy.

New design (pushed as `3c58206`): **bounded-prefix scan + stream-remainder**.
- Scan only a bounded, **configurable** prefix (`security.csrf.token_scan_bytes`, **2 MiB default**, independent of `max_request_size_bytes`) for `_csrf`, without erroring on overflow — mirrors the existing `collect_body` pattern in `submit_token.rs`.
- Reconstruct the full body for the handler by **chaining** the prefix with the remaining stream, so the tail streams through unbuffered (a regression test asserts the handler receives the full, unmodified body).
- **Dropped the CSRF-layer 413** (`CsrfVerdict::BodyTooLarge` removed). Oversized bodies now stream through and are rejected **downstream** by the real body/upload limits (natural 413); a token beyond the prefix falls to the existing 403.
- Documented the token-early constraint — scaffold forms already emit `_csrf` before the file field, and the cap is the escape hatch for hand-written forms.

fmt/clippy clean; `cargo test -p autumn-web --lib security::csrf::` → 37 passed.

---

madmax983:
Thanks for the codex P2 catch — it was valid. This codebase applies `AUTUMN_SECURITY__*` env overrides through the manual `apply_security_env_overrides_with_env` (no automatic figment-style nested provider), and that function only consumed the CSRF `ENABLED`/`TOKEN_HEADER`/`COOKIE_NAME` vars. So the documented `AUTUMN_SECURITY__CSRF__TOKEN_SCAN_BYTES` was silently ignored on env-driven deployments.

Fixed in a7f16d6: added a `parse_env` arm for `AUTUMN_SECURITY__CSRF__TOKEN_SCAN_BYTES` mapping to `security.csrf.token_scan_bytes`, mirroring the sibling numeric upload overrides (invalid values ignore-with-warn, matching the existing convention). Added tests covering a valid value (`8388608`) and an invalid value (falls back to the 2 MiB default). `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, and the targeted `config::` tests are all green.

---

madmax983:
Addressed the codex P2 finding on the CSRF scan cap vs. a lowered global body limit.

**Fix:** the effective CSRF body-token scan prefix is now clamped to `min(security.csrf.token_scan_bytes, security.upload.max_request_size_bytes)`.

- The small `token_scan_bytes` prefix (2 MiB default) stays the **primary** bound — this is *not* a return to #1866's original bug where `max_request_size_bytes` was used directly as the cap (32 MiB full-buffering). The `min` keeps the prefix at 2 MiB in the normal/high-upload case; the upload limit only acts as an **upper clamp**.
- When an operator deliberately lowers `max_request_size_bytes` below the prefix cap (e.g. 64 KiB), the effective scan cap now drops to 64 KiB, so the CSRF layer never buffers a larger prefix than the deployment's global body ceiling (anything larger is rejected downstream by `DefaultBodyLimit` anyway).
- Token discovery is preserved: when the body limit is below the prefix cap the whole body is ≤ that limit, so an early `_csrf` token is still within the scanned prefix.

Implemented via a small re-introduced `CsrfLayer::with_max_scan_bytes` builder, applied at the router CSRF wiring where both configs are in scope. Added two construction-level tests asserting the clamp (2 MiB + 64 KiB upload → 64 KiB effective) and the normal case (2 MiB + 32 MiB upload → stays 2 MiB). All existing csrf tests (stream-through / token-beyond-cap / configurable-cap) still pass.

---

PR #1888, opened 2026-07-13
Clean up orphaned upload blobs on scaffold handler early returns

madmax983:
Closes #1872

## Before / After

**Before:** the scaffold generator emits create/update handlers that stream each uploaded file to the blob store *during* multipart parsing — before changeset validation and before the DB write. So a user submitting an invalid create/edit with a file attached — a validation error, a duplicate/unique violation, or a DB error — leaves the just-uploaded file orphaned in the blob store. This leaks on **every ordinary invalid submission**, not just rare error paths.

**After:** the generated handler best-effort-deletes the just-saved file on those failure paths, so invalid submits don't leak blobs. The store already holds a `delete` (a no-op when the key is absent), so cleanup is cheap and safe.

## How

Generator restructure (`autumn-cli/src/generate/scaffold.rs`):

- The handlers record the keys of the freshly-saved uploads into `saved_blob_keys` right after the multipart save, and splice a best-effort cleanup (`for key in &saved_blob_keys { let _ = store.delete(key).await; }`) before every early return that follows the save.
- The `into_new(...)?` parse and the update's current-row load `?` are converted into explicit `match`es that run cleanup, then return the exact same error/response the `?` would have produced.
- **Only the newly-saved `<field>_blob` is ever deleted** — the update path binds `new.<field> = <field>_blob.or(current.<field>)`, and only the fresh `<field>_blob` keys are enrolled, so a **preserved existing blob is never touched**.
- Covers validation (422), `into_new` failure, unique-violation (422), the generic DB `Err`, and the update's row-load / not-found guard — on **both** create and update.
- The **success path is unchanged** and never deletes. Multiple attachment fields are supported (each saved blob is cleaned up).

## Scope

The successful-replacement orphan — the *old* blob left behind when an update successfully swaps in a new file — is intentionally **out of scope** and left to the periodic sweep, consistent with how the direct-upload path deliberately tolerates orphans (bind-step promotion + sweep). This fix targets only the leak on *failed* submissions.

## Gate output (verbatim)

**1. `cargo fmt --all -- --check`**
```
fmt: clean (exit 0)
```

**2. `cargo clippy --workspace --all-targets -- -D warnings`**
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 52s
```
(no warnings; only unrelated "Tailwind CSS CLI not found" build-script notices)

**3. `cargo test -p autumn-cli --test cli_tests scaffold_form_for`**
```
running 12 tests
test ...::generated_attachment_scaffold_cargo_checks ... ignored
test ...::generated_form_for_scaffold_cargo_checks ... ignored
test ...::generated_scaffold_with_missing_reference_target_cargo_checks ... ignored
test ...::scaffold_attachment_emits_zero_js_multipart_handlers ... ok
test ...::scaffold_attachment_cleans_up_saved_blob_on_early_return ... ok
test ...::scaffold_attachment_emits_generated_multipart_write_path_test ... ok
test ...::scaffold_create_and_edit_views_use_single_form_for_call ... ok
test ...::rescaffold_with_added_column_leaves_view_form_call_unchanged ... ok
test ...::scaffold_decimal_column_overrides_number_step ... ok
test ...::scaffold_enum_column_overrides_to_select ... ok
test ...::scaffold_references_column_with_missing_target_falls_back_to_number_input ... ok
test ...::scaffold_references_column_overrides_to_select_and_loads_options ... ok

test result: ok. 9 passed; 0 failed; 3 ignored; 0 measured; 194 filtered out
```

**Generated-code compile proof** — `cargo test -p autumn-cli --test cli_tests generated_attachment_scaffold_cargo_checks -- --ignored`
```
running 1 test
test ...::generated_attachment_scaffold_cargo_checks ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 205 filtered out; finished in 69.22s
```

https://claude.ai/code/session_01L9uehBWRkPbZC8PjUjPRUT

---

### DIFF
--- a/autumn-cli/src/generate/scaffold.rs
+++ b/autumn-cli/src/generate/scaffold.rs
@@ -2117,9 +2117,14 @@ fn render_routes_file(
     // Issue #1872: the create/update handlers stream each uploaded file to the
     // blob store *before* changeset validation and the DB write, so any early
     // return after the save would orphan the just-uploaded blob. We record only
-    // the freshly-saved `<field>_blob`s (into `saved_blob_keys`) and best-effort
-    // delete them before every early-return path — never a preserved existing
-    // blob. This snippet is spliced before each such `return`.
+    // the freshly-saved `<field>_blob`s (into `saved_blob_keys`) — pushed at the
+    // point of each save, so the key set is populated incrementally as blobs are
+    // persisted — and best-effort delete them before every early-return path that
+    // can occur after the first save: a later attachment field's save failure,
+    // `decode_form` on malformed scalar input, changeset validation, the
+    // unique-violation 422, the generic DB error, and (update) the current-row
+    // load / not-found. A preserved existing blob is never enrolled. This snippet
+    // is spliced before each such `return`.
     let blob_cleanup = if has_attachments {
         "for key in &saved_blob_keys {\n        let _ = store.delete(key).await;\n    }\n    "
     } else {
@@ -2128,19 +2133,19 @@ fn render_routes_file(
     let form_decode_block = if has_attachments {
         let mut blob_decls = String::new();
         let mut match_arms = String::new();
-        let mut key_captures = String::new();
         for f in &attachment_fields {
             let name = &f.name;
             let _ = writeln!(
                 blob_decls,
                 "    let mut {name}_blob: Option<autumn_web::storage::Blob> = None;"
             );
-            let _ = write!(
-                key_captures,
-                "\n    if let Some(blob) = &{name}_blob {{\n        \
-                 saved_blob_keys.push(blob.key.clone());\n    \
-                 }}"
-            );
+            // Issue #1872: record the freshly-saved blob's key into
+            // `saved_blob_keys` *immediately* after each successful save, so any
+            // later fallible early return (a subsequent attachment field's save,
+            // `decode_form`, validation, the DB write) can best-effort delete it.
+            // The save's own `?` is expanded into an explicit `match` that runs
+            // the cleanup first, covering the case where an earlier attachment
+            // field was already persisted when a later one fails to save.
             let _ = write!(
                 match_arms,
                 "            Some(\"{name}\") => {{\n                \
@@ -2150,7 +2155,14 @@ fn render_routes_file(
                  chrono::Utc::now().timestamp_nanos_opt().unwrap_or_default(),\n                        \
                  uuid::Uuid::new_v4()\n                    \
                  );\n                    \
-                 {name}_blob = Some(field.save_to_blob_store(&*store, key).await?);\n                \
+                 let blob = match field.save_to_blob_store(&*store, key).await {{\n                        \
+                 Ok(blob) => blob,\n                        \
+                 Err(err) => {{\n                            \
+                 {blob_cleanup}return Err(err);\n                        \
+                 }}\n                    \
+                 }};\n                    \
+                 saved_blob_keys.push(blob.key.clone());\n                    \
+                 {name}_blob = Some(blob);\n                \
                  }}\n            \
                  }}\n"
             );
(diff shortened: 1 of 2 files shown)


madmax983:
## Fix for codex P2: cleanup keys recorded too late

Good catch — `saved_blob_keys` was populated *after* the `decode_form(...)?` rebuild step, so malformed scalar input in the same request (e.g. a non-numeric value for an integer field) made `decode_form(...)?` return early **before** the keys were recorded, orphaning the just-saved upload. A mid-loop `save_to_blob_store(...).await?` failure on a later attachment field could similarly orphan an earlier one.

**Fix (both create and update handlers):** the cleanup key set is now populated *incrementally at save time* — each `<field>_blob` key is pushed onto `saved_blob_keys` immediately after the blob is persisted, inside the multipart loop. The blob save's `?` and the `decode_form(...)?` are both expanded into explicit `match`es that run the existing `blob_cleanup` snippet (`for key in &saved_blob_keys { let _ = store.delete(key).await; }`) before returning the same error `?` would have produced.

Every fallible early return that can occur after the first blob is persisted now cleans up:

- **create:** mid-loop subsequent-field save failure, `decode_form` (malformed scalar), invalid-changeset 422, `into_new` parse, unique-violation 422, generic DB error.
- **update:** mid-loop subsequent-field save failure, `decode_form` (malformed scalar), invalid-changeset 422, current-row load (not-found / DB error), `into_new` parse, unique-violation 422, generic DB error, not-found (`updated == 0`) guard.

All previously covered paths keep their exact error→response mapping. Only freshly-saved keys are ever deleted — a preserved existing blob (update's `.or(current.<field>)`) is never touched, the success path never deletes, and non-attachment scaffolds emit byte-identical output with no cleanup machinery.

Tests in `scaffold_form_for.rs` now assert `saved_blob_keys.push(...)` appears before `decode_form(` in the emitted source, and that cleanup guards the `decode_form` and mid-loop save paths (create: 6 `store.delete` early returns, update: 8). `cargo fmt --check`, `clippy -D warnings`, `scaffold_form_for`, and the ignored generate-then-compile harness all pass.

---

madmax983:
Addressed the codex P2 (orphaned upload on a malformed multipart request after a blob save).

Previously only the `save_to_blob_store` and `decode_form` `?`s ran cleanup; the other fallible ops in the parse span reached *after* a save but *before* `decode_form` — `multipart.next_field().await?`, `field.bytes_limited().await?`, and the UTF-8 `map_err(...)?` — still bypassed cleanup and leaked the recorded blob.

**Fix (centralized):** the generated create/update handlers now run the entire fallible multipart-parse-and-decode span inside a single

```rust
let __parse_result: AutumnResult<{Pascal}Form> = async {
    // next_field()?, each save_to_blob_store()?, bytes_limited()?, utf8 map_err(...)?, decode_form(...)?
}.await;
let form = match __parse_result {
    Ok(form) => form,
    Err(err) => { for key in &saved_blob_keys { let _ = store.delete(key).await; } return Err(err); }
};
```

`saved_blob_keys` and each `<field>_blob` local are declared outside the async block (captured by mutable ref), so the post-block `match` reads them after the borrow ends. This covers **every** `?` in the span (current and future) through one cleanup path without per-line wrapping, and re-returns the exact error the original `?` produced (same 422/DB/response mapping). Post-parse early returns (validation, `into_new`, current-row load, unique-violation, DB error, not-found) keep their existing cleanup-before-return. Non-attachment scaffolds emit byte-identical output; preserved existing blobs (`.or(current.<field>)`) are never touched; the success path never deletes.

Tests updated to assert the central block + coverage of `next_field`/`bytes_limited`/utf8/save/decode inside the span, and that `saved_blob_keys` is declared before the parse span (per-branch `store.delete` counts: create 6→5, update 8→7). Gates green locally: `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings` (the app.rs:5873 CI-only lint did not fire locally), `scaffold_form_for` suite, and the ignored `generated_attachment_scaffold_cargo_checks` compile-proof.

---

PR #1946, opened 2026-07-15
feat(a11y): typed TextArea/Select/Checkbox/FileField primitives + hx-* escape hatch (Part of #1706)

madmax983:
Part of #1706. Implements the primitives portion of #1933.

## Before / After

**Before:** the scaffold generator emits a labeled, accessible control only for single-line text fields (`TextField`). Every other field kind — multi-line text, dropdowns, booleans, file uploads — falls back to raw `html!` markup, which bypasses the type-level accessible-name obligation. It is trivially possible to ship a `<textarea>`, `<select>`, `<input type="checkbox">`, or `<input type="file">` with no associated label, and nothing catches it until an audit (or a user relying on assistive technology). Separately, `storage::form_helper::direct_upload_input` derives an `aria-label` from the field name that, per the ARIA accessible-name computation, *overrides* a caller's visible `<label>` — a documented caveat awaiting a typed file-field primitive.

**After:** `TextArea`, `Select`, `Checkbox`, and `FileField` join `TextField` as label-obligated typed primitives. Each is a typestate builder: constructing one yields a `<Primitive><NoLabel>`, which does **not** implement `maud::Render`. Only after attaching a label (`.label(..)` / `.aria_label(..)` / `.labelled_by(..)`) does it become a `<Primitive><Labeled>` that can render. An unlabeled primitive therefore has no `.render()` method — building it is a compile error (`E0599`), so the accessible-name obligation (WCAG 1.3.1 / 3.3.2 / 4.1.2) is discharged at compile time. `FileField` with a visible label emits **no** competing derived `aria-label`, resolving the `direct_upload_input` precedence caveat.

## How

- Each primitive mirrors the existing `TextField<State>` pattern exactly: `PhantomData` typestate; constructor + `label`/`aria_label`/`labelled_by` in the `NoLabel` impl (each returning `Labeled`); presentational/validation setters in the state-generic impl; `impl Render` only for `Labeled`. The `LabelSource` / `NoLabel` / `Labeled` types are reused across all five primitives, not redefined.
- `TextArea` carries its value as the element's **text content** (`<textarea>…</textarea>`), never a `value=` attribute. `Select` uses a companion `SelectOption` struct (named to avoid shadowing `std::option::Option`). `Checkbox` is bool-only (a nullable/tri-state boolean reuses `Select`) and emits the input before its `<label for>`. `FileField` carries no `value` (a file input cannot be repopulated on error).
- An `hx(name, value)` escape hatch is added to all five form primitives (including `TextField`) so the htmx `--live-validation` path keeps the typed label obligation. maud 0.27 requires static attribute names (`Named` parses an `HtmlName`, not a splice) and forbids control flow in attribute position, so a dynamic `hx-*` set can't be expressed in `html!` tag syntax. It is spliced via one centralized helper: because maud escapes every `>` to `&gt;`, the first literal `>` in a rendered element is always the end of its opening tag, so the escaped attributes are inserted there. An empty `hx` set renders the element byte-for-byte unchanged (existing `TextField` output is untouched).
- New trybuild compile-fail goldens (one per new primitive) prove an unlabeled primitive has no `.render()`. The compile-pass fixture and the runtime render integration tests are extended: value-as-text-content, option value/label/`selected` rendering, checkbox label association, the no-competing-`aria-label` file-field fix, and ordered / escaped / no-op `hx` rendering.

## Scope

This is the **primitives only**. The scaffold routing that consumes them (#1933's routing boxes in `autumn-cli/src/generate/scaffold.rs`) is a separate follow-up slice gated on PR #1935, so this PR intentionally does **not** close #1933.

CI note: trunk-dev may currently show pre-existing reds (doc-build / saas drift) that are cleared by the separate chore PR #1942 until this branch is rebased.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LcL5aB12Krn1HzL9cM6hYA)_

### DIFF
--- a/autumn/src/a11y.rs
+++ b/autumn/src/a11y.rs
@@ -522,6 +522,22 @@ enum LabelSource {
     LabelledBy(String),
 }
 
+/// Returns `true` when `name` is a valid `hx-*` attribute suffix: a non-empty
+/// run of ASCII lowercase letters, digits, or `-`.
+///
+/// This mirrors the attribute-name validation in [`crate::links`]
+/// (`is_valid_attr_name`) but narrows the alphabet to the `hx-*` suffix
+/// grammar. It deliberately excludes whitespace and `=` — which
+/// [`maud::Escaper`] does **not** escape — so a caller-supplied name can never
+/// break out of the `hx-{name}="…"` attribute into a separate, injected
+/// attribute.
+fn is_valid_hx_suffix(name: &str) -> bool {
+    !name.is_empty()
+        && name
+            .bytes()
+            .all(|b| b.is_ascii_lowercase() || b.is_ascii_digit() || b == b'-')
+}
+
 /// Splice a dynamic set of `hx-*` attributes into an already-rendered element.
 ///
 /// maud requires a *static* attribute name (its `Named` attribute parses an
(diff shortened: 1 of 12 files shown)


madmax983:
Addressed both P2s in 94b53fd: `Select::selected_value` now clears stale selections (exclusive single-select — it sets the matching option and clears every other option's flag); `hx()` validates the suffix against a safe `[a-z0-9-]` alphabet in `with_hx_attrs` to prevent attribute-name injection (invalid names trip a `debug_assert!` and are skipped in release, mirroring `links.rs`). Tests added.

---

Issue #1972, opened 2026-07-15
MCP: tool inputSchema degrades to a generic object placeholder without a manual OpenApiSchema impl; Query&lt;T&gt; can't express nested structures


madmax983:
## Context (builder feedback)

Real-world feedback from a builder who shipped an MCP server on autumn-web (`.mount_mcp("/mcp")`, `#[api_doc(mcp)]`, `RequireApiToken`; overall verdict positive — the "no second schema" promise from #1117 held for the plumbing). The quality gap: unless every arg struct has an `OpenApiSchema` impl, the advertised tool `inputSchema` collapses to a generic object placeholder, so the *semantics* of tool arguments end up living in the description text — exactly the drift-prone place the feature was designed to avoid. Compounding it, `Query<T>` extraction can't express nested sequences/maps, so builders fall back to comma-separated strings and JSON-in-a-string fields.

## What the code does today (trunk-dev @ 2cf9a2e)

**Schema fallback:**
- `build_input_schema` (`autumn/src/mcp.rs:437`) assembles the tool schema from `doc.query_schema` / `doc.request_body`, inlining component refs via `rewrite_refs` (`autumn/src/mcp.rs:487`). Any component name not in the registry resolves to `json!({ "type": "object", "title": name })` (`autumn/src/mcp.rs:503-506`).
- The OpenAPI generator has the same backfill for unregistered names (`autumn/src/openapi.rs:611-625`).
- `OpenApiSchema` (`autumn/src/openapi.rs:277-283`) is a manual trait: "A blanket default is not provided — routes whose types do not implement this trait simply emit a generic `object` placeholder referring to the type name." There is **no standalone derive**; the only automatic impls come from `#[model]` codegen (`autumn-macros/src/model.rs:3470-3487`) and the primitive macro impls (`autumn/src/openapi.rs:298-311`). Plain handler arg structs (query param structs, non-model request bodies) require a hand-written impl plus `OpenApiConfig::register_schema` (`autumn/src/openapi.rs:257`).
- Net effect: an MCP client calling `tools/list` sees `{"type":"object","title":"MySearchParams"}` for the `query`/`body` property — no fields, no types, no required list — unless the builder hand-maintains schema impls, i.e. the second schema is back.

**Query<T> expressiveness:**
[...]

PR #1983, opened 2026-07-15
feat(macros): OpenApiSchema derive for real MCP tool inputSchemas (part of #1972)

madmax983:
## Before / After

**Before:** an MCP tool's `inputSchema` degraded to the generic `{"type":"object"}` placeholder unless the arg struct hand-implemented `OpenApiSchema`.

**After:** `#[derive(OpenApiSchema)]` on a plain arg struct yields a real field-level JSON schema, back-filled into `openapi::generate_spec_at` via an inventory registry so the MCP catalog surfaces real fields.

## How

The derive reuses `#[model]`'s `emit_schema_fn_body` to generate the field-level schema body, so the derived and model-emitted schemas share one code path. Each derived struct registers a `DerivedSchemaDescriptor` via `inventory::collect!`, and `openapi::generate_spec_at` walks that registry to back-fill real inputSchemas. The trait is re-exported at `autumn_web::openapi::OpenApiSchema`.

**Part 2 — the `Query<T>` / `serde_urlencoded` nested-args problem — is intentionally out of scope for this slice and stays open on the issue.**

Part of #1972

---
_Generated by [Claude Code](https://claude.ai/code/session_01W1kjhJB6QjfaNMMjKhtDfb)_

### DIFF
--- a/autumn-macros/src/lib.rs
+++ b/autumn-macros/src/lib.rs
@@ -34,6 +34,7 @@ mod model;
 mod oauth2_callback;
 mod one_off_task;
 mod one_off_tasks_macro;
+mod openapi_schema;
 mod param_helpers;
 mod parse;
 mod paths_macro;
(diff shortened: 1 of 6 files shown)


madmax983:
Thanks — both P2s are correct, and I'm deferring both to the #1972 Part 2 follow-up rather than widening this first slice (which lands only the `#[derive(OpenApiSchema)]` → real-schema wiring):

- **serde renames / raw idents:** confirmed. It's a pre-existing gap in the shared `emit_schema_fn_body` that every `#[model]` schema already has (a `#[serde(rename_all = "camelCase")]` model advertises snake_case property names today), not something this derive introduces. No current derive consumer renames and no test asserts otherwise. The right fix is once in the shared helper — honoring field `rename` / `rename_all` / `r#` raw idents, with a renamed-struct schema test — so `#[model]` and the derive stay in lockstep. Tracked for Part 2.
- **same-name collision (`create::Args` vs `update::Args`):** also correct and inherent to the short-name `$ref`/component-name scheme the OpenAPI spec already uses; a proper fix needs a module-path-qualified component identity that route references also adopt. No same-named derived consumer exists yet. Tracked for Part 2 alongside the `Query<T>` / serde_urlencoded work.

---

PR #2011, opened 2026-07-16
feat(dist): prebuilt curl-able autumn CLI binaries + install script

madmax983:
## What changed

**Before:** the only way to get the `autumn` CLI was `cargo install` — a from-source compile. Poor onboarding DX ("you have to cargo install just to start an autumn project") and unworkable as the distribution mechanism for downstream CI (issue #1931's generated-app runners have no `autumn` on PATH).

**After:** `autumn` ships as a prebuilt, curl-able binary for macOS and Linux, built from tagged crate releases. Onboarding leads with:

```sh
curl -fsSL https://raw.githubusercontent.com/madmax983/autumn/trunk-dev/scripts/install.sh | sh
```

— a checksum-verified binary in `~/.local/bin/autumn`, no Rust toolchain, no compile. `cargo install` remains the fallback.

Closes #2005.

## How

- **New reusable workflow `.github/workflows/cli-release.yml`** builds the `autumn` binary for four targets — **`x86_64-unknown-linux-musl`**, **`aarch64-unknown-linux-musl`** (via `cross` on `ubuntu-latest`, static musl) and **`x86_64-apple-darwin`**, **`aarch64-apple-darwin`** (native `cargo` on `macos-latest`) — packages each as a **version-less** tarball `autumn-<target>.tar.gz` + companion `.sha256` (sha256sum on Linux, shasum on macOS), uploads them as workflow artifacts, and attaches them to the GitHub Release with `gh release upload --clobber`.
  - **Wired into the release pipeline (deterministic, tagged-releases-only).** `cli-release.yml` is a `workflow_call` reusable workflow invoked by a small additive **`binaries` job in `release.yml`**, which runs right after the GitHub Release for a `v[0-9]+.*` tag is created — i.e. only once the Publish Gate has passed and the release exists. It builds from that tag and uploads to that release. An `on: release` trigger would **never fire**: `release.yml` creates the release with the default `GITHUB_TOKEN`, and GitHub suppresses new workflow runs from `GITHUB_TOKEN`-raised events — so invoking it as a reusable job is the correct deterministic fix (thanks @chatgpt-codex-connector for catching this). `releases/latest/download/…` therefore always means *the latest released version* — no rolling trunk builds. `ci.yml` is untouched.
  - **`workflow_dispatch` remains a manual escape hatch:** supply a `tag` to (re)build a specific released tag and attach to its release, or leave it blank to build the current ref for validation (artifacts only, no release upload).
  - **Static musl** (Linux) so a single binary runs across distros with no glibc-version dependency; **version-less asset names** make the stable `releases/latest/download/<asset>` URL work alongside version-pinned URLs.
- **New `scripts/install.sh`** (`curl | sh`, POSIX `sh`): detects OS + arch via `uname` (Darwin/Linux × x86_64/aarch64 → the right target triple), resolves the download URL for `latest` or a pinned `AUTUMN_VERSION`/`--version`, downloads the tarball + `.sha256`, **verifies the checksum** (POSIX `read` + parameter expansion; sha256sum/shasum, no `awk`), extracts, and installs to `AUTUMN_INSTALL_DIR` (default `~/.local/bin`), printing a PATH hint if needed. `--help` prints from a heredoc; missing flag values fail cleanly via `err()`. Overridable base URL/target make it testable + mirror-friendly.
- **README** `Install a prebuilt binary (macOS & Linux)` subsection now leads with the curl one-liner, documents the direct latest/pinned URLs + the four target triples, and keeps `cargo install` as the fallback.

**Checksum/signature story:** every asset ships a sha256 the installer verifies. Cryptographic signing (cosign/minisign) is intentionally deferred — sha256 covers integrity now; signing layers on later without changing the URL contract.

**Windows:** deferred to a follow-up (tracked on #2005) — the shell installer doesn't serve Windows, and adding an `x86_64-pc-windows-msvc` artifact + a manual-download/PowerShell path is separable from the macOS/Linux curl story. Windows users build from source for now.

## Verification (local — no release tag exists yet, so CI-side builds are unexercised)

- Both workflow files parse (`yaml.safe_load`); `release.yml` exposes `release.outputs.tag` and its `binaries` job `uses: ./.github/workflows/cli-release.yml` with `tag: ${{ needs.release.outputs.tag }}`.
- `sh -n scripts/install.sh` clean; `cargo build -p autumn-cli --release --bin autumn` → `autumn 0.6.0` (confirms crate/bin name).
- **Install-script E2E** against a locally-built artifact served over `file://`, OS/arch **auto-detected** (no `AUTUMN_TARGET`): `latest` path → `checksum ok`, installs, exit 0; pinned `--version` path → exit 0; corrupted tarball → `checksum mismatch`, **exit 1** (verify fails closed). Darwin triple assembly confirmed via a forced `AUTUMN_TARGET=aarch64-apple-darwin` dry run.
- macOS/musl compiles run only on CI runners (no tag yet); the Linux host can't produce them locally (`cross` supplies the musl toolchain; macOS builds use the `macos-latest` runners CI already exercises).

## Follow-up

#1931 (wire `autumn a11y verify` into generated-app CI via this install script) is the first consumer and lands as a separate PR after this merges. Windows binaries are a separate follow-up (see #2005).

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -3,14 +3,20 @@ name: CLI Release Binaries
 # Builds prebuilt `autumn` CLI binaries for Linux + macOS (x86_64 + aarch64)
 # and attaches them — with sha256 checksums — to a GitHub Release.
 #
-# Binaries correspond to DELIBERATE, TAGGED CRATE RELEASES, not trunk. The canonical
-# trigger is `release: published`, which fires only when a release is cut from a
-# version tag (release.yml runs on `v[0-9]+.*` tags) — cutting a release is always an
-# explicit maintainer action. The binary is built from that tag's source, so
-# `releases/latest/download/...` always means "the latest RELEASED version"; there
-# are no rolling/nightly trunk-dev builds.
+# Binaries correspond to DELIBERATE, TAGGED CRATE RELEASES, not trunk. This is a
+# reusable workflow invoked by `release.yml` (its `binaries` job) right after the
+# GitHub Release for a `v[0-9]+.*` tag is created — so it runs only once the Publish
+# Gate has passed and the release exists, and it uploads to that release. Cutting a
+# release is always an explicit maintainer action; there are no rolling/nightly
+# trunk-dev builds, and `releases/latest/download/...` always means "the latest
+# RELEASED version".
 #
-# `workflow_dispatch` is a manual escape hatch only: supply a `tag` to (re)build a
+# It is deliberately NOT triggered by `on: release`: release.yml creates the release
+# with the default GITHUB_TOKEN, and events raised by GITHUB_TOKEN do not start new
+# workflow runs, so an `on: release` trigger would silently never fire. Invoking it
+# as a reusable job from release.yml side-steps that entirely and is deterministic.
+#
+# `workflow_dispatch` is a manual escape hatch: supply a `tag` to (re)build a
 # specific released tag and attach to its release, or leave it blank to build the
 # current ref for validation (uploads workflow artifacts, no release upload).
 #
@@ -22,8 +28,12 @@ name: CLI Release Binaries
 # natively on macOS runners. Windows is a documented follow-up (see #2005).
 
 on:
-  release:
-    types: [published]
+  workflow_call:
+    inputs:
+      tag:
+        description: "Release tag to build from and attach binaries to (e.g. v0.6.0)."
+        required: true
+        type: string
   workflow_dispatch:
     inputs:
       tag:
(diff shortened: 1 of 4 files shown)


madmax983:
Thanks — adopted both in ab75e6a:
- `--help` now prints from a heredoc (works under `curl | sh`), and missing `--version`/`--dir`/`--target` values route through `err()` with explicit `$#` guards.
- Checksum parsing now uses POSIX `read` + `${x%% *}` — no `awk` in the verify path.

Re-validated locally: `--help`/missing-value paths behave, and the install E2E still passes on the latest + pinned paths and fails closed (exit 1) on a corrupted download.

---

madmax983:
Confirmed and fixed in 132629d. You're right — `release.yml` creates the release with the default `GITHUB_TOKEN`, so an `on: release` trigger would never fire on the standard release path. Reworked `cli-release.yml` into a reusable (`workflow_call`) workflow and added a `binaries` job to `release.yml` that invokes it right after the release is created (post Publish Gate) — deterministic, no token/nesting guesswork. `workflow_dispatch` stays as the manual build-from-tag escape hatch.

---

PR #2015, opened 2026-07-16
fix(cli): commit deploy state markers atomically after the proxy flip

madmax983:
**Before:** after the health-gated kamal-proxy flip, `autumn deploy` wrote its on-disk state markers (`previous-release`, the `current` symlink, `live-slot`) as three separate SSH ops. If that sequence was interrupted after the flip (dropped connection, disk error), the proxy served the new/rolled-back release while the markers still described the old one — drift that makes the *next* deploy read a stale `live-slot` and mis-pick the idle slot.

**After:** the three post-flip marker updates are one `commit-markers` SSH transaction, `&&`-chained so it lands or fails as a unit, with each marker file written to a `mktemp` temp in the same directory and `mv`-renamed atomically (no torn single-marker writes; one round-trip instead of three). Cutover and rollback share a single `commit_markers_command` helper. Marker formats (`{slot}\t{port}`, `{dir}\t{slot}\t{port}`) and the reader are unchanged, and the op stays strictly after the flip so the existing "post-flip failure surfaces the raw error and runs no teardown" fail-safe is preserved.

**Scope (Part of #1938):** this collapses the drift *window*. Fully *eliminating* the residual drift — a reconcile at deploy-start that reads kamal-proxy's live target (queryable on-host via `kamal-proxy list`, mapped to a slot by the `public+1`/`public+2` port offset) and self-heals a drifted marker before slot-selection — is a larger, higher-risk change to the well-tested deploy planner and is left as a follow-up.

Part of #1938.

---
_Generated by [Claude Code](https://claude.ai/code/session_014XPpqH7hPPF3B29HZxGbAd)_

### DIFF
--- a/autumn-cli/src/deploy/exec.rs
+++ b/autumn-cli/src/deploy/exec.rs
@@ -697,12 +697,13 @@ pub fn first_deploy_ops(
 /// 7. run pending migrations BEFORE cutover (`AUTUMN_MIGRATE=1` one-shot),
 /// 8. bounded `/ready` poll on the candidate's separate loopback port,
 /// 9. health-gated proxy flip old→candidate (THE cutover),
-/// 10. record the previous-release marker (the release being replaced: its dir +
-///     live slot), BEFORE `current` moves off it,
-/// 11. promote `current` to the new release,
-/// 12. record the live slot marker,
-/// 13. drain (stop) the old release,
-/// 14. prune old releases beyond `keep_releases`, always protecting the releases
+/// 10. commit the state markers as ONE atomic remote op (#1938): record the
+///     previous-release marker (the release being replaced: its dir + live slot)
+///     BEFORE `current` moves off it, promote `current` to the new release, then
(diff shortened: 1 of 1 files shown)


madmax983:
Correct, and intentional for this PR's scope. Collapsing the three writes into one `&&`-chained transaction with atomic per-file renames narrows the window, but — as the PR description notes — it cannot make a symlink swap plus two file writes atomic *across all three states* over SSH; `&&` stops subsequent commands but can't roll back a completed `ln -sfn`. Truly eliminating the residual requires a reconcile at deploy-start: read kamal-proxy's live target (`kamal-proxy list` → `Target`, mapped to a slot by the `public+1`/`public+2` port offset) and self-heal a drifted `live-slot` marker before slot-selection. That reconcile is the documented follow-up (noted in the description), and I'll open it as a separate PR — it's the piece that actually closes #1938. Keeping this PR to the window-collapse so the higher-risk deploy-planner change lands separately with focused tests.

---

PR #2035, opened 2026-07-17
autumn-cli: SQLite dialect SQL emitter (declarative-schema slice 5)

madmax983:
Part of #1975 (declarative-schema epic). Slice 5 — dedicated per-dialect SQL emitters.

## Before / After
**Before:** `autumn schema diff` targeting SQLite refused every ALTER-family change — `ALTER COLUMN TYPE`, `SET`/`DROP NOT NULL`, `SET DEFAULT`, `ADD CHECK`, `ADD CONSTRAINT` — with an "unsupported on SQLite" error, because SQLite's `ALTER TABLE` has no such statements. Slice 4 was Postgres-first and explicitly deferred these to this slice.

**After:** those changes now emit a real SQLite migration using SQLite's standard table-recreate procedure, so a declarative-schema diff applies on SQLite. Postgres output is byte-for-byte unchanged.

## How
- **Additive dialect seam.** New `SchemaContext` carries the full desired + baseline table shapes (the per-change `SchemaChange` deltas don't, and the recreate needs a table's whole shape). New `emit_up_sql_with_context` / `emit_down_sql_with_context` dispatch on backend; `run_diff` passes the context it already has. The existing `emit_up_sql` / `emit_down_sql` functions and the `MigrationPlan` / `SchemaChange` public shapes are unchanged, so the parallel migrate/doctor lane is unaffected. The Postgres arm reuses the pre-existing rendering verbatim.
- **SQLite table-recreate.** ALTER-family changes are coalesced per table into a single recreate block: `CREATE {table}__autumn_new` with the target shape → `INSERT..SELECT` the columns common to both shapes → `DROP TABLE` the old → `RENAME` → recreate the target indexes → wrapped with `PRAGMA foreign_keys=OFF` … `PRAGMA foreign_key_check` … `PRAGMA foreign_keys=ON`. All of a rebuilt table's other same-plan changes fold into the one recreate, so there's no double emission.
- **Transaction semantics (documented in the emitted SQL).** The block must run inside one transaction (diesel wraps each migration in one). Because `PRAGMA foreign_keys` is a no-op inside a transaction, the emitted header notes the harness must ensure FK enforcement is disabled around the migration; `PRAGMA foreign_key_check` validates referential integrity after the row copy.
- **Refuse loudly.** Anything the recreate can't express safely (a missing table shape in the context) is refused via the new `EmitError::SqliteRebuildUnsupported` rather than emitting broken SQL. `SET NOT NULL` and NOT-NULL-without-default remain refused at the guard layer as before.
- **Deferred follow-ups from the #1975 slice-4 comment.** (1) A down-leg ordering regression test built through the real diff engine, pinning `DROP COLUMN target_id` before `DROP TABLE target` for the create-table + add-FK-column pattern. (2) Bounded collision-resistant FK-constraint naming: an engine-generated `{table}_{column}_fkey` over Postgres's 63-byte limit (or colliding after truncation) now gets a deterministic `truncate + short hash` name used identically in the up `ADD CONSTRAINT` and down `DROP CONSTRAINT`, instead of being refused.

## Testing
- `cargo fmt --all -- --check`, `cargo clippy -p autumn-cli --all-targets -- -D warnings` — clean.
- `cargo test -p autumn-cli` (4107 passed) and `cargo test -p autumn-schema-core` (16 passed).
- Per-dialect golden up/down SQL; coalescing (one recreate block per table); common-column copy (dropped columns excluded); a missing-context refusal; the deferred down-ordering regression; the deferred bounded-FK-name tests; and a **real in-memory SQLite** test that executes the generated recreate SQL and asserts the seeded rows survive with the new shape and a clean `integrity_check` / `foreign_key_check`.

## Scope
Touches only `autumn-cli/src/schema/diff.rs` and `mod.rs`. No migrate/doctor command-surface changes (parallel lane), no CHANGELOG / skills / version-bump edits.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XDEk2VWTaRngi5YF5wNyKs)_


madmax983:
Thanks for the review. Addressed:

**FNV-1a for bounded identifier hashing (high) — applied** in 715ec85. The over-63-byte FK-constraint suffix now uses a hand-rolled, dependency-free FNV-1a 64-bit hash instead of `DefaultHasher`, so the emitted identifier is deterministic across toolchains and machines (and the up `ADD CONSTRAINT` / down `DROP CONSTRAINT` names still derive identically). Good catch — this matches the determinism posture the rest of the diff engine already holds (BTree ordering throughout).

**`writeln!`/`write!` → `push_str` (medium, lines 2037 & 2497) — declining.** These emitters run once per migration generation, not on any hot path, so the formatting overhead is immaterial, and the surrounding module builds all its SQL with `write!`/`writeln!` into `String`. Line 2497 in particular is inside `render_create_table_body`, which is preserved byte-for-byte from the pre-slice-5 `emit_create_table` to guarantee Postgres output is unchanged — rewriting it to `push_str` would churn deliberately-stable code and leave the module's string-building style inconsistent for no measurable gain. Keeping the existing idiom.

---

madmax983:
Thanks @codex — all three addressed in 7e85929.

**Triggers/views not preserved on rebuild (P1) — advisory added.** Correct that `DROP TABLE` drops user-defined triggers and views and the recreate only restores indexes. Autumn's declarative model has no trigger/view concept and this is an offline emitter (no schema introspection), so it can't restore them and refusing every rebuild would gut the feature. Instead the rebuild block now carries an explicit `-- autumn-safety:` advisory that triggers/views are dropped and must be re-created in a manual migration — matching the existing invisible-facet advisory this engine already emits for `DROP TABLE` association FKs.

**Staging-name collision (P2) — now refused.** The emitter checks whether `{table}__autumn_new` collides with a known table in the diff context and returns `EmitError::SqliteRebuildUnsupported` with a clear message instead of emitting unappliable SQL. Added a test covering the collision.

**`foreign_key_check` overclaim (P2) — comment corrected + test hardened.** You're right that the pragma reports violation rows rather than raising, so run fire-and-forget it enforces nothing. The comment now states accurately that it only *reports* rows, that the recreate copies values verbatim and introduces no new orphans, and that the runner must inspect the output for pre-existing orphans. The in-memory SQLite test now actually `load()`s `PRAGMA foreign_key_check` and `PRAGMA integrity_check` and asserts them clean, instead of discarding the rows via `batch_execute`.

All local checks still green (fmt, clippy `-D warnings`, `cargo test -p autumn-cli`).

---

madmax983:
Follow-ups from the latest reviews — all landed on this branch; PR #2035 remains the single source of truth.

**Parser-invisible baseline facets preserved (P1) — fixed in e39214bd.** Correct and important catch. The recreate was building the new table from the parser's partial `desired` shape, so a rebuild triggered by one change could silently drop baseline facets the parser can't see — a non-convention default, a skipped CHECK/enum. The up-leg recreate shape is now built from `baseline` + the table's explicit diff deltas (new `baseline_with_changes_applied` helper), so unchanged baseline facets survive; this makes the SQLite recreate uphold the same "never destroy parser-invisible state" invariant the Postgres per-change `ALTER` path relies on. New test `sqlite_rebuild_preserves_parser_invisible_baseline_default` proves a baseline `DEFAULT 'draft'` and a table `CHECK` both survive an unrelated column-type rebuild, and the existing goldens are byte-unchanged (the fix only affects the parser-blind-spot case). The down leg already targeted the baseline shape.

**Triggers-vs-views advisory (accuracy) — corrected in 87e035b1.** `DROP TABLE` drops the table's triggers but does not drop views that reference it; the advisory now states that accurately (views are left dangling and can block the rename) instead of lumping both together as "dropped."

**On the earlier "committed 7650cfc / created a PR" summary:** that commit and that PR don't exist in this repo — verified against the GitHub API and every local/remote ref. There was nothing to merge; this branch already carries the equivalent, verified advisory fix. Noting it only so nobody hunts for a phantom PR.

Convergence: with the recreate now baseline-faithful, everything representable in the schema model — columns, PK (incl. `AUTOINCREMENT` high-water), indexes, defaults, CHECKs — is preserved across a rebuild. Triggers and views aren't in the schema model and can't be introspected by an offline emitter, so they stay documented `-- autumn-safety:` advisories; full fidelity there belongs to the later DB-introspection slice. Local checks are green throughout (fmt, clippy `-D warnings`, `cargo test -p autumn-cli`: 4111 passed).

---

madmax983:
Both P1s reproduced and fixed in 16ca5650.

**Down-leg rollback source (P1) — fixed.** Correct: the up-leg baseline-merge landed, but the down leg still copied from the partial `desired`, so a rollback omitted parser-invisible retained columns and lost their data. The down leg now sources from the post-migration shape (baseline + diff deltas) — the mirror of the up-leg fix. Reproduced against the pre-fix code (the rollback `INSERT` omitted the retained `status` column); after the fix it copies it. Pinned by `sqlite_down_rebuild_preserves_parser_invisible_column`.

**Dependent-view rename abort (P1) — fixed, and validated in-transaction.** The offline emitter can't see or recreate views, so it genuinely can't refuse-on-view-presence or preserve them without inventing introspection. But the failure is preventable without introspection via the canonical `PRAGMA legacy_alter_table=ON/OFF` around the rename — a blind, always-emitted pragma that makes the rename succeed and lets the view re-validate against the recreated table. I reproduced the exact `error in view posts_v: no such table: main.posts` against the pre-fix code, then confirmed the pragma is honoured inside a transaction (unlike `foreign_keys`): the new `sqlite_rebuild_survives_dependent_view` test runs the rebuild via `conn.transaction(...)` with a dependent view present and asserts it applies, the view still resolves, and the row survives.

Net: the recreate's copy semantics are now symmetric (up and down each source/target the real shapes) and view-safe. Local checks green — fmt, clippy `-D warnings`, `cargo test -p autumn-cli`: 4113 passed.

---

madmax983:
Fixed in d6d5fff3 (reproduced first).

**`sqlite_sequence` read before it exists (P2) — fixed.** Real regression from the high-water preservation. The capture was gated on the *target* being `AUTOINCREMENT`, but `sqlite_sequence` is created lazily, so a migration that introduces autoincrement (an `i32`→`i64` primary-key change) on a database with no `AUTOINCREMENT` table hit `no such table: sqlite_sequence`. I reproduced that exact failure, then gated the capture/restore on **both** the copy source and the target being `BigSerial`: when the source is already `AUTOINCREMENT` its `sqlite_sequence` row is guaranteed to exist (so the read can't fail), and when it isn't there's no prior high-water to preserve (the new table correctly starts fresh), so skipping is also correct.

New test `sqlite_rebuild_skips_sequence_when_source_not_autoincrement` runs an `i32`→`i64` PK rebuild on a fresh non-autoincrement DB and asserts it applies without error and the row survives. Existing goldens are byte-unchanged; local checks green (fmt, clippy `-D warnings`, `cargo test -p autumn-cli`: 4114 passed).

---

madmax983:
Good catch — fixed in 4991a930.

**FK-bound type changes on SQLite (P2) — fixed.** You're right that the recreate can express this. The plan already carries the real `AlterColumnType` alongside the `AlterColumnTypeBlockedByFk` marker, and the SQLite rebuild already applies it — the only thing blocking it was the guard firing on every backend. Gated `find_fk_type_change_block` to Postgres-only (matching the already-Postgres-only `find_non_implicit_conversion` / `find_identifier_limit_violation`), so SQLite FK-column (and referenced-key) type changes now generate a recreate that applies the new type and preserves the inline `REFERENCES`. Postgres still refuses — pinned by the existing two pg tests plus an explicit backend-contrast assertion.

Since SQLite (unlike Postgres) won't reject a type that no longer matches the referenced key, the recreate now emits an `-- autumn-safety:` advisory naming the FK column(s) and noting that referenced-key type-consistency is the author's responsibility. New tests cover the SQLite recreate + advisory, the plain-rebuild-omits-advisory case, and a real in-memory SQLite run that applies the FK-column-type-change recreate (FK preserved, row survives). Local checks green — fmt, clippy `-D warnings`, `cargo test -p autumn-cli`: 4142 passed.

---

PR #2036, opened 2026-07-17
schema migrate + doctor (declarative schema slice 6)

madmax983:
Part of #1975

## Before / After

Before, the declarative-schema loop stopped at authoring: `autumn schema diff --write-migration` (slice 4) could generate `migrations/<ts>_<name>/{up,down}.sql`, but there was no CLI way to *apply* those generated migrations end-to-end, and no way to diagnose whether a project's declarative-schema state was healthy.

After:
- **`autumn schema migrate`** applies whatever is pending against the configured database and then refreshes the checked-in snapshot to the freshly-applied state. Postgres is advisory-locked (concurrent migrators serialize cleanly); SQLite applies unlocked under the backend-flip feature (single-writer, no advisory-lock primitive). It provider-locks against the snapshot's dialect *before* applying anything, prints an applied-migration summary, and re-snapshots the declared models so the baseline advances.
- **`autumn schema doctor`** is a read-only health report over six checks — project scaffolding, snapshot presence, model-vs-snapshot drift, backend provider-lock, snapshot-dialect-vs-DB, and pending migrations — each `OK`/`WARN`/`ERROR` with a one-line remediation. It exits non-zero only on an actionable `ERROR`; a `WARN` alone (uncommitted drift, unreachable DB, non-Postgres skips) never fails, so doctor stays runnable offline. `--json` emits the same checks as a machine-readable array.

## How

- Consumes the existing diff/snapshot/parse public surface (`diff_schema` / `DiffOptions`, `SchemaSnapshot` / `write_snapshot` / `load_snapshot` / `ensure_backend_matches`, `parse_models_path`) and the runtime `autumn_web::migrate` harness (`run_pending_locked` / `run_pending_sqlite` / `pending_migrations`) via diesel's `FileBasedMigrations`.
- Provider-lock refuses cross-dialect *before* applying: `migrate` loads the snapshot and calls `ensure_backend_matches(detected_backend)`, returning an error with no apply on a mismatch.
- `-- autumn-safety:` advisories are inert SQL comments here — the destructive-change guards ran at diff time (slice 4), so migration files apply verbatim, exactly as `autumn migrate` applies them.
- A snapshot-write failure *after* a successful DB apply is surfaced as a warning (the migration did apply), never reported as an apply failure; the refresh is skipped with a note when the project has no declarative models.
- `doctor`'s check computation is a pure function over gathered facts (no I/O), unit-tested directly; all filesystem/DB/env I/O lives in a thin shell. Secrets are scrubbed on every path — the resolved DB URL is only ever passed as an argument, never interpolated into output.
- `schema/diff.rs` is untouched (a parallel lane owns it).

## SQLite backend-flip note

autumn-cli now carries the same mutually-exclusive SQLite flip as autumn-web. The **default build targets Postgres**; the SQLite lane builds with default features off. `mail` was moved out of the direct `autumn-web` feature list into a **default-on `postgres` feature bundle** (`postgres = ["autumn-web/mail"]`), because autumn-web's `mail` `DbSuppressionStore` is hard-typed to `Pool<AsyncPgConnection>` and cannot compile under the backend-flip. The default build's `autumn-web` feature set is therefore **byte-for-byte identical to before** — only the feature *graph* changed so the `sqlite` build can resolve a mail-free set. Three localized `#[cfg]` gates in the pre-existing `autumn-cli/src/doctor.rs` keep its single `lettre` reference (the alert-email mailbox check) out of the sqlite build, with a conservative syntactic fallback; `schema/doctor.rs` references no `mail`/SQLite symbol.

Exact both-lane commands (never `--all-features`, never pg+sqlite in one invocation):

```
# Default (Postgres) lane
cargo fmt --all -- --check
cargo clippy -p autumn-cli --all-targets -- -D warnings
cargo test -p autumn-cli

# SQLite backend-flip lane
cargo clippy -p autumn-cli --no-default-features --features sqlite --all-targets -- -D warnings
cargo test -p autumn-cli --no-default-features --features sqlite
```

## Tests

- **Postgres e2e** (`schema_migrate.rs`, `#[ignore]`, Docker/testcontainers): snapshot → diff → migrate → refreshed snapshot → second migrate no-op → doctor clean, asserting no credential leak throughout.
- **SQLite e2e** (`schema_migrate_sqlite.rs`, feature-gated, **no Docker**): applies against a temp-file SQLite database end-to-end; runs in the sqlite lane.
- **Doctor/migrate unit tests**: `compute_checks` over synthesized states (drift OK/WARN, provider-lock error, missing/unreadable snapshot, dialect-vs-db mismatch, pending reachable/unreachable, ordering, JSON serialization, ERROR→non-zero exit), provider-lock refusal, the default-build SQLite apply seam naming the feature, and models-path resolution.

## Verification

Both lanes verified locally on this branch (rebased on current `trunk-dev`):

- **Default (pg):** `cargo fmt --all -- --check` clean; `cargo clippy -p autumn-cli --all-targets -- -D warnings` clean; `cargo test -p autumn-cli` — lib 4115 passed, integration 222 passed (schema migrate/doctor unit tests all green), Docker e2e `#[ignore]`d.
- **SQLite flip:** `cargo clippy -p autumn-cli --no-default-features --features sqlite --all-targets -- -D warnings` clean; `cargo test -p autumn-cli --no-default-features --features sqlite` — the SQLite file-DB e2e test compiles and runs green (223 integration passed).

Note: an unrelated pre-existing hygiene test (`repo_hygiene::first_run_docs_match_current_release_line`) fails identically on `trunk-dev` — `docs/guide/getting-started.md` pins the CLI install at `0.5.0` while the README and workspace are `0.6.0`. It is independent of this slice (this branch touches none of those files) and is intentionally left for the docs/release lane.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UFWREUn6jHMrrsM6ovU2Hv)_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -417,6 +417,7 @@ dependencies = [
  "crossterm",
  "ctrlc",
  "diesel",
+ "diesel_migrations",
  "directories",
  "getrandom 0.4.3",
  "hex",
--- a/autumn-cli/Cargo.toml
+++ b/autumn-cli/Cargo.toml
@@ -24,22 +24,56 @@ path = "src/main.rs"
 # explaining the feature is disabled). Enabling it turns on `autumn-web`'s
 # `tls` feature, whose `autumn_web::tls::inspect_leaf` does the offline
 # certificate parsing.
-default = ["tls"]
+# Default = the Postgres CLI: `tls` (below) plus the `postgres` bundle, which
+# carries the pg-only `autumn-web` features that predate — and do not compile
+# under — the sqlite backend-flip. The DEFAULT build enables exactly the same
+# `autumn-web` feature set as before (nothing dropped); only the feature GRAPH
+# is refactored so the mutually-exclusive `sqlite` build can select a
+# sqlite-compatible set.
+default = ["tls", "postgres"]
 tls = ["autumn-web/tls"]
+# The pg-only `autumn-web` features the CLI enables that assume the Postgres
+# runtime and fail to compile under the `sqlite` backend-flip. Only `mail`
+# (its `DbSuppressionStore` is hard-typed to `Pool<AsyncPgConnection>`) is a
+# genuine breaker today; it is grouped here (rather than left unconditionally on)
+# so the `sqlite` build — which excludes the `postgres` bundle — resolves a
+# mail-free `autumn-web` set that mirrors what autumn-web's OWN sqlite CI builds.
+# The CLI's own source references NO `autumn_web::mail` symbol (only generated
+# string literals), so no source `#[cfg]` gating is needed. Kept in `default`,
+# so a normal `cargo build -p autumn-cli` is byte-for-byte identical to before.
+postgres = ["autumn-web/mail"]
 # Mirrors autumn-web's `acme` feature (issue #1608). The `autumn doctor` ACME
 # preflight checks compile in every build (they only need `std`), but enabling
 # this keeps CLI/library feature parity and turns on `autumn-web/acme` for any
 # future ACME-aware CLI tooling. Implies `tls` (ACME builds on it).
 acme = ["tls", "autumn-web/acme"]
+# SQLite backend-flip for the CLI's `autumn schema migrate` (issue #1975 slice 6,
+# on top of autumn-web's #1614 sqlite runtime lane). Non-default and MUTUALLY
+# EXCLUSIVE with the default Postgres backend: it turns on `autumn-web/sqlite`,
+# which flips autumn-web's `RuntimeConnection` type alias from Postgres to SQLite
+# for the WHOLE build graph, and pulls diesel's sqlite migration backend. Because
+# cargo feature unification is global AND the `postgres` bundle (a default
+# feature) pulls the pg-only `mail` path that does not compile under the flip,
+# the sqlite CLI MUST be built with default features OFF:
+#   `cargo build -p autumn-cli --no-default-features --features sqlite`
+# NEVER `--all-features` (co-enables the Postgres path and fails to compile), and
+# never co-built with the `postgres` bundle. Default builds target Postgres and
+# reference no SQLite symbol. (See autumn-web's `sqlite` feature for parity.)
+sqlite = ["autumn-web/sqlite", "diesel_migrations/sqlite"]
 
 [dependencies]
 autumn-web = { version = "0.6.0", path = "../autumn", default-features = false, features = [
-    "db", "htmx", "maud", "mail", "storage", "i18n", "telemetry-otlp", "redis", "cache-moka", "http-client", "reporting"
+    "db", "htmx", "maud", "storage", "i18n", "telemetry-otlp", "redis", "cache-moka", "http-client", "reporting"
 ] }
 autumn-schema-core = { path = "../autumn-schema-core" }
 base64 = "0.22"
 clap = { version = "4", features = ["derive", "env"] }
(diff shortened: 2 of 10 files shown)


madmax983:
Thanks for the review — all six findings are addressed in `8d64e801`.

1. **(HIGH) pending-migrations connectivity false-positive** — fixed. A missing/unreadable `migrations/` dir no longer short-circuits to `Reachable(empty)`. When a DB URL is configured the check now probes connectivity against a freshly-created, guaranteed-empty `tempfile::tempdir()` (not the shared temp dir), so `pending_migrations` still connects: zero available migrations → empty set on a reachable DB (correctly "up to date"), and an error on an unreachable/misconfigured one → `Unreachable`/WARN. No URL configured still skips with the existing WARN. Extracted a `probe_pending` seam and added a pg-lane unit test (`missing_migrations_dir_still_probes_the_database`) proving missing-dir + unreachable-URL → `Unreachable`.
2–5. **migrate notes/warnings → stderr** — the two NOTEs (no snapshot / no declarative models) and both snapshot-refresh WARNINGs now use `eprintln!`; genuine RESULT/success output (applied-migration lines, "Applied N migration(s)"/"Database already up to date", "refreshed schema snapshot", the doctor report, `--json`) stays on stdout. Integration assertions unaffected (they only assert success lines on stdout).
6. **`existing_models_path` dedup** — hoisted to `schema/mod.rs` (shared by `migrate` and `doctor`); both verbatim copies deleted.

Both lanes re-verified green — default pg and `--no-default-features --features sqlite` (fmt/clippy `-D warnings`/tests), with the sqlite file-DB e2e still passing. The only failing test is the pre-existing, unrelated `repo_hygiene::first_run_docs_match_current_release_line` (0.5.0-vs-0.6.0 docs pin), which is out of scope for this slice.

---

madmax983:
Thanks @codex — all three findings were real and are fixed in `d2b9cfe5`:

1. **migrate:** the snapshot no longer advances on an empty/no-op apply — it refreshes only after a real apply — so newly-edited, ungenerated model changes stay visible as drift on the next `schema diff` / `doctor` instead of being silently baked into the baseline.
2. **migrate:** the migration backend is now derived from the profile-resolved database URL (falling back to a profile-aware project default when no URL is configured), so `--profile <name>` selects the correct apply path / provider-lock instead of the ambient backend. With no `--profile` the resolution is unchanged.
3. **doctor:** the project/selected backend behind the provider-lock and pending-migration checks is resolved from the same profile-resolved context as the URL, so `schema doctor --profile <name>` diagnoses the profile it claims to.

A shared `backend_for_url` helper keeps migrate and doctor from deriving the backend inconsistently. Both lanes re-verified green: default (Postgres) and the `--no-default-features --features sqlite` flip (fmt, clippy `-D warnings`, and tests — including the SQLite file-DB migrate e2e).

---

madmax983:
Thanks @codex — good catch. Fixed in commit `04df33e6`: an unreadable or non-directory `migrations/` path (an I/O error, a permissions failure, or a regular file where a directory was expected) now returns a descriptive error naming the path and cause, instead of silently reporting "No migrations to apply" and exiting 0. An absent or genuinely-empty directory still no-ops without a DB call. The emptiness check is now a pure, unit-tested `classify_migrations_dir` (`Absent` / `Empty` / `HasMigrations` / `Unreadable`). Both lanes (default Postgres and `--features sqlite`) re-verified green.

---

PR #2047, opened 2026-07-18
Part of #1910: SQLite FTS5 full-text search (replace ASCII LIKE fallback)

madmax983:
## Before / After (what a searchable-repo caller experiences on SQLite)

**Before:** On the SQLite backend, `#[repository(..., searchable)]` fell back to an ASCII-only `lower(col) LIKE '%term%'` substring match over `SEARCH_FIELDS`, ordered `id DESC` (unranked). Non-ASCII case-folding did not work (`äpfel` never matched `Äpfel`). The `--search` / `#[searchable]` generator scaffold was **rejected at generate time** on SQLite, citing this issue.

**After:** SQLite searchable repositories run real **FTS5** full-text search — `unicode61` tokenization (full-Unicode case/diacritic folding, so `äpfel` matches `Äpfel`), **bm25** ranking with per-column weights derived from the `#[searchable(weight=...)]` priorities, and a **fail-closed, injection-safe** MATCH query builder. `--search` now scaffolds on both backends. Tenant / soft-delete / owner isolation is preserved exactly.

## How

- **FTS5 table = external-content virtual table.** The `AddSearch` migration's new SQLite arm emits `CREATE VIRTUAL TABLE "<t>__fts" USING fts5(<searchcols>, content='<t>', content_rowid='id', tokenize='unicode61')`, kept in sync with `AFTER INSERT/DELETE/UPDATE` triggers, backfilled via the FTS5 `'rebuild'` command; `down.sql` drops the triggers + table. The base table stays the source of truth (mirrors Postgres's `search_vector` generated column — the FTS objects live outside the model/schema surface).
- **Macro query swap.** The generated `search` / `search_page` / `search_page_scoped` SQLite `backend_select!` arms now join `"<t>__fts"` to the base table, match with `WHERE "<t>__fts" MATCH ?`, and rank with `ORDER BY bm25("<t>__fts", <weights>), id DESC`. The **tenant / soft-delete / owner predicates and the two-phase (SELECT ids → hydrate) shape are unchanged** — only the id-selection SQL + ORDER BY changed. bm25 weights: A=10, B=5, C=2, else 1 (higher priority → more-negative bm25 → sorts first).
- **Fail-closed sanitization.** New pure helper `repository::sqlite_fts5_match_query(&str) -> Option<String>`: tokenizes on Unicode whitespace, wraps each token in a quoted `"..."` phrase (embedded quotes doubled), joins with implicit-AND spaces. Every FTS5 operator (`AND`/`OR`/`NOT`/`NEAR`/`col:`/`*`/parens/quotes/`-`) becomes a literal term. Empty/whitespace/no-token input returns `None`, and the caller then yields an **empty page with no query run** — never an unfiltered scan.
- **Capability probe.** `libsqlite3-sys` is built bundled (amalgamation defines `SQLITE_ENABLE_FTS5`), and the `sqlite` cargo feature now pulls it in explicitly so FTS5 is available even under `--no-default-features --features sqlite`. `build_sqlite_pool`'s per-connection setup probes FTS5 (create + drop a throwaway `temp` fts5 table) and fails loudly with an actionable message if it's missing — **no silent LIKE downgrade**. The AddSearch migration's `CREATE VIRTUAL TABLE` is the hard stop at boot.

## Tenant-isolation guarantee

FTS5 indexes only the search columns, so the tenant/soft-delete/owner predicates filter the **base table** via the join, exactly as the LIKE path did. An adversarial test drives an FTS injection attempt (`secret`, `secret OR alpha`, `tenant_id:tenant-a`) from tenant B against tokens that exist only in tenant A's rows — the MATCH genuinely hits A's row, but the `tenant_id = ?` predicate filters it out, so B sees nothing.

## Tests

- **Macro golden** (`autumn-macros`): the three searchable-fork tests updated to assert the FTS5 MATCH + bm25 SQL, keeping the tenant / soft-delete / owner predicate assertions in the new golden output. Unit tests for `sqlite_fts5_match_query` (operator neutralization, quote-doubling, empty→None).
- **CLI** (`autumn-cli`): the two reject tests flipped to assert FTS5 DDL (vtable + triggers + rebuild, down.sql drops, no tsvector/GIN leak); a schema_edit unit test for the SQLite DDL builder.
- **SQLite runtime** (`autumn/tests/sqlite_searchable_repository.rs`): flipped the pinned case-folding test (`äpfel` now matches `Äpfel`); replaced the LIKE-metachar test with an FTS5 operator-sanitization + fail-closed test (`OR NEAR *` → empty, not everything); added a bm25 ranking-sanity test (title-field hit ranks above body-field hit); strengthened cross-tenant + owner-scoped isolation with injection probes.

## Verification run (locally, both lanes)

- `cargo fmt --all -- --check` — PASS
- `cargo test -p autumn-macros` — PASS (572)
- `cargo clippy --workspace --all-targets -- -D warnings` (pg lane) — PASS
- `cargo build -p autumn-web` / `cargo build -p autumn-cli` — PASS
- `cargo build -p autumn-web --features sqlite` — PASS
- `cargo clippy -p autumn-web --features sqlite --lib -- -D warnings` — PASS
- `cargo test -p autumn-web --features "sqlite,test-support" --test sqlite_searchable_repository` — PASS (7)

(clippy run with system stable clippy 0.1.94; CI targets 1.97 — the lints exercised, `doc_markdown` / `doc_lazy_continuation`, exist in both.)

## Docs follow-up (out of scope for this PR)

`docs/guide/sqlite-in-production.md` still asserts the old "FTS rejected at generate time / deferred (#1910)" claim (around L150 / L186 / L314) and `docs/guide/full-text-search.md` has no SQLite section. These should be corrected in the docs batch — this PR corrects only the source rustdoc.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hmi69LxECaxbHAuALnVB9T)_

### DIFF
--- a/autumn-cli/src/generate/migration.rs
+++ b/autumn-cli/src/generate/migration.rs
@@ -11,8 +11,8 @@ use super::dsl::parse_fields;
 use super::emit::Plan;
 use super::naming::pascal_to_snake;
 use super::schema_edit::{
-    MigrationShape, add_columns_down_sql_for, add_columns_up_sql_for, add_search_down_sql,
-    add_search_up_sql, detect_migration_shape, encrypt_columns_down_sql, encrypt_columns_up_sql,
+    MigrationShape, add_columns_down_sql_for, add_columns_up_sql_for, add_search_down_sql_for,
+    add_search_up_sql_for, detect_migration_shape, encrypt_columns_down_sql, encrypt_columns_up_sql,
     parse_model_search_config_for_table, remove_columns_down_sql_for, remove_columns_up_sql_for,
     singularize,
 };
@@ -147,12 +147,11 @@ pub fn plan_migration_with_options(
             encrypt_columns_down_sql(table, columns),
         ),
         MigrationShape::AddSearch { ref table } => {
-            // Postgres FTS (`tsvector` + GIN) has no SQLite equivalent in this
-            // slice; reject at generate time citing #1910 (issue #1614 AC #4)
-            // rather than emit Postgres-only DDL that breaks on SQLite.
-            if backend == autumn_web::config::DatabaseBackend::Sqlite {
-                return Err(super::sqlite_search_unsupported_error());
-            }
+            // Full-text search is backend-aware (issue #1910): Postgres emits a
+            // `tsvector` generated column + GIN index; SQLite emits an
+            // external-content FTS5 virtual table + maintenance triggers (see
+            // `add_search_up_sql_for`). Neither leaks DDL that breaks the other
+            // backend, so `--search` now works on both.
             let singular = singularize(table);
 
             // Collect all potential model file candidates in order of preference
@@ -213,8 +212,8 @@ pub fn plan_migration_with_options(
             };
 
             (
-                add_search_up_sql(table, &language, &fts_fields),
-                add_search_down_sql(table),
+                add_search_up_sql_for(backend, table, &language, &fts_fields),
+                add_search_down_sql_for(backend, table),
             )
         }
         _ => (String::new(), String::new()),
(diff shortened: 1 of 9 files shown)


madmax983:
@codex review

---

madmax983:
@codex review

---

madmax983:
@codex review

---

madmax983:
@codex review

---

PR #2080, opened 2026-07-19
Fix live_broadcast/auto_broadcast Docker-sweep test assertions

madmax983:
## Before / After

**Before:** the consolidated `integration_tests` "Run Docker-dependent tests" sweep deterministically fails four `live_broadcast`/`auto_broadcast` tests (newly reachable since `ws` was folded into the sweep). The failures are all test-only bugs:
- dom-id assertions used the underscore form `live_item` / `live_item-{id}` while the test's own `LiveItem::dom_id_for` and the product emit the hyphenated `live-item-{id}`;
- `no_broadcasts_attr_emits_nothing` moved the only `Channels` handle into `.scope(channels, …)`, dropping the sender so `rx.recv()` returned `Err(Closed)` immediately and tripped the negative assertion instead of pending → `Elapsed`;
- the `auto_broadcast` delete assertions matched the HTTP `format_value` delete-form shape `hx-swap-oob="delete:#{id}"`, but the SSE broadcast path emits the tombstone `<div id="{id}" hx-swap-oob="delete"></div>`.

**After:** the assertions match the product's actual behavior — the `live-item-{id}` dom-ids, the SSE `delete` tombstone envelope (attribute value `delete`, id as a separate attribute), and a cloned `channels` handle keeps the negative-assertion `recv` pending.

## How

- `live_broadcast.rs`: swap the underscore dom-id substrings to hyphenated form in `update_broadcasts_true_swap` and `delete_broadcasts_oob_delete` (matching `LiveFragment::dom_id_for`, mirroring the hyphen form asserted in `inline_broadcast_prefetch.rs`); clone the `Channels` handle into the `CURRENT_CHANNELS.scope` in `no_broadcasts_attr_emits_nothing` so the broadcast sender outlives `rx.recv()`.
- `auto_broadcast.rs`: replace each of the four delete-form assertions with a pair asserting `hx-swap-oob="delete"` and the `id="…-{id}"` attribute separately — the tombstone shape pinned by the product unit test `oob_envelope_delete_emits_tombstone`.

## Freeze-fit note

Test-only assertion corrections; product untouched; fixes a standing trunk-dev Docker-sweep breakage; OUTSIDE the pre-tag release set.

## Verification

Docker was unavailable in this environment, so CI is the runtime proof. The consolidated test binary compiles cleanly with the exact CI feature set: `cargo test -p autumn-web --test integration_tests --features "test-support,offline-sync,ws,mail,redis" --no-run`. `git diff --name-only` is exactly the two test files.

### DIFF
--- a/autumn/tests/integration/auto_broadcast.rs
+++ b/autumn/tests/integration/auto_broadcast.rs
@@ -273,7 +273,12 @@ mod tests {
             .expect("timeout waiting for custom update broadcast")
             .expect("recv error");
         let html_content = msg.into_string();
-        assert!(html_content.contains("hx-swap-oob=\"beforeend:#custom-posts-list\""));
+        // A topic-change re-insert publishes with the model's `insert_swap()`
+        // (not the create-path container), so the OOB selector comes from
+        // `BroadcastPost::insert_swap()` = Target(BeforeEnd, "#broadcast_posts-list").
+        // `OobSwap::Target::format_value` uses that selector verbatim and ignores
+        // the target id, so the SSE envelope is `beforeend:#broadcast_posts-list`.
+        assert!(html_content.contains("hx-swap-oob=\"beforeend:#broadcast_posts-list\""));
         assert!(html_content.contains(&format!("id=\"custom-post-{}", custom_post.id)));
         assert!(html_content.contains("world"));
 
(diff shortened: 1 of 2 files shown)


madmax983:
@codex review

---

PR #2203, opened 2026-08-15
perf: stop deep-cloning AutumnConfig on request paths

madmax983:
## The report

A downstream app (Nexus CRM, built on autumn-web) profiled its request path and found whole-`AutumnConfig` deep clones at **~650 allocations and ~101 KB per request — 30% of the request's allocations and 42% of its bytes**, with the suggested fix being "an `Arc<AutumnConfig>` inside `AppState`".

## What was actually happening

`AppState` does not deep-clone config when it is cloned — config lives in the extension map behind an `Arc`, so `AppState::clone()` is already a refcount bump, and #2199 made the framework's own ingress path allocation-free (`config_alloc_gate` pins a request at ≤ 288 blocks).

The clones come from explicit `AppState::config()` calls. `config()` returns an *owned* `AutumnConfig`, so each call deep-clones every section — measured at **64 allocations / 1,384 bytes** against a default config, more as an app's config grows — even to read a single `bool` or `usize`. Handlers pay that per request, and a handler reading two or three sections pays it two or three times over. ~10 such reads per request against a fatter config is exactly the reported shape.

`AppState::config_arc()` already exists and hands back the shared `Arc<AutumnConfig>`. What was missing was the code an app actually runs using it.

## Changes

Converted every request-path `config()` read to `config_arc()`, binding the handle once and borrowing sections off it (`&config.auth.password`) instead of moving an owned section out of a fresh deep clone:

- **generated auth handlers** (`autumn generate auth`, incl. the `--totp`, `--passkeys`, `--magic-link` and `--oauth` variants) — the largest surface, since every scaffolded app inherits it. Login and logout each dropped two whole-config clones; signup, reset-password, change-password, reauth, session tracking, magic-link and the OAuth redirect/callback each dropped one or more.
- **admin plugin** — the two `session.secure` reads.
- **`saas` starter** and the **`blog` / `saas` / `teams` examples**.
- **`docs/guide/authentication.md`** — now teaches `config_arc()` as the default for handlers, with `config()` framed as the per-boot accessor.

## Not changed

- No public signature moved. `config()` remains the per-boot owned-snapshot accessor, and apps calling it in their own handlers keep compiling — switching them to `config_arc()` is the fix.
- The framework ingress path, which was already clean.
- The one generated call site still on `config()`: the boot-time `remember_me_startup` hook, which runs once and needs an owned `RememberConfig`.
- `AppState` keeps config in the extension map rather than gaining a plain `Arc` field. That is deliberate: `app::build` re-inserts a mutated config after `build_state`, and plugins can install one later, so already-cloned states must read through the map to observe it. The map read is allocation-free and already delivers the refcount-bump property the report asked for.

## Regression guards

- New generator test `generated_handlers_read_config_through_the_shared_arc` renders each auth variant and fails if `state.config()` appears anywhere outside the startup hook. Verified it has teeth by reintroducing one call site and watching it fail.
- `config_signature_is_unchanged` now pins `config_arc`'s signature too, since generated code depends on it returning an owned `Arc` that outlives the borrows taken off it.

## Verification

- `./scripts/pre-push-check.sh` green — fmt, `clippy --workspace --all-targets -- -D warnings`, workspace test-target compile, doctests.
- Generator conformance (these actually compile a scaffolded app): `generated_auth_totp_cargo_checks`, `generated_auth_passkeys_cargo_checks`, `generated_auth_magic_link_cargo_checks`, `generated_auth_magic_link_with_totp_cargo_checks` — all pass. The `--oauth` borrow pattern is covered by `examples/blog`, which compiles in the workspace gate.
- `cargo test -p autumn-cli --bin autumn generate::auth` — 212 pass.
- `config_alloc_gate` (4) and `state::tests` (21) pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ReHEDzSmxgayGqsb9fWQgY)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Performance
 
+- **config reads on the request path:** generated auth handlers, the admin
+  plugin, the `saas` starter, and the `blog`/`saas`/`teams` examples now read
+  configuration through `AppState::config_arc()` instead of
+  `AppState::config()`. `config()` returns an owned `AutumnConfig`, so every
+  call deep-clones **every** config section — 64 allocations and 1,384 bytes
+  against a default config, and more as an app's config grows — even to read a
+  single `bool` or `usize`. On a handler that cost is paid per request, and a
+  handler reading two or three sections paid it two or three times over: a
+  downstream app profiling its request path measured whole-config clones at
+  ~30% of its per-request allocations and ~42% of its per-request bytes.
+  `config_arc()` hands back the shared `Arc<AutumnConfig>` the state already
+  holds, so the same read is a refcount bump and handlers borrow the section
+  they need off the handle (`&config.auth.password`).
+
+  Nothing about the framework's own ingress path changed — that was already
+  allocation-free as of #2199 — and no public signature moved: `config()`
+  remains the per-boot owned-snapshot accessor, and the one generated call site
+  that still uses it is the boot-time `remember_me_startup` hook, which needs an
+  owned `RememberConfig`. Apps calling `state.config()` in their own handlers
+  keep compiling; switching them to `state.config_arc()` is the fix, and
+  `docs/guide/authentication.md` now teaches that as the default. A new
+  generator test pins the emitted handlers to `config_arc()` so the deep clone
+  cannot reappear in scaffolded apps.
+
 - **jobs:** the Postgres job worker's claim query (`SELECT … FOR UPDATE SKIP
   LOCKED`) no longer scans and sorts the entire ready backlog for a queue
   before picking one row, for apps that don't configure `[jobs] queues`
--- a/autumn-admin-plugin/src/routes.rs
+++ b/autumn-admin-plugin/src/routes.rs
@@ -697,7 +697,9 @@ async fn model_create(
     if let Some(Value::String(secret)) = record.get("token") {
         // Mirror the session cookie's Secure flag so the reveal cookie is
         // accepted in both HTTPS and explicit HTTP-only deployments.
-        let secure_attr = if state.config().session.secure {
+        // `config_arc` reads through the shared handle: reading one `bool` must
+        // not deep-clone every config section on a request path.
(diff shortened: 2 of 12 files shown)


madmax983:
CI hit a failure in `Test (ubuntu-latest)`, in `pagination_cursor_proptest::signed_token_rejects_wrong_key`. It is unrelated to this PR's change — no pagination or HMAC code is touched here — but it is a real latent test defect rather than noise, so I fixed it in `a922977` rather than leaving it to keep flaking on trunk.

**Cause.** The forgery property guarded itself with `prop_assume!(key_a != key_b)`, then required that a token signed with `key_a` fail to verify under `key_b`. Raw slice inequality is not key inequality: HMAC-SHA256 zero-pads any key shorter than its 64-byte block, so `[]`, `[0]` and `[0, 0]` are **one key** and mint byte-identical signatures:

```
HMAC(key=[],      "payload") = f81a95af…0ea0f5f3
HMAC(key=[0],     "payload") = f81a95af…0ea0f5f3
HMAC(key=[0,0,0], "payload") = f81a95af…0ea0f5f3
```

Whenever proptest drew such a pair, the test demanded a token be rejected under the very key that signed it — impossible. CI's minimal input was exactly that: `key_a = [0]`, `key_b = []`.

`Cursor::encode_signed` / `decode_signed` behave exactly as HMAC specifies; nothing in `autumn/src/pagination.rs` changed.

**Fix.** The assumption now compares keys after stripping trailing zero bytes — the padded form HMAC actually uses — so "different key" means what the property needs. A companion `keys_differing_only_in_zero_padding_are_one_hmac_key` test pins the padding property directly, so the reason the assumption isn't simply `key_a != key_b` can't be lost to a later cleanup.

**Verification.** Replaying CI's reported seed (`cc b81df94f2429d130bdf8b15a1b37971151ead4e1b07aab3eface5f24329e68e5`) fails against the old assumption and passes against the new one, so this addresses the actual failure rather than papering over it. Suite green at `PROPTEST_CASES=8192`; fmt and `clippy -p autumn-web --all-targets -- -D warnings` clean.

Happy to split this into its own PR if you'd rather keep the perf change isolated.

---

Issue #2320, opened 2026-08-26
Docs & examples coverage audit (0.7.0)


madmax983:
# Autumn (`autumn-web` 0.7.0) — Docs & Examples Coverage Audit

Follow-up to #2099 (the 0.6.0 audit), re-run against the 0.7.0 release.

## Headline numbers

```
Distinct features/subsystems audited (N):     112
Guide coverage:   102.0 / 112 = 91.1%   (yes=1, partial=0.5)
Example coverage:  66.5 / 112 = 59.4%   (yes=1, partial=0.5)

New-in-0.7.0 features audited as distinct rows: 27
  • with NO guide:                1   (2 more only partial)
  • with NO example:             18   (3 more only partial)
  • both undocumented AND unexemplified: 0
```

**The 0.6.0 T1 tier is closed and did not reopen.** 0.7.0 shipped its guide
alongside almost every new capability: of 27 new-in-0.7.0 rows exactly one
(route-level SEO) has no `docs/guide/` narrative, and **no** new-in-0.7.0
feature is both undocumented *and* unexemplified — the defining failure mode of
the 0.6.0 audit (7 such features). Guide coverage moved 79.4% → 91.1% on a
matrix that grew by 27 rows.

**The gap has moved wholesale from docs to examples.** Example coverage is
essentially flat (56.5% → 59.4%) while the surface grew ~32%. Thirteen of the
0.6.0 audit's 35 gaps are fully closed and four more partially, but 0.7.0 added
eighteen new guide-but-no-example rows, so the *absolute* number of
unexemplified features went up. Every T1/T2 item below is now an example, not a
guide.

---

## Coverage matrix

Legend: **Guide/Example** = `yes` (1) · `partial` (0.5) · `no` (0). `[new]` =
notable new-in-0.7.0 capability. Paths relative to repo root; guides live in
`docs/guide/`.

### Routing / UI

| Feature/Subsystem | Guide | Example | Notes |
|---|---|---|---|
| HTTP routing & handler macros (`#[get/post/put/patch/delete]`, `routes[]`, `#[main]`) | yes — `getting-started.md`, `tutorial/02-*` | yes — hello | Core; exercised by every example |
| Path helpers & `link_to`/`button_to` | yes — `path-helpers.md` | partial — bookmarks, reddit-clone | Used, not showcased |
| Middleware & interceptors | yes — `middleware.md` | partial — all apps | `interceptor.rs` (OAuth) rustdoc-only |
[...]

PR #2325, opened 2026-08-26
docs(seo): add the SEO guide and wire route-level SEO into reddit-clone

madmax983:
Closes the **T1 gap** from the 0.7.0 docs & examples coverage audit ([#2320](https://github.com/autumn-foundation/autumn/issues/2320)): route-level SEO was the only new-in-0.7.0 feature with no guide. `autumn/src/seo.rs` was rustdoc-only and `docs/guide/` had **zero** `sitemap`/`robots` mentions, even though the `seo(...)` route attribute shipped in 0.7.0 and `examples/blog` already wired `SitemapSource` + `[seo] base_url`.

## `docs/guide/seo.md` (new)

Covers, in this order:

- turning the feature on with `[seo] base_url` (and what it actually mounts)
- the `seo(...)` route argument: every key, which macros accept it, string-literal-only, compile errors on unknown/repeated keys
- the `SeoMeta` extractor (infallible), its Open Graph / Twitter fallbacks, and the "values, not markup" contract
- refining attribute defaults from a database row in the handler
- where to render — one `(seo.render())` call in the layout `<head>`
- canonical URLs, and reading `base_url` once instead of cloning `AutumnConfig` per request
- `robots = "noindex"`, which half of the directive to pick (`follow` vs `nofollow`), and **exactly** which sitemap entries it filters: derived `#[static_get]` paths only, never `SitemapSource` entries, with the reason
- **why `Disallow` and `noindex` must not share a URL** — they cancel out, and a `noindex` behind `#[secured]` is unreachable for the same reason
- `[seo.robots]`: profile defaults, `allow_all`, `additional_rules`, `sitemap_url`
- `SitemapSource`: entry fields, reading a database from one (no `Db` extractor at router-build time), the **start-up-snapshot** semantics, the custom-`/sitemap.xml`-route escape hatch and its collision behaviour, what a `LIMIT` does and does not bound, and the 50,000-URL limit
- **giving `lastmod` the page, not one column** — the write-vs-derive tradeoff, with the query
- `hreflang` alternates and locale-prefixed sitemaps, including the two exclusions that keep the sitemap in step with the router
- static builds (`autumn build`), including that it never overwrites a file your own routes produced
- a test recipe, a deploy checklist, and a reference table

The guide is written in **ASD-STE100** Simplified Technical English as requested: short sentences, active voice, simple tenses, one instruction per sentence, one meaning per word. A note at the foot asks editors to keep that style.

## `examples/reddit-clone` — the runnable sample

| Part | File |
|------|------|
| DB-backed `RedditSitemapSource`, base-URL/canonical helpers, `summarize` | `src/seo.rs` (new) |
| `[seo]` + `[seo.robots]`, documented inline | `autumn.toml` |
| Attribute-only tags + canonical on a `#[static_get]` page | `src/routes/about.rs` |
| Attribute defaults refined from a row + canonical URL | `src/routes/posts.rs` (`show`), `src/routes/subreddits.rs` (`show`) |
| Attribute-only tags + canonical on index pages | `src/routes/posts.rs` (`front_page`), `src/routes/subreddits.rs` (`list`) |
| `noindex, follow` on a public, crawlable page — where the directive actually works | `src/routes/auth.rs` (`profile`) |
| `noindex, nofollow` on a `#[secured]` page, with a note on why the tag is belt-and-braces there | `src/routes/posts.rs` (`submit_form`) |
| One `SeoMeta::render()` in the shared `<head>` | `src/routes/layout.rs` |
| `posts.updated_at` advanced on every update path | `src/hooks.rs` |

**The sitemap.** Lists the front page, the community index, the communities and the posts. Each post's `<lastmod>` is *derived* rather than read from a column — `GREATEST` of `posts.updated_at`, the newest live comment, and the newest comment deletion — because a comment changes the page without touching the `posts` row. Ordering runs on that same derived value, so the entry cap keeps the genuinely freshest pages. Entry caps of 1,000 communities and 5,000 posts bound the number of URLs, **not** the work the query does; the constants say so and say when to stop building the sitemap at boot. The source builds its own pool from the config (no `AppState` exists yet when the framework collects entries) and degrades to an empty list rather than failing the boot.

`routes/layout.rs` gains `layout_with_seo(seo, …)`; the existing `layout(title, …)` now delegates to it, so every untouched page keeps its `<title>` and its `— Autumn Reddit` suffix.

Verify a running app:

```bash
curl http://localhost:3000/robots.txt
curl http://localhost:3000/sitemap.xml
curl -s http://localhost:3000/ | grep -E '<title>|og:|canonical'
```

## Tests

- `src/routes/layout.rs` — rendered-tag assertions: every declared tag, the OG/Twitter fallbacks, "unset value emits no tag", `noindex, nofollow` and `noindex, follow`, and the preserved plain-title path.
- `src/routes/about.rs` — the pre-rendered page carries its attribute tags and a canonical URL.
- `src/hooks.rs` — `before_update` advances `posts.updated_at` (on an edit, and with no field diff) and still re-slugs on a title change.
- `src/seo.rs` — `summarize` (whitespace collapse, word-break cut, multibyte boundary, blank input) and the canonical helper with and without a base URL.
- `tests/seo_pg_integration.rs` — `robots_txt_never_blocks_a_url_that_relies_on_a_noindex_tag` parses the committed `autumn.toml` and asserts both halves of the crawl/index contract (runs unignored). Behind `#[ignore]` (testcontainers): the URL set and absolute-URL invariant, the rendered `sitemap.xml`/`robots.txt`, and `lastmod_follows_comment_activity_not_just_post_edits`, which covers a live reply, a deletion, a no-comment post, and a `Subreddit`-typed comment sharing a post's id.

## Also updated

`skills/autumn-web/` (SKILL.md + `references/api-reference.md`) — the plugin documented the `seo(...)` attribute but nothing of the `sitemap.xml` / `robots.txt` half. `README.md`, `EXAMPLES.md`, `examples/reddit-clone/README.md`, and `CHANGELOG.md` under `## [Unreleased]`.

## Review rounds

Four rounds of Codex findings, all verified against the code before acting:

1. **`<lastmod>` reported the creation date forever** — nothing advanced `posts.updated_at`. Fixed in `PostHooks::before_update` using `ctx.now` (the determinism seam bans `chrono::Utc::now()`).
2. **`/about` had no canonical URL** despite the changelog claiming otherwise.
3. **Silent sitemap truncation** while docs claimed "every community and post" — caps kept, now logged and honestly documented.
4. **The hook covered only `repo.update`** — votes, comments and tags bypass it. Moved to deriving `lastmod` at read time rather than fanning timestamp writes across every subsystem.
5. **The guide's copyable query dropped the polymorphic discriminator**, and **comment deletions were invisible** (and could move `lastmod` *backward*). Both fixed; the live filter moved from the join to the aggregate.
6. **`Disallow` + `noindex` on the same URL** — the guide advised a combination that cannot work, and the example implemented it. Rewritten, and `profile` added as a demonstration that actually functions.

One suggestion was declined with reasoning on [its thread](https://github.com/autumn-foundation/autumn/pull/2325#discussion_r3867711671): pre-filtering the sitemap query's candidate set. The false claim it rested on is gone, but a denormalised page-modification column is the write-fan-out this PR deliberately moved away from, and a three-way candidate `UNION` is a query-optimisation lesson in a file whose job is SEO wiring.

## Checks run locally

`cargo check --workspace --all-targets`, `cargo test -p reddit-clone --lib` (54), `cargo test -p reddit-clone --test seo_pg_integration` (2 unignored; 3 testcontainer tests compile but **do not execute** — no Docker in this environment), `cargo clippy -p reddit-clone --all-targets -- -D warnings`, `cargo fmt --all`, and `check-plugin-freshness.sh`, `check-skill-version-markers.sh`, `check-release-notes.sh`, `check-docs.sh`, `check-examples.sh`.

https://claude.ai/code/session_01XnVwd4ffd6npS4obEMrjQe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the start-up-snapshot semantics and the custom-route escape hatch for larger
   or live sitemaps), the 50,000-URL limit, `hreflang` alternates on
   locale-prefixed sites, and static builds. `examples/reddit-clone` is now the
-  runnable sample: a database-backed (and deliberately bounded) `RedditSitemapSource` whose `<lastmod>` is derived from the whole page — the later of `posts.updated_at` and the newest live comment — rather than read from one column, `[seo]` +
+  runnable sample: a database-backed (and deliberately bounded) `RedditSitemapSource` whose `<lastmod>` is derived from the whole page — the latest of `posts.updated_at`, the newest live comment, and the newest comment deletion — rather than read from one column, `[seo]` +
   `[seo.robots]` in `autumn.toml`, `seo(...)` on the front page, the community
   index, the community page, the post page, and the about page, canonical URLs
   on each of them, `robots = "noindex, nofollow"` on the submit form, and a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the start-up-snapshot semantics and the custom-route escape hatch for larger
   or live sitemaps), the 50,000-URL limit, `hreflang` alternates on
   locale-prefixed sites, and static builds. `examples/reddit-clone` is now the
-  runnable sample: a database-backed (and deliberately bounded) `RedditSitemapSource`, `[seo]` +
+  runnable sample: a database-backed (and deliberately bounded) `RedditSitemapSource` whose `<lastmod>` is derived from the whole page — the later of `posts.updated_at` and the newest live comment — rather than read from one column, `[seo]` +
   `[seo.robots]` in `autumn.toml`, `seo(...)` on the front page, the community
   index, the community page, the post page, and the about page, canonical URLs
   on each of them, `robots = "noindex, nofollow"` on the submit form, and a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the start-up-snapshot semantics and the custom-route escape hatch for larger
   or live sitemaps), the 50,000-URL limit, `hreflang` alternates on
   locale-prefixed sites, and static builds. `examples/reddit-clone` is now the
-  runnable sample: a database-backed `RedditSitemapSource`, `[seo]` +
+  runnable sample: a database-backed (and deliberately bounded) `RedditSitemapSource`, `[seo]` +
   `[seo.robots]` in `autumn.toml`, `seo(...)` on the front page, the community
   index, the community page, the post page, and the about page, canonical URLs
   on each of them, `robots = "noindex, nofollow"` on the submit form, and a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **SEO guide and a runnable SEO example (#2320, T1 Gap 1):** `seo.rs` was
+  rustdoc-only and `docs/guide/` had zero `sitemap`/`robots` mentions, even
+  though the `seo(...)` route attribute shipped in 0.7.0. The new
+  [`docs/guide/seo.md`](docs/guide/seo.md) covers the `seo(...)` argument and
+  its keys, the `SeoMeta` extractor and its Open Graph/Twitter fallbacks,
+  canonical URLs, `robots = "noindex"` and exactly which sitemap entries it
+  filters, `[seo]`/`[seo.robots]` configuration, `SitemapSource` (including
+  the start-up-snapshot semantics and the custom-route escape hatch for larger
+  or live sitemaps), the 50,000-URL limit, `hreflang` alternates on
+  locale-prefixed sites, and static builds. `examples/reddit-clone` is now the
+  runnable sample: a database-backed `RedditSitemapSource`, `[seo]` +
+  `[seo.robots]` in `autumn.toml`, `seo(...)` on the front page, the community
+  index, the community page, the post page, and the about page, canonical URLs
+  on each of them, `robots = "noindex, nofollow"` on the submit form, and a
+  single `SeoMeta::render()` call in the shared layout. Covered by unit tests
+  in `routes/layout.rs`/`seo.rs` and a Postgres integration suite in
+  `tests/seo_pg_integration.rs`.
+
 - **`#[query_budget(N)]` — a compile-time, per-route database query budget
   (#1667):** declare a handler's query ceiling and the build fails when any
   statically reachable path can exceed it. The canonical case is the N+1 — a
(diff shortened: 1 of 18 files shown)


madmax983:
## `Test (macos-latest)` is failing, and it is not this PR's

**Failing check:** [`Test (macos-latest)`](https://github.com/autumn-foundation/autumn/actions/runs/33029350600/job/98381340450) on `666da4a`.

```
---- integration::rich_text::deeply_nested_blocks_render_in_linear_time stdout ----
panicked at autumn/tests/integration/rich_text.rs:502:5:
rendering 80000 bytes of nested blockquotes took 13.344120084s
test result: FAILED. 1737 passed; 1 failed; 278 ignored
```

### Why it is not this PR's

**It is red on the base branch.** `trunk-dev` at `bc99a4b8` — the exact base commit of this PR — failed [the same job](https://github.com/autumn-foundation/autumn/actions/runs/33014697098) on the same test, hours before this branch existed:

| | base `bc99a4b8` (trunk-dev, 22:22Z) | this PR `666da4a` (02:04Z) |
|---|---|---|
| Failing test | `rich_text::deeply_nested_blocks_render_in_linear_time` | identical |
| Message | `took 14.408409708s` | `took 13.344120084s` |
| Result line | `1737 passed; 1 failed; 278 ignored` | `1737 passed; 1 failed; 278 ignored` |

The passed/failed/ignored counts are byte-identical, which is the other half of the proof: this PR adds no test to that binary. `git diff --name-only origin/trunk-dev...HEAD` returns **zero files under `autumn/`** — the diff is `docs/`, `examples/reddit-clone/`, `skills/`, and three root markdown files. Nothing in it can reach the rich-text renderer.

### No fix exists to port

I searched open PRs and issues for this test and the macOS timing failure and found nothing, so there is no change to bring into this branch. Flagging it here rather than widening this PR.

### Proposed patch, for whoever owns it

The assertion is a wall-clock absolute (`elapsed.as_secs() < 5`) evaluated on an **unoptimised** build. The test's own comment says it is "sized to catch a return of quadratic behaviour, not to benchmark" — but an absolute ceiling cannot separate "quadratic" from "slow contended ARM runner in debug mode". The observed 13–14s sits between the linear case (milliseconds) and the pre-fix quadratic case (~111s), which is exactly the ambiguous band an absolute threshold cannot resolve.

Testing the *shape* instead removes the machine from the equation:

```rust
// Render 1/4 the depth, then the full depth. Linear work scales ~4x;
// the quadratic behaviour this guards against scales ~16x.
let quarter = "> ".repeat(10_000);
let full = "> ".repeat(40_000);

let t0 = Instant::now();
let _ = render_user_content_html(&quarter);
let baseline = t0.elapsed().max(Duration::from_millis(1));

let t1 = Instant::now();
let html = render_user_content_html(&full);
let elapsed = t1.elapsed();

assert!(
    elapsed < baseline * 8,
    "4x the input took {elapsed:?} vs {baseline:?} for the quarter \
     (>8x ⇒ super-linear); the renderer has regressed",
);
```

That keeps the regression guard (a return to O(depth²) fails it on any machine) while a uniformly slow runner scales both measurements together. Raising the constant ceiling would also unblock CI, but it trades away the guard's precision on fast runners.

### What I am doing

Re-running the failed job once, per the flake protocol — that is the one re-run this PR gets. Everything else on this head is green: 24 checks including `Lint`, `MSRV`, `SemVer check`, `Sim sweep`, `SQLite runtime`, `Edge capsule conformance`, `Documentation build`, `Downstream smoke test`, all five `Fuzz` jobs, and every drift gate. I will keep watching until this PR is green, merged, or closed.

---

Issue #1149, opened 2026-06-07
Add a PWA generator for an installable manifest and service worker


madmax983:
## Problem (the "So What?")

Shipping an installable app — add-to-home-screen, standalone window, an app icon,
a basic offline shell — currently requires an autumn user to hand-write a web app
manifest, author a service worker, generate a full icon set, and inject a pile of
`<head>` tags into their layout. None of that is autumn-specific; it is identical
boilerplate for every app, and getting the service-worker scope/headers wrong
silently breaks installability. autumn already has the pieces to make this trivial
— a shared `layout`, asset fingerprinting (`.autumn-manifest.json`), static asset
serving, and the single-binary embed goal (#1004) — but ships nothing that turns
those into an installable PWA. Making "installable app" a one-command outcome is
squarely on autumn's "time-to-deployed-app" mission and rides a clear industry
trend (Rails 8 now scaffolds PWA files by default).

## User Story

As a framework user, I want a single command to make my autumn app installable as
a PWA (manifest + service worker + icons + the right `<head>` tags), so that users
can add it to their home screen and get an app-like, basic-offline experience
without me hand-authoring web-platform boilerplate or debugging service-worker
scope.

## Acceptance Criteria

- [ ] A generator (e.g. `autumn generate pwa`, and/or an `autumn new --pwa` flag)
      scaffolds, at minimum: a `manifest.webmanifest`, a service worker
      (`service-worker.js` or equivalent), and a maskable icon set covering the
      sizes required for install (incl. 192px and 512px).
- [ ] The generator injects the required tags into the app's **shared** `layout`
      (the polished one `autumn new` ships, per #1130) — `<link rel="manifest">`,
      `theme-color`, `apple-touch-icon`, and a correct `viewport` — rather than a
      stub, and is idempotent (re-running does not duplicate tags or files).
- [ ] The manifest is served as `application/manifest+json` and the service worker
      is served from a scope that controls the whole app (root scope, or the
      `Service-Worker-Allowed` header set) so registration is not silently
      scope-limited.
- [ ] The generated service worker precaches an offline shell and serves it on
      navigation failure, so loading the app with the network offline renders a
      real page (not the browser error page) — verified by a test.
- [ ] PWA assets participate in the existing asset pipeline: they are served in
      dev and embeddable in the single binary (#1004) — a release build does not
      require shipping loose manifest/SW/icon files alongside the binary.
- [ ] A generated smoke/system test (reusing the shipped headless-Chromium system
[...]

Issue #1148, opened 2026-06-07
Add a first-class in-app notifications store with read/unread feed


madmax983:
## Problem (the "So What?")

Nearly every shipped web app has a notification bell: a persistent, per-user
feed with read/unread state ("3 new"). Autumn already ships the *ingredients* —
the `#[model]`/`#[repository]` macros for persistence, `channels` for realtime
push, and `mailer` for email — but no primitive that ties them together. Today a
framework user hand-rolls a notifications table, a model, a read/unread column, a
query, an unread-count query, and a feed view from scratch for an outcome that is
identical across apps. This is the same "primitive ships, the common composition
doesn't" gap we keep closing (forms, pagination, widgets). It is explicitly *not*
flash (#1008 — transient, single-request, no persistence, no per-user inbox) and
*not* mailer (email is a different channel; this is in-app).

## User Story

As a framework user building any app with accounts, I want a first-class in-app
notification store with read/unread state and an unread count, so that I can add
a "notification bell" feed without hand-writing a migration, model, and four
queries for an outcome that is the same in every app.

## Acceptance Criteria

- [ ] A persistent notification record exists per recipient with at minimum:
      `id` (i64), `recipient_id` (i64), `kind` (String), `payload` (JSON), and a
      nullable `read_at` timestamp; created via a generated migration that uses
      the existing migration tooling (no hand-edited SQL).
- [ ] A `Notifications` API (service or extractor, consistent with how `Session`/
      `Auth` are surfaced) provides: `notify(recipient_id, kind, payload)`,
      `list(recipient_id, ListQuery)` returning a `Page<Notification>` (reuses the
      shipped pagination), `unread_count(recipient_id)`, `mark_read(id)`, and
      `mark_all_read(recipient_id)`.
- [ ] `unread_count` and `list(unread_only=true)` ignore notifications whose
      `read_at` is set; `mark_read`/`mark_all_read` are idempotent (re-marking a
      read notification is a no-op, not an error).
- [ ] `notify` for a recipient with no prior notifications creates the row and is
      immediately visible to `list`/`unread_count` (read-path and write-path both
      work — verified by a test, not just compilation).
- [ ] A generator command (e.g. `autumn generate notifications`) scaffolds the
      migration + model + a minimal feed handler, registers routes, and the
      generated smoke test exercises notify → list → mark_read using the shipped
      in-process `TestClient`.
- [ ] Docs/example show wiring an optional realtime push over the existing
[...]

PR #2334, opened 2026-08-27
Add Web Push so notifications reach devices when the site is closed (#1392)

### LINKED ISSUE #1392: Add Web Push so notifications reach devices when the site is closed
## Problem (the "So What?")
Autumn ships the two ends of the re-engagement story but not the middle. The PWA generator (#1149) emits an installable manifest and a service worker, and the in-app notification store (#1148) persists a read/unread feed — yet a notification only ever reaches the user **while they already have the tab open**. #1148's sole "push" is a best-effort broadcast over `channels`, which needs a live SSE/WebSocket connection, and the #1149 service worker contains **no `push` event handler at all** (verified: zero push/VAPID code in `pwa.rs`). So the headline reason a developer installs a PWA — getting users *back* when the app is closed — is exactly the capability autumn can't provide. Today they must hand-roll VAPID key handling, RFC 8291 payload encryption, subscription storage, and the browser permission/subscribe dance themselves. No Rust web framework ships this first-class, and autumn is one generator away from owning it.

## User Story
As a framework user who has generated a PWA, I want to send a Web Push notification to a subscribed browser so that my users are re-engaged even when the app's tab is closed, without writing VAPID handshakes or payload encryption myself.

## Acceptance Criteria
- [ ] A first-class API mints/loads a VAPID key pair from config (e.g. an `autumn`-managed key or `[push]` config block); the public key is exposed to the client for subscription. Keys are loaded once at boot and a missing/invalid key fails fast with a clear error, never silently no-ops.
- [ ] A `PushSubscriptionStore` trait (in-memory default for tests, managed Postgres table for prod) persists a subscription's `endpoint`, `p256dh`, and `auth` keys bound to a principal/user id; storing the same endpoint twice upserts rather than duplicates.
- [ ] A built-in subscribe endpoint accepts a browser `PushSubscription` JSON payload and records it for the current authenticated principal; an unsubscribe/expiry path removes it. A `410 Gone` or `404` from the push service on send marks the subscription stale and prunes it (no infinite re-send to dead endpoints).
- [ ] A framework send API delivers an (title, body, optional url/icon) payload to one principal's subscriptions (and a fan-out helper for many); VAPID signing and RFC 8291 payload encryption are performed by the framework — the developer writes **zero** crypto.
- [ ] `autumn generate pwa` emits a service worker that includes a `push` event handler (renders the notification) and a `notificationclick` handler (focuses/opens the target url), plus the client-side subscribe snippet wired to the public VAPID key. Re-running on an existing project is idempotent / `--dry-run` honored, consistent with other generators.
- [ ] Composes with #1148: a documented example shows `notify(...)` also delivering a Web Push when the recipient has a subscription, with push failure never failing the in-app write (best-effort, mirroring #1148's channel-broadcast posture).
- [ ] End-to-end test (read AND write path): subscribe → send → assert the encrypted request was dispatched to the recorded endpoint with a valid VAPID `Authorization` header; a stale-endpoint (410) send prunes the subscription. Verified by test, not just compilation.
- [ ] Public API has doc comments with a runnable `rust,ignore` example; schema additions are additive (new `push_subscriptions` table); no breaking change to `autumn-web` surfaces; minor version bump.

## Success Metric
From `autumn generate pwa` to a real subscribed browser receiving a notification **with the tab fully closed**, a developer writes **zero** lines of crypto/VAPID/encryption code and ≤ ~5 lines of glue (config the keys, call `push::send`). Measured by a system test (headless browser) asserting the `push` service-worker event fires from a framework-sent notification while the page is not focused, and that VAPID + payload encryption are produced entirely by the framework.

## Out of Scope
- Native mobile push (APNs/FCM device tokens) — this slice is browser Web Push (VAPID/Push API) only.
- A general multi-channel `Notifiable` abstraction that fans one notification across mail + in-app + push — push is added as a delivery leg that *composes* with #1148; the unifying abstraction is a separate decision.
- SMS, email (already `mailer`'s job), and realtime in-page updates (already `channels`).
- Rich notification UX (actions/buttons, images, badging counts) beyond title/body/url/icon.
- Per-user notification preferences / quiet hours / batching — a follow-up once delivery exists.

## Gap Analysis
**Rails** has no first-class Web Push; apps reach for the `web-push` gem and hand-write the service worker. **Laravel** relies on a community channel (`laravel-notification-channels/webpush`). **Django** needs `django-webpush`/`django-push-notifications` (third-party). **Phoenix** ships nothing first-class. **axum / actix-web / rocket / loco.rs** give you handlers and nothing else — you own VAPID, encryption, storage, and the JS. So *no* mainstream Rust web framework closes this loop, and the broader ecosystem treats it as bolt-on. The autumn-shaped answer: because autumn already *generates* the PWA (#1149) and *stores* notifications (#1148), it can wrap the (mature, existing) Rust push/VAPID primitives into a `PushSubscriptionStore` + send API + a service-worker template change — turning three disconnected pieces into one working "notify users when they're away" capability. That vertical integration (generator + store + send + #1148 composition) is the differentiator competitors' bolt-on libraries can't match.

## Complexity Tier
L — new `push` module + `PushSubscriptionStore` and a managed migration in `autumn`, a subscribe/unsubscribe route, a service-worker + client-snippet change in `autumn-cli` (`pwa.rs`), and a documented `#1148` composition. Additive public API, additive schema, minor version bump; the cryptographic heavy lifting (VAPID, RFC 8291) is delegated to existing ecosystem primitives rather than built in-house.

madmax983:
Closes #1392.

Autumn already generated an installable PWA (#1149) and stored an in-app notification feed (#1148), but a notification could only ever arrive while the user already had the tab open. This adds the missing leg: the browser subscribes, the app calls `push.send(...)`, and the service worker raises a notification on the device — **with the tab fully closed**. The developer writes **zero** lines of crypto.

```toml
[push]
private_key = "…"                     # mint offline with `VapidKey::generate()`
subject     = "mailto:ops@example.com"
```

```rust
push.send(user.id, &PushMessage::new("Build failed", "main is red").url("/builds/42")).await?;
```

`autumn generate pwa` emits the service worker handlers, the client subscribe snippet, the `push_subscriptions` migration, and the router mount — so that config block plus the `send` call is the entire glue budget.

## What's in it

- **`autumn_web::push`** — VAPID key handling and ES256 JWT signing (RFC 8292), `aes128gcm` payload encryption (RFC 8291), a pluggable `PushSubscriptionStore` (in-memory + Postgres/SQLite), a `PushTransport` seam, and the `WebPush` send API with fan-out.
- **`push::router()`** — built-in `vapid-public-key` / `subscribe` / `unsubscribe` routes that resolve the caller server-side and `401` when they cannot.
- **`autumn generate pwa`** — service worker `push` + `notificationclick` handlers, a client subscribe snippet wired to the framework's own key endpoint, the `push_subscriptions` migration, and the router mount. Idempotent (with `--force`, as every generator), `--dry-run` honored, fully reversed by `autumn destroy pwa`.
- **Docs** — `docs/guide/web-push.md`, including the #1148 composition where the feed write is awaited and the push is best-effort.

**No new crate enters the dependency graph.** `p256` was already resolved for `jsonwebtoken`'s ES256 backend; `aes-gcm`/`hmac`/`sha2`/`base64` were already non-optional. Only a direct edge and p256's `ecdh` feature flag are new.

## The load-bearing test

`matches_the_rfc_8291_published_vector` feeds RFC 8291 §5's own inputs in and requires RFC 8291 §5's own output back, byte for byte. That is what proves the payload actually decrypts in a real browser rather than merely round-tripping against our own code. HKDF is separately pinned to RFC 5869 A.1, and the end-to-end test decrypts a dispatched body back with the receiving user agent's private key — written longhand, not by reusing the sending code — and verifies the VAPID signature against the public half.

## Failure posture

Deliberate, and tested:

- An unusable `[push] private_key` fails the **boot** — including a *blank* `AUTUMN_PUSH__PRIVATE_KEY`, which is preserved and refused rather than clearing the setting, because the commonest cause is a secret that failed to interpolate. Sending with no key is an error raised before any dispatch, never an `Ok` report of zero deliveries.
- `404`/`410` prunes the subscription so a dead endpoint is never re-sent to; a `5xx` or rate limit leaves it in place, because pruning on a transient failure would silently unsubscribe every user during an outage.
- One dead device never blocks a live one — devices are dispatched to concurrently (bounded), and per-device outcomes are reported through `PushDeliveryReport` rather than raised.

## Security

The endpoint is a client-supplied URL the framework later `POST`s to, so it is treated as one:

- The transport resolves the host, refuses any address on the SSRF deny-list, pins the connection to a checked address (closing DNS rebinding), declines redirects (closing `307`-to-internal), and **discards the response body unread** — it needs only the status, so a registered endpoint cannot stream unbounded memory into the process.
- Subscribe-time validation requires `https`, refuses IP-literal and loopback hosts (including the `localhost.` trailing-dot form), normalizes the URL so one device cannot become several rows, and caps its length.
- Moving an endpoint between accounts requires the request to present **both** stored keys — `p256dh` *and* `auth` — enforced in the same SQL statement so it cannot race. `p256dh` alone would not do: it is a public key, so requiring `auth` too means learning an endpoint and its public half still cannot take the row.
- Endpoint URLs are logged by origin only, never echoed back in an error, and the subscription types carry redacting `Debug` impls so a `tracing::debug!(?subscription)` cannot copy a send capability into an app's logs.

## Review history

Nine automated review rounds were answered on this branch, on top of four independent pre-PR review passes (security/crypto, API design, generator, test coverage). The most serious findings, all fixed here:

- A **deadlock in the `WebPush` extractor** — `extension_or_insert_with` holds the extensions write lock while calling its closure, and resolving the service reads other extensions off the same state, so any app declaring `WebPush` without registering one would have hung the request. Two integration tests take that default path specifically to guard it.
- Unbounded outbound response buffering, a blank env override silently disabling push, a sign-out hook that left the logout button dead, and a sign-out hook that revoked a `PushSubscription` shared with another signed-in account.

The review budget agreed for this PR (3 rounds, extended to 5 for genuine P1s) is spent. Seven remaining findings are tracked in **#2336** rather than pushed here, so the PR stops moving under the reviewer; the exceptions I continued to fix are argued individually in the comments.

## Also fixed

`cargo test --workspace` aborted with a stack overflow before ~5,600 `autumn-cli` tests could run: clap's derive macro expands the whole CLI into one `augment_args` frame that, unoptimized, exceeds the 2 MiB stack libtest gives a test thread. A workspace `.cargo/config.toml` sets `RUST_MIN_STACK`, so the fix travels with the code rather than living in a CI environment variable.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — clean, exit 0.
- `./scripts/check-docs.sh` — clean (it reproduces the CI `Documentation build` job exactly).
- `cargo test --workspace --doc` — clean.
- 115 push unit tests, 28 push integration tests, 78 pwa generator tests, plus the 78 existing `http_client` tests to confirm the new `discard_response_body` builder flag leaves the shared client untouched. Seven Postgres testcontainer tests are `#[ignore]`d in the house pattern and are swept automatically by CI's Docker step.

Additive throughout: a new `[push]` block, a new `push_subscriptions` table, new public API. No existing `autumn-web` surface changed, and the version is deliberately left at `0.7.0` with a `## [Unreleased]` entry per `CLAUDE.md`.

## Out of scope (by design)

Native mobile push (APNs/FCM device tokens), notification actions/images/badges, and per-user preferences or quiet hours.

https://claude.ai/code/session_01XNrTdqNn4zh2iW1aJZubZc

---
_Generated by [Claude Code](https://claude.ai/code/session_01XNrTdqNn4zh2iW1aJZubZc)_

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,18 @@
+# Workspace-wide Cargo configuration.
+
+[env]
+# `cargo test` on the `autumn-cli` binary aborts with a stack overflow without
+# this, in the very first test that calls `Cli::try_parse_from`.
+#
+# The cause is not the tests: clap's derive macro expands the whole `Commands`
+# tree — every subcommand, every argument — into a single `augment_args`
+# function, and this CLI has a lot of both. Unoptimized, that one frame is
+# larger than the 2 MiB stack libtest gives a test thread, so building the
+# command tree overflows before any assertion runs. Release builds inline it
+# away and never hit it, which is why only `cargo test` is affected.
+#
+# `[env]` (rather than exporting the variable in CI) keeps the fix with the
+# code: every `cargo test` invocation picks it up — local, CI, and the
+# pre-push gate alike — so the suite cannot pass on one machine and abort on
+# another. Raise it further if the CLI grows enough to overflow 16 MiB again.
+RUST_MIN_STACK = "16777216"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   generated snippet, which is what keeps push opt-in working under the
   production defaults without exempting the push routes from CSRF.
 
+  Because every endpoint is client-chosen, the delivery path is written for a
+  hostile one: a principal's devices are dispatched to concurrently (bounded),
+  so a device that accepts a connection and never answers cannot hold up its
+  live siblings; the transport reads only the status code and discards the
+  response body unread, so a registered endpoint cannot stream unbounded memory
+  into the process; and the generated snippet unsubscribes on sign-out, so a
+  subscription never outlives the session that created it.
+
   Failure posture is deliberate: a `[push] private_key` that is present but
   unusable fails the **boot** rather than leaving push silently dead, and
   sending with no key configured is an error raised before any dispatch, never
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`cargo test` on the workspace no longer aborts with a stack overflow:**
+  clap's derive macro expands the whole `autumn` CLI — every subcommand, every
+  argument — into a single `augment_args` function, and unoptimized that one
+  frame is larger than the 2 MiB stack libtest gives a test thread. The first
+  test to call `Cli::try_parse_from` therefore overflowed and took the process
+  down with `SIGABRT`, stopping the suite before ~5,600 other `autumn-cli`
+  tests ran. Release builds inline the frame away, which is why only `cargo
+  test` was affected. A workspace `.cargo/config.toml` now sets
+  `RUST_MIN_STACK`, so the fix travels with the code instead of living in a CI
+  environment variable — the suite cannot pass on one machine and abort on
+  another.
+
 - **TLS-enabled migrations no longer panic when applied from an app's own
   async `on_startup` hook:** the sync migration/wait-check path bridges to
   Postgres through diesel-async's `AsyncConnectionWrapper` whenever the
@@ -32,6 +44,63 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `multi_thread` tokio runtime, called directly from an async body.
 
 ### Added
+
+- **Web Push — notifications that reach a device with the tab closed
+  (#1392):** Autumn already generated an installable PWA (#1149) and stored an
+  in-app notification feed (#1148), but a notification could only ever arrive
+  while the user already had the app open. The new `autumn_web::push` module
+  closes that loop, and a developer writes **zero** lines of crypto: the
+  framework mints/loads the VAPID key pair, signs the ES256 identity JWT
+  (RFC 8292), performs the ECDH + HKDF + AES-128-GCM payload encryption
+  (RFC 8291), and dispatches to the push service. Configure `[push]
+  private_key`, mount `autumn_web::push::router()` (which `autumn generate pwa`
+  now does for you), and call `push.send(user_id, &PushMessage::new(title,
+  body).url(target))`.
+
+  `autumn generate pwa` now also emits a service worker with `push` and
+  `notificationclick` handlers, a client subscribe snippet wired to the
+  framework's own public-key endpoint, and a `push_subscriptions` migration —
+  all idempotent, `--dry-run`-honoring, and fully reversed by
+  `autumn destroy pwa`.
+
+  Subscription storage is hardened against the fact that an endpoint URL is a
+  *capability*: the subscribe boundary normalizes the endpoint (so one browser
+  can never become several rows), caps subscriptions per principal, and allows
+  an endpoint to move between accounts only when the request presents the
+  stored `p256dh` — which keeps the shared-device case working (the browser
+  returns the same endpoint **and** keys) while refusing a takeover by anyone
+  who merely learned the URL. The outbound transport resolves the endpoint host
+  and refuses any address on the framework's SSRF deny-list, pins the connection
+  to the checked address, and declines redirects, so neither a DNS record
+  pointing at `169.254.169.254` nor a `307` can steer the POST at an internal
+  host.
+
+  Failure posture is deliberate: a `[push] private_key` that is present but
+  unusable fails the **boot** rather than leaving push silently dead, and
+  sending with no key configured is an error raised before any dispatch, never
+  an `Ok` report of zero deliveries. A `404`/`410` from the push service prunes
+  the subscription so a dead endpoint is never re-sent to, while a `5xx` or
+  rate limit leaves it in place — pruning on a transient failure would silently
+  unsubscribe every user during an outage. Because the subscribe endpoint takes
+  a client-supplied URL the framework later POSTs to, it requires `https` and
+  refuses IP-literal and loopback hosts, closing the SSRF shape at the
+  boundary.
+
+  The encryption is pinned to RFC 8291 §5's own published test vector — fed the
+  RFC's inputs it must reproduce the RFC's output byte-for-byte — and the
+  end-to-end test decrypts a dispatched body back with the receiving user
+  agent's private key and verifies the VAPID signature, so what is asserted is
+  what a real browser would receive. Ships `RecordingPushTransport` so
+  applications can assert their own push behaviour the same way.
+
+  Adds **no new crate** to the dependency graph: `p256` was already resolved
+  for `jsonwebtoken`'s ES256 backend, and `aes-gcm`/`hmac`/`sha2`/`base64` were
+  already non-optional dependencies. All additions are additive — a new
+  `[push]` config block, a new `push_subscriptions` table, and new public API;
+  no existing `autumn-web` surface changed. Guide:
+  `docs/guide/web-push.md`. Out of scope by design: native mobile push
+  (APNs/FCM device tokens), notification actions/images/badges, and per-user
+  preferences or quiet hours.
 
 - **`#[query_budget(N)]` — a compile-time, per-route database query budget
   (#1667):** declare a handler's query ceiling and the build fails when any
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,7 +76,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   host.
 
   `[push]` honors the usual environment overrides
-  (`AUTUMN_PUSH__PRIVATE_KEY` and friends), and `subject` is validated at boot
+  (`AUTUMN_PUSH__PRIVATE_KEY` and friends) — with one deliberate divergence: a
+  *blank* private-key override is preserved and refused at boot rather than
+  clearing the setting the way other `AUTUMN_*` overrides do, because the
+  commonest cause is a secret that failed to interpolate, and clearing it would
+  silently disable delivery (and erase a good key from `autumn.toml`). The
+  `subject` is validated at boot
   against what RFC 8292 permits — a bare email address instead of a `mailto:`
   URI otherwise boots cleanly and has every delivery refused remotely. Because
   Autumn's CSRF layer rejects an unaccompanied POST and its cookie is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   pointing at `169.254.169.254` nor a `307` can steer the POST at an internal
   host.
 
+  `[push]` honors the usual environment overrides
+  (`AUTUMN_PUSH__PRIVATE_KEY` and friends), and `subject` is validated at boot
+  against what RFC 8292 permits — a bare email address instead of a `mailto:`
+  URI otherwise boots cleanly and has every delivery refused remotely. Because
+  Autumn's CSRF layer rejects an unaccompanied POST and its cookie is
+  `HttpOnly`, the public-key response carries the caller's CSRF token for the
+  generated snippet, which is what keeps push opt-in working under the
+  production defaults without exempting the push routes from CSRF.
+
   Failure posture is deliberate: a `[push] private_key` that is present but
   unusable fails the **boot** rather than leaving push silently dead, and
   sending with no key configured is an error raised before any dispatch, never
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -665,6 +665,7 @@ dependencies = [
  "opentelemetry",
  "opentelemetry-otlp",
  "opentelemetry_sdk",
+ "p256 0.13.2",
  "percent-encoding",
  "pin-project-lite",
  "postgresql_embedded",
--- a/autumn-cli/src/generate/commentable.rs
+++ b/autumn-cli/src/generate/commentable.rs
@@ -232,7 +232,7 @@ fn comments_table_state(project_root: &Path) -> (bool, bool) {
 /// `UNLOGGED` is a durability setting, not a different kind of object: the
 /// relation is still permanent and still occupies the name, so a later
 /// `CREATE TABLE comments` fails with "already exists" — confirmed against
-/// PostgreSQL, not assumed. Missing it made the scan report no table and emit a
+/// `PostgreSQL`, not assumed. Missing it made the scan report no table and emit a
 /// duplicate migration.
 ///
 /// `TEMPORARY`/`TEMP` is deliberately NOT here, and that is the more
@@ -247,7 +247,7 @@ const CREATE_VERBS: &[&str] = &["create table", "create unlogged table"];
 /// `DROP TABLE [ IF EXISTS ] name [, ...] [ CASCADE | RESTRICT ]` — the LIST is
 /// the point. Requiring `comments` immediately after the verb missed
 /// `DROP TABLE audit_log, comments;`, which really does drop it (confirmed
-/// against PostgreSQL). The replay then kept a table the database no longer
+/// against `PostgreSQL`). The replay then kept a table the database no longer
 /// has, the next scaffold skipped creating it, and every generated helper
 /// queried a missing relation — silently, until the first request.
 fn comments_drops(lowered: &str) -> Vec<usize> {
@@ -559,7 +559,7 @@ fn comments_creations(sql: &str) -> Vec<(usize, &str)> {
 ///
 /// Ordering is why. Diesel keys on the version prefix, and the shared table's
 /// version is `{timestamp}2` while a parent scaffolded in the same second is
-/// `{timestamp}` — so the parent migration runs FIRST, and PostgreSQL refuses
+/// `{timestamp}` — so the parent migration runs FIRST, and `PostgreSQL` refuses
 /// `CREATE TRIGGER` when its function does not yet exist. A parent that carries
 /// its own function cannot be ordered wrongly. The body may safely reference
 /// `comments` before that table exists: plpgsql resolves it at execution.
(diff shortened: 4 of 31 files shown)


madmax983:
## Automated-review budget spent (5 rounds)

Head is `0dc4697`. Five Codex rounds are addressed — **12 findings, every one verified against the code before being acted on**. The agreed budget for this PR was 3 rounds, extended to 5 because genuine P1s surfaced in rounds 3 and 4. That ceiling is now reached, so I'm standing down from further review rounds on this PR.

| Round | Findings | Highlights |
|---|---|---|
| 1 | 4 | CSRF token unreachable by the generated snippet (P1); `AUTUMN_PUSH__PRIVATE_KEY` inert (P1); stale schema snapshot; unvalidated `subject` |
| 2 | 4 + CI | Rotated-key lockout; comment-aware mount detection; per-principal cap; multi-address transport fallback — plus the `Documentation build` failure |
| 3 | 2 | Blank key override silently disabled push (P1); `WebPush::new` bypassed subject validation |
| 4 | 3 | Unbounded response-body buffering (P1); subscription outliving its session (P1); serial dispatch stalling on one dead device |
| 5 | 2 | `p256dh`-only ownership transfer; `503` response dropping the CSRF token |

Two were CI-blocking and are fixed. One (round 3's P1) was **self-inflicted by round 1's fix** — wiring the env override through a shared helper whose blank-value semantics silently cleared the key.

### Two corrections to the record

- Round 5 claimed the in-memory store already required both subscription keys. It did not — both stores had the same single-key predicate, so both were fixed. A store-specific claim in a review is a hypothesis to check, not a fact to route around.
- The `schema_keys_snapshot_guard` test documents regenerating with `--all-features`, but `CLAUDE.md` forbids that (it flips the `sqlite` backend). Following the test's own instructions **deletes 58 keys** for the optional sections and silently blinds the guard. I used the explicit feature list from `ci.yml`. Worth fixing those instructions separately — it isn't specific to this PR.

### Verification on `0dc4697`

`clippy --workspace --all-targets -- -D warnings` → 0 · `./scripts/check-docs.sh` → 0 · `cargo test --workspace --doc` → 0 · 114 push unit + 28 push integration + 77 pwa generator + 78 `http_client` tests green. Seven Postgres testcontainer tests are `#[ignore]`d in the house pattern and swept by CI's Docker step.

### From here

Further review findings go to a follow-up issue rather than more commits here, so this PR stops moving under the reviewer. CI is a separate matter: I'm still watching it, and a red check on this PR remains work-now.

---

madmax983:
Round 6 arrived after the budget was spent, so its two findings are filed as **#2336** rather than pushed here.

Both were verified against the code first; neither is a security defect and neither blocks this PR:

- `mailto:` subjects are validated by `@` position rather than parsed, so `mailto:ops team@example.com` passes boot and is then refused by push services (`autumn/src/push/config.rs:79`).
- `new URL()` in the generated `notificationclick` handler throws on a malformed `PushMessage::url`, after the notification is closed but before `waitUntil` — a dead click (`autumn-cli/src/generate/pwa.rs:396`). Reachable only via an app-authoring mistake, since the payload is encrypted by the app's own server.

Each has a proposed patch and a test to extend in the issue.

CI on `0dc4697` has nothing failed — all checks are green or still running. I'm still watching this PR: a red check or a merge conflict remains work-now, and only review findings are deferred.

---

madmax983:
Round 7: two findings filed to #2336, one fixed here in `898419a`.

**Fixed, as a deliberate exception to the spent budget** — because it is a defect *this PR introduced*, not a pre-existing gap the reviewer found:

The sign-out hook added in round 4 calls `preventDefault()` and then chains the logout form's submit off `autumnPushUnsubscribe()`. That awaits `navigator.serviceWorker.ready`, which **by specification never rejects** — it stays pending until a worker becomes active. On a page whose service worker failed to install, neither `catch` nor `finally` ever ran and **the logout button did nothing**. Blocking a sign-out is far worse than the stale subscription the hook exists to prevent (the push service prunes that with a `410` anyway). The submit is now raced against a two-second deadline.

Worth recording how that slipped through: the test guarding it was named `a_failed_unsubscribe_never_traps_the_user_in_their_session` and asserted `.finally(() => form.submit())` — the right invariant, an assertion that could not enforce it. The test now asserts the race and its bound, so it fails against the code it was originally written to protect.

The budget still stands for everything else. The other two findings (a cross-principal move bypassing the destination's subscription quota, and `send_many` still serial across principals) are in #2336 with proposed patches.

---

madmax983:
Round 8: one fixed here (`ac51c4f`), one filed to #2336.

**Fixed — another regression this PR introduced.** An origin has one `PushSubscription` shared by every tab, so with two accounts open the row belongs to whichever signed in last. The round-4 sign-out hook revoked that shared subscription unconditionally, on the strength of a `204` from a delete that is *deliberately* indistinguishable (it returns `204` whether or not the caller owned the row, so it cannot be used to probe endpoint ownership). A stale tab signing out of account A therefore took **account B's** endpoint down with it — B's next push `410`s, gets pruned, and B silently stops receiving until it opts in again.

Split into `autumnPushForget()` (drops the server row; what sign-out uses) and `autumnPushUnsubscribe()` (forgets *and* revokes; the explicit "stop notifying this device" opt-out, where revoking is the intent).

**Filed** — the multi-address transport fallback can re-send a non-idempotent POST when a failure happens *after* transmission, producing a duplicate notification. Narrower, and a trade-off rather than a broken flow; needs `is_connect()` plumbed through `ClientError`, or reqwest's own connection-level fallback. Details and two candidate fixes in #2336.

### On the pattern

Rounds 6, 7 and 8 each found something in the **previous round's fix** — round 7 in round 4's sign-out hook, round 8 in that hook again and in round 4's transport fallback. The findings have stopped converging, which is where continuing to push stops reliably improving the code.

So the line I'm holding: **regressions this PR introduced get fixed here** (two so far — a dead logout button, and this one), **everything else goes to #2336**. Five items are tracked there now. That keeps a real defect from shipping without letting review churn drive the PR indefinitely.

CI note: the `Lint` failure on `898419a` was exit 143 — the runner received a shutdown signal while compiling example crates, after `cargo fmt --check` had already passed and before clippy emitted any diagnostic. `ac51c4f` starts a fresh run, which supersedes it.

---

madmax983:
Round 9: one fixed here (`5d94288`), two filed to #2336.

**Fixed — this PR asserted a security property its own types violated.** The send path deliberately logs endpoint *origins* only, with a comment explaining that a push endpoint URL is a capability and logs travel further than the subscription table. `VapidKey` gets a redacting `Debug` for the same reason. But `BrowserSubscription`, `SubscriptionKeys` and `StoredSubscription` derived `Debug` and printed endpoint + `p256dh` + `auth` — together, everything needed to send arbitrary notifications to that device. One `tracing::debug!(?subscription)` in an app's own subscribe handler would put a working send capability in its logs.

All three now render the origin and redact the key material. A test pins it.

**Filed** — the mount scanner not recognising Rust string literals (same shape as the earlier comment-only finding, one layer out), and `autumnPushUnsubscribe()` skipping the local revoke when server cleanup rejects. I checked the second against `898419a`: the pre-refactor code had the identical flaw, so it is pre-existing rather than something the sign-out split introduced.

### Why this one crossed the line and those two didn't

The rule I've been holding is "a regression this PR introduced that breaks a working flow gets fixed; everything else is filed." Findings 6 and 7 are pre-existing gaps — verified, not assumed — so they were filed.

This one isn't a regression either, so it needed a different justification, and I want it on the record rather than looking like the rule quietly widened: **the code claimed something about itself that was false.** Not "a reviewer suggested an improvement" but "this PR says it never puts push credentials in logs, and its own types do." Shipping a false security claim is worse than the churn the budget protects against — same reasoning as the two regressions, different route to it.

Seven items now tracked in #2336. The line otherwise stands.

---

madmax983:
Round 10: nothing pushed. One finding filed to #2336, one to a new issue #2337.

**The P1 — push principals carry no tenant identity — is real, but it is not this PR's.** I verified the chain before deciding: tenancy is *row-level* (per `TenancyConfig`'s own docs), so tenants share one database; `push_subscriptions` has no tenant column; and principal resolution returns the scoped actor or session `user_id` verbatim.

But the same is true of **`notifications`** — `recipient_id -> BigInt`, no tenant column, and the word "tenant" appears nowhere in that module. It ships today. So:

- Namespacing only push principals leaves the identical shape in `notifications` while implying the class of bug was handled.
- It would **break composition**: `docs/guide/web-push.md` documents writing a feed entry and sending a push for the *same* principal. Tenant-namespace one side and they stop referring to the same user.
- It needs a schema change, a migration, and a decision about where the tenant lives — a maintainer's call on the tenancy contract, not a bug fix inside a feature PR.

Filed as **#2337** with what I verified and, importantly, what I could not: whether Autumn deployments actually produce colliding user ids across tenants. With one `users` table and a single id sequence, user `42` is one user and nothing collides. If ids are globally unique by construction, #2337 closes as working-as-designed — and that guarantee is worth writing down.

**The P2** — one malformed stored row aborting the whole send, because `list_for` collects into a `Result` — is filed as item 8 in #2336 with a patch. Real, and it cuts against the per-device failure posture, but it is a robustness gap rather than a regression, and "the fix is small" is not a criterion I get to use ten rounds in.

Neither met the two exceptions I've been holding to (a regression this PR introduced, or a claim the code makes about itself that is false), so neither was pushed.

---

madmax983:
## CI: `Test (macos-latest)` red on `5d94288` — `rich_text::deeply_nested_blocks_render_in_linear_time`

```
rendering 80000 bytes of nested blockquotes took 13.235303958s
```

**It is not a renderer regression.** Evidence, in the order I gathered it:

1. **The diff cannot reach the renderer.** No file under `markdown/`, `rich_text`, or any render path is touched. The only shared-surface changes are additive: `pub mod push;` in `lib.rs`, a prelude re-export, and an `#[allow(unknown_lints)]` in `pagination.rs`.
2. **The same render takes 0.02s on this branch's code.** Run locally on `5d94288` with `--features markdown --test-threads=1`: `1 passed … finished in 0.02s`. The test's own comment says a linear render is "milliseconds" and the pre-fix quadratic bug took ~111s. 0.02s is squarely linear — the guard is doing its job and the code passes it.
3. **The base branch is green.** `trunk-dev` at `69bb9d18` — this PR's exact base — last ran CI to `success`.

So the 13.2s is ~660× wall-clock inflation of a ~20ms operation on the runner, not a complexity change.

**What I think is actually happening, including the part that is on me.** The assertion measures *wall clock* but is making a claim about *algorithmic complexity*. Under `cargo test`'s default parallelism that is only as stable as the runner is idle, and this job ran the binary for 1170s with 2051 tests. This PR adds ~143 tests that do real P-256 ECDH, HKDF and AES-GCM work in an unoptimized build, in that same binary. I can't prove that tipped it, but it is the most plausible mechanism and I am not going to write it off as someone else's flake.

**What I am not doing:** touching that test. Loosening the threshold or skipping it would weaken a genuine quadratic-regression guard that exists because this was a real 111s bug, and it is not my test to relax to get my PR green.

**Next step:** re-running the failed job once — the single re-run these situations allow, so far unspent on this PR. The API currently refuses it (`403 This workflow is already running`, run `33083264731` still in progress), so it is queued and will be spent when the run completes. If it fails a second time the failure is deterministic under this branch's load, and I will bring a proposed patch for the test (measuring CPU time rather than wall clock, or marking it serial) rather than pushing a change to it unilaterally.

---

madmax983:
Round 11: one finding, filed as item 9 in #2336. Nothing pushed.

`is_valid_vapid_subject` trims before checking, but `build_request` signs the untrimmed `self.subject` — so `WebPush::new(…, "  mailto:ops@example.com  ", …)` passes validation and then signs a `sub` that carries whitespace. Verified at `service.rs:393` / `:530` against `config.rs:80`.

### Why I'm not invoking an exception for this one

This is the closest call so far, and it's worth being explicit rather than quietly deciding either way — because it's a gap in **my own** round-3 fix, and I could construct an argument for fixing it.

That argument: `validate_subject()` exists so an unusable `sub` fails as `InvalidConfig` before dispatch instead of as an opaque `report.failed`. For a padded subject it returns `Ok`, so the exact symptom that fix was added to eliminate comes back. A validator that checks a different string than the one it signs is a latent trap, and the fix is one `.trim()`.

Against it, decisively:

- **Nothing that worked breaks.** Before round 3 there was no validation at all and a padded subject failed remotely; it still fails remotely. The outcome is identical — only the error message is worse. That is an *incomplete* fix, not a *regression*, and my exception is for regressions.
- **The "false claim" exception was for a documented security property contradicted by the code** — the PR stating endpoints are logged by origin only while the types printed full send capabilities. Stretching it to cover "a validator has a normalization gap" is exactly the widening I warned against when I wrote the rule down.

My own test was: *if the argument feels strained, it is not an exception — file it.* It does. So it's filed, with the structural fix (normalize in the constructor, so validation and signing cannot diverge) and a note that the test should assert on the signed JWT rather than on the validator — the validator was never the thing that was wrong.

Nine items now in #2336, plus #2337.

---

madmax983:
## CI: `Test (windows-latest)` red on `40d1d2a` — Windows linker file lock

```
error: linking with `link.exe` failed: exit code: 1104
LINK : fatal error LNK1104: cannot open file 'D:\a\autumn\autumn\target\debug\deps\migrate.exe'
error: could not compile `bookmarks-sharded` (bin "migrate")
```

Not this PR's, and not a code failure:

- `LNK1104` on the **output** path is the linker being unable to open the file it is writing — the standard Windows file-lock symptom (a scanner or a stale toolchain process holding the handle). The job's own cleanup confirms one was around: `Terminate orphan process: pid (7596) (vctip)`.
- It is in `bookmarks-sharded`, an **example** crate. This diff touches no example crate — `git diff --name-only origin/trunk-dev...HEAD` returns nothing under `examples/`.
- It failed at **link** time, before any test body ran. Nothing in the suite was reached, so there is no test result to interpret.

Windows passed this same job on the previous head (`5d94288`), and the merge that produced `40d1d2a` changed only `config.rs` (keeping both features' env-override calls) and `CHANGELOG.md`.

**Next step:** re-running the failed job once — the single re-run these situations allow, still unspent on this PR. The API refuses it while run `33097819955` is in progress (`403 This workflow is already running`), so it is queued. If it fails a second time the failure is real and I will treat it as such rather than re-running again.

---

PR #2361, opened 2026-08-28
Cap workspace clippy parallelism so the Lint job stops being OOM-killed

madmax983:
## The failure

`Lint`'s first step, `cargo clippy --workspace --all-targets`, is intermittently killed on `ubuntu-latest`:

```
    Checking autumn-web v0.7.0 (/home/runner/work/autumn/autumn/autumn)
    Checking bench-runtime-gate v0.1.0 (…/benchmarks/runtime/gate)
##[error]Process completed with exit code 143.
##[error]The runner has received a shutdown signal. This can happen when the
runner service is stopped, or a manually started runner is canceled.
```

Always while `Checking autumn-web`, 60–90 s after the last compile line, **with no clippy diagnostic in the log**. With nothing to point at, it reads as infrastructure flake, so it has been re-run rather than fixed.

Four consecutive `trunk-dev` runs failed this way:

| Run | Commit | Last log line | Outcome |
|---|---|---|---|
| [33129145297](https://github.com/autumn-foundation/autumn/actions/runs/33129145297) | `61bdd9c` Web Push | `Checking autumn-web` 00:17:53 | exit 143 at 00:19:08 |
| [33123617901](https://github.com/autumn-foundation/autumn/actions/runs/33123617901) | `24bf151` examples | `Checking autumn-web` 22:46:05 | exit 143 at 22:47:32 |
| [33091693589](https://github.com/autumn-foundation/autumn/actions/runs/33091693589) | `eba7fc6` shadow build | — | failure |
| [33014697098](https://github.com/autumn-foundation/autumn/actions/runs/33014697098) | scaffold DSL | — | failure |

## The cause

Memory, not infrastructure. `lint` is `runs-on: ubuntu-latest` — 4 cores, 16 GB — and cargo defaults to one job per core. Checking the whole workspace with `--all-targets` peaks past 16 GB while `autumn-web`, the largest crate, is in flight; the runner is reclaimed and the step reports 143.

Reproduced on a 15 GB box, same toolchain CI pins (1.98.0):

| Command | Result |
|---|---|
| `cargo clippy --workspace --all-targets -- -D warnings` (default `-j 4`) | OOM-killed, no diagnostic |
| `cargo clippy -j 2 --workspace --all-targets -- -D warnings` | **exit 0**, clean |

That it is marginal rather than deterministic explains the shape of the failure: the same command succeeds on some runs (for instance [#2345](https://github.com/autumn-foundation/autumn/pull/2345)'s `Lint` passed at 02:33–02:53 while another branch's was dying), which is exactly why it looked like a flake worth re-running.

## The fix

`-j 2` on that one step. It trades some wall-clock for a step that currently fails outright.

Scoped deliberately to the workspace step — the one with evidence behind it. The two `-p autumn-web` clippy steps check a single crate, and because the job dies before reaching them they have never been observed to fail; capping them too would be speculation.

## Why it is worth fixing rather than waiting out

`test` declares `needs: [lint, meta]`, so while `Lint` is red the entire `test` matrix — **`ubuntu-latest`, `windows-latest` and `macos-latest`** — never runs. Every branch cut from `trunk-dev` inherits both the red check and the missing cross-platform coverage.

## Verification

- `cargo clippy -j 2 --workspace --all-targets -- -D warnings` — exit 0 on 1.98.0, the toolchain this job pins.
- Workflow parses: `yaml.safe_load` on `.github/workflows/ci.yml` succeeds.
- No change to what is checked — same command, same `-D warnings`, same crates and targets. Only the codegen job count differs.

https://claude.ai/code/session_01HUZJFJRqZvf4GQUbwdNd9x

---
_Generated by [Claude Code](https://claude.ai/code/session_01HUZJFJRqZvf4GQUbwdNd9x)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,20 @@ jobs:
       - name: Check formatting
         run: cargo fmt --all -- --check
       - name: Clippy
-        run: cargo clippy --workspace --all-targets -- -D warnings
+        # `-j 2` caps codegen parallelism. This step checks the whole workspace
+        # with `--all-targets`, and at cargo's default (one job per core, 4 on
+        # ubuntu-latest) its peak memory exceeds the runner's 16 GB while
+        # `Checking autumn-web` — the largest crate. The runner is then
+        # reclaimed mid-step and the job reports `exit code 143` with
+        # "The runner has received a shutdown signal" and no clippy diagnostic
+        # at all, which reads like infrastructure flake rather than what it is.
+        #
+        # It is marginal rather than deterministic — the same command succeeds
+        # on some runs — so it surfaced as an intermittent red that re-running
+        # sometimes cleared, on trunk-dev and on every branch cut from it.
+        # Reproduced locally on a 15 GB box: `--workspace --all-targets` is
+        # OOM-killed at `-j 4` and completes clean at `-j 2`.
+        run: cargo clippy -j 2 --workspace --all-targets -- -D warnings
       - name: Clippy (autumn-web, gated request-path features — panic gate)
         # The default-feature Clippy above does not compile the feature-gated
         # request-path modules (channels→ws, mail→mail, sync/*→offline-sync,


madmax983:
## `Test (macos-latest)` is red, and it is not this PR's — but it needs a fix

```
integration::rich_text::deeply_nested_blocks_render_in_linear_time
rendering 80000 bytes of nested blockquotes took 5.808296041s
test result: FAILED. 1783 passed; 1 failed; 290 ignored
```

**It cannot be this PR.** The entire diff is 14 lines of `.github/workflows/ci.yml`, adding `-j 2` to one clippy step in the `lint` job. It touches no Rust, and nothing in the `test` job's command line. There is no mechanism by which it changes how fast a markdown renderer runs.

### It is also not a renderer regression

I ran the same test on this workspace:

```
$ cargo test -p autumn-web --features markdown --test integration_tests \
      deeply_nested_blocks_render_in_linear_time
test integration::rich_text::deeply_nested_blocks_render_in_linear_time ... ok
finished in 0.02s
```

**20 milliseconds** — exactly what the test's own comment predicts ("a linear render of this input is milliseconds"). The quadratic behaviour it was written to catch took ~111 s. So 5.8 s is neither: it is a linear render on a machine that was not giving the thread much CPU.

That is the flaw. The assertion is wall-clock:

```rust
assert!(elapsed.as_secs() < 5, ...)
```

measured while `cargo test --workspace` runs the other 1783 tests in parallel on a 3-core macOS runner. It is not measuring the renderer, it is measuring the renderer *plus* whatever else the box is doing, against a ceiling only 16% above what it recorded.

### Why it is surfacing now

This leg has not run in days: `test` declares `needs: [lint, meta]`, and `lint` has been failing on `trunk-dev` since at least 2026-08-27. So this is not new — it is **newly visible**, because this PR is the first thing to get `lint` green and let the matrix run at all. Finding it is arguably the fix working.

That also means I cannot demonstrate it red on `trunk-dev` in the usual way: `trunk-dev`'s own macOS leg never starts.

### Proposed patch — not applied here

```rust
-        elapsed.as_secs() < 5,
+        // Wall-clock under `cargo test --workspace`, so this competes with
+        // ~1800 other tests for a 3-core runner. Sized to catch the return of
+        // O(depth²) — which took ~111s at this input size — not to benchmark a
+        // linear render, which is ~0.02s. A 5s ceiling failed on macOS at 5.8s
+        // purely from contention.
+        elapsed.as_secs() < 30,
```

30 s still trips on the 111 s quadratic case with room to spare, while tolerating a loaded runner. A sturdier alternative is to assert the *shape* — render at depth N and 2N and compare — but that doubles the test's cost and is still wall-clock underneath.

I have not applied it: it is unrelated code, and widening a CI-config PR to change a markdown test's threshold is the author's call, not mine. Say the word and I will push it here, or open it separately against `trunk-dev`.

I will re-run the macOS job once to confirm the contention reading as soon as the run finishes (`rerun_failed_jobs` is refused while the other legs are still going). If it passes on the re-run, that settles it.

---

PR #2365, opened 2026-08-29
🧼 Add `autumn db scrub` to anonymize data for non-production use (#1602)

### LINKED ISSUE #1602: Add `autumn db scrub` to anonymize data for non-production use
## Problem (the "So What?")

The single most common reason a team copies a production database is to reproduce a bug or rehearse a migration in staging/dev — and the moment #1595's `autumn db backup`/`restore` ships, that copy is one command away, PII and all, onto laptops and shared staging boxes. Autumn has every ingredient to prevent that leak and uses none of them for this job: `autumn db` offers only `create`/`drop`/`reset` (`autumn-cli/src/db.rs`), the GDPR registry's anonymize strategy is per-user erasure at runtime, not environment-level scrubbing (`autumn/src/gdpr.rs`), the PII scrubber only filters logs (`docs/guide/logging-pii.md`), and seeding is synthetic-only (`docs/guide/seeding.md`). Today the "safe" path is hand-rolled UPDATE scripts that rot the first time someone adds an `email` column, and the unsafe path is the default.

## User Story

As a framework user on a team, I want a one-command way to turn a production database (or backup artifact) into an anonymized copy safe for staging and dev, so that I can debug with realistic data without ever putting real customer PII on a non-production machine.

## Acceptance Criteria

- [ ] A documented `autumn db scrub` flow takes either a backup artifact (from #1595's `autumn db backup`) or a resolved database URL and produces a scrubbed database/artifact in which no PII-classified column retains its original values.
- [ ] Classification reuses what the framework already knows before asking the developer for anything: at minimum, `#[encrypted]` columns and tables registered with the GDPR anonymize strategy are PII-classified automatically; remaining columns are classified via a per-app declaration the developer maintains.
- [ ] Fail-closed: if any user-authored column is neither PII-classified nor explicitly marked safe, the scrub exits non-zero listing the unclassified columns — a newly added column can never silently pass through with real data.
- [ ] The scrubbed database still works: replacement values satisfy NOT NULL, uniqueness, and foreign-key constraints; `autumn migrate` reports a clean status against it; and the app boots and answers `GET /health` with 200 using it.
- [ ] Scrub refuses to write its output over the production target, using the same profile guard protocol as `autumn db drop` (`ProductionRefused` without `--force`).
- [ ] A CI round-trip test proves the contract: seed known PII values → backup → scrub → assert zero occurrences of the original values in any PII-classified column of the result, then boot the app against it.
- [ ] Docs: a "Refreshing staging from production" recipe (in `docs/guide/deployment.md` or a new guide) covers the backup → scrub → restore drill and the classification workflow, cross-linked from #1595's backup docs.

## Success Metric

A documented prod→staging refresh drill (backup → scrub → restore) completes in under 10 minutes on a fresh checkout, and the CI round-trip test proves 0 original values survive in any PII-classified column on every PR.

## Out of Scope

- Statistical/synthetic data generation and differential privacy (replacement values just need to be constraint-valid and fake).
- Row subsetting/sampling for very large databases (scrub the full copy first; sampling is a follow-up).
- Continuous/streaming sync into staging.
- Non-Postgres targets.
- Admin-plugin UI for running or auditing scrubs.

## Gap Analysis

Rails, Django, Phoenix, and Laravel ship nothing built-in; the demand shows up as ecosystem tooling — Greenmask, pgsync + obfuscation config, Snaplet (acquired precisely because this job is universal), the PostgreSQL Anonymizer extension, django-scrubber. Every one of them is schema-blind: the developer hand-maintains a column list in a YAML/SQL config that silently rots when the schema changes, which is exactly the failure mode that leaks PII. Autumn is the only framework in this comparison that already holds machine-readable PII semantics — `#[encrypted]` columns, GDPR anonymize registrations, the models themselves — so it can infer most of the classification and *enforce the rest fail-closed at the schema level*, turning "we think staging is clean" into a CI-verified invariant. Composed with #1595, autumn would ship the full loop (backup → scrub → restore) that peers need three third-party tools to approximate.

## Complexity Tier

**M** — new subcommand and round-trip tests in `autumn-cli` plus a classification surface fed by metadata `autumn` already exposes; no breaking public-API change to `autumn-web`, no migration burden for existing apps.

---

Implementation should target the **trunk-dev** branch, never `trunk`.

madmax983:
Closes #1602.

## Problem

The moment `autumn db backup` shipped (#1595), a production database became one command away from a laptop or a shared staging box — PII and all. The "safe" path was hand-rolled `UPDATE` scripts that rot the first time someone adds an `email` column; the unsafe path was the default.

Every peer tool for this job (Greenmask, pgsync + obfuscation config, the PostgreSQL Anonymizer extension, django-scrubber) is **schema-blind**: the developer hand-maintains a column list that silently drifts from the schema, which is exactly the failure mode that leaks PII.

## What this adds

`autumn db scrub` takes either a backup artifact or the resolved database URL and rewrites every PII-classified column with deterministic, constraint-valid fake values.

```sh
AUTUMN_ENV=prod    autumn db backup --keep 7
AUTUMN_ENV=staging autumn db scrub --artifact backups/prod/<run> --force
```

### Classification comes from the schema, not from a config file alone

Three sources, in precedence order:

1. **`[tables.<t>.pii]` in `scrub.toml`** — the explicit declaration, including the strategy.
2. **`#[encrypted]` model columns** — read by a new `schema::parse::parse_encrypted_columns`. An at-rest-encrypted column is PII by construction, so a `safe` declaration may *not* override it.
3. **GDPR `ModelRegistration::anonymize("t")` registrations** — read statically with `syn`, classifying that table's non-key columns. A non-literal table name is reported, never silently skipped.

Everything left over is **unclassified**, and that is a hard failure. Because the column universe comes from introspecting the **live database** (not the config file), a column added yesterday cannot be missing from it — so adding a column without declaring it breaks the scrub, which is the whole point. The refusal lists the columns and prints a paste-ready `scrub.toml` stanza; `autumn db scrub --check` writes nothing and exits non-zero, so CI can hold the line.

### Safety properties

- **Constraint-preserving.** PII on a primary- or foreign-key column is refused outright, so referential integrity is untouched; `null` is refused on `NOT NULL`; a constant-valued strategy is refused on a `UNIQUE` column; a `varchar(n)` bound narrows the generated token or refuses rather than truncating into collisions; `NULL`s stay `NULL`.
- **Unique per row, distinct per column, stable across runs.** Every replacement derives from an `md5` over the row's primary key salted with the column name.
- **Atomic.** Every target is classified before any target is written (an undeclared column on one shard cannot leave the topology half anonymized), and each database's statements run in a single transaction.
- **Production guard.** Writing refuses outside `dev`/`test` without `--force` — the same `ProductionRefused` protocol as `autumn db drop`. As defence in depth, it also refuses when the write target is the database an artifact's own non-dev/test profile *config file* declares.
- **Credential-safe.** No error or report line ever embeds a resolved URL.
- **Framework-aware.** Introspection excludes `autumn_*` tables from the classified universe, so the ones carrying app-supplied payloads (queued jobs, offline-sync rows, and `api_tokens` — production tokens inherited by staging are a live credential leak) are reported separately, and emptied when the app opts in via `[framework] purge`.

One honest caveat, stated loudly by the command and in the docs: with `--artifact`, the restore must run before the classification can read the schema it creates, so a refusal *after* a restore leaves unscrubbed data in the target. `--check` catches an incomplete declaration before any restore happens.

## How it was built

Red → green → refactor, as three separate commits:

| | |
|---|---|
| 🔴 `08bb540` | 47 tests written against the intended API — fail-closed classification, automatic classification and its precedence, constraint safety, replacement expressions, the production guard, the flag surface. Deliberately did not compile. |
| 🟢 `7fdf387` | The implementation. |
| ♻️ `f943159` | Error qualification, the two-pass classify-then-apply protocol, the post-restore warning, framework-table alignment, docs, CHANGELOG, CI wiring. |

## Testing

- **59 unit tests** in `autumn-cli/src/db/scrub.rs` and `autumn-cli/src/schema/parse.rs` covering classification, precedence, every refusal, every replacement expression, and the config schema.
- **`autumn-cli/tests/integration/db_scrub.rs`** — testcontainer round-trips that seed known PII, back it up, scrub it into a separate staging database, and then sweep **every character-typed column of every table** asserting zero occurrences of any original value; plus the fail-closed refusals (undeclared column, renamed column, production profile), `--check`/`--dry-run` writing nothing, the `--output` scrubbed artifact, and framework-table purge/warning.
- **`scrubbed_database_migrates_clean_and_boots_the_app`** — scaffolds a real app, migrates, seeds, scrubs, re-checks migrations, then compiles and boots it and asserts `GET /health` → 200.

### CI

`autumn-cli`'s Docker-gated tests are not auto-swept the way `autumn-web`'s are, so both lanes are named explicitly:

- `.github/workflows/ci.yml` — the round-trip suite in the existing "Run Docker-dependent tests" step (which already installs `pg_dump`/`pg_restore`), skipping the app-boot test.
- `.github/workflows/generator-conformance.yml` — the app-boot gate in the `scaffold-postgres` job, which has the `diesel` CLI; that job now also installs the Postgres client tools.

## Docs

New [`docs/guide/data-scrubbing.md`](docs/guide/data-scrubbing.md) with the backup → scrub → restore drill, the classification workflow, the strategy table, the guarantees and the limits. Cross-linked from the backup section of the daemon guide, a new "Refreshing staging from production" section in the deployment guide, and the README guide index. CHANGELOG bullet under `[Unreleased]`.

https://claude.ai/code/session_01TBfd5sAyAVnNFC3yVahkvT

---
_Generated by [Claude Code](https://claude.ai/code/session_01TBfd5sAyAVnNFC3yVahkvT)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -271,6 +271,19 @@ jobs:
           # above). Runs in the existing Docker-dependent Linux step rather than a
           # separate required gate.
           cargo test -p autumn-cli --test cli_tests -- --ignored offsite
+          # PII scrub round-trip (#1602): seed known PII -> `autumn db backup`
+          # -> `autumn db scrub --artifact` into a separate staging database ->
+          # assert ZERO occurrences of any original value in the result, plus
+          # the fail-closed refusals (undeclared column, renamed column,
+          # production profile). Needs Docker (testcontainers) + host
+          # pg_dump/pg_restore (installed above). autumn-cli's Docker-gated
+          # tests are NOT auto-swept the way autumn-web's are, so this line is
+          # the only thing that runs them.
+          # `scrubbed_database_migrates_clean_and_boots_the_app` is skipped
+          # here: it compiles and boots a generated app and needs the diesel
+          # CLI, so it runs in generator-conformance.yml's Postgres gate.
+          cargo test -p autumn-cli --test cli_tests -- --ignored db_scrub \
+            --skip scrubbed_database_migrates_clean_and_boots_the_app
           # Consolidated `integration_tests` binary: discover-and-run ALL the
           # Docker/testcontainer `#[ignore]`d DB tests instead of a hand-curated
           # per-feature allowlist (#1923). Enabling `test-support` (testcontainers)
--- a/.github/workflows/generator-conformance.yml
+++ b/.github/workflows/generator-conformance.yml
@@ -28,6 +28,13 @@ on:
       - "autumn-cli/tests/generate.rs"
       - "autumn-cli/tests/integration/console.rs"
       - "autumn-cli/tests/integration/scaffold_validation.rs"
+      # The `autumn db scrub` boot gate (#1602) lives in this workflow because
+      # it needs the diesel CLI and compiles a generated app. Without these
+      # paths a scrub-only change would run the ci.yml round-trip and skip the
+      # one test that proves the scrubbed database still boots an app.
+      - "autumn-cli/src/db/**"
+      - "autumn-cli/src/schema/**"
+      - "autumn-cli/tests/integration/db_scrub.rs"
       - "autumn/src/**"
       - "autumn/Cargo.toml"
       - "autumn-macros/**"
@@ -47,6 +54,13 @@ on:
       - "autumn-cli/tests/generate.rs"
       - "autumn-cli/tests/integration/console.rs"
       - "autumn-cli/tests/integration/scaffold_validation.rs"
+      # The `autumn db scrub` boot gate (#1602) lives in this workflow because
+      # it needs the diesel CLI and compiles a generated app. Without these
+      # paths a scrub-only change would run the ci.yml round-trip and skip the
+      # one test that proves the scrubbed database still boots an app.
+      - "autumn-cli/src/db/**"
+      - "autumn-cli/src/schema/**"
+      - "autumn-cli/tests/integration/db_scrub.rs"
       - "autumn/src/**"
       - "autumn/Cargo.toml"
       - "autumn-macros/**"
@@ -490,9 +504,15 @@ jobs:
       # The scrub gate below drives `autumn db backup`, which shells out to the
       # HOST pg_dump/pg_restore; ubuntu-latest does not guarantee them on PATH.
       - name: Install Postgres client tools
+        # Guarded: this job can route to the self-hosted runner, where
+        # `pg_dump` is already present and passwordless apt may not be — an
+        # unconditional install would take the four pre-existing gates in this
+        # job down with it.
         run: |
-          sudo apt-get update
-          sudo apt-get install -y postgresql-client
+          if ! command -v pg_dump >/dev/null 2>&1; then
+            sudo apt-get update
+            sudo apt-get install -y postgresql-client
+          fi
           pg_dump --version
 
       - name: generated_scaffold_serves_posts_index_and_json_api
--- a/.github/workflows/generator-conformance.yml
+++ b/.github/workflows/generator-conformance.yml
@@ -487,6 +487,14 @@ jobs:
       - name: Install diesel_cli
         run: cargo install diesel_cli --version "=2.2.0" --no-default-features --features postgres --locked
 
+      # The scrub gate below drives `autumn db backup`, which shells out to the
+      # HOST pg_dump/pg_restore; ubuntu-latest does not guarantee them on PATH.
+      - name: Install Postgres client tools
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y postgresql-client
+          pg_dump --version
+
       - name: generated_scaffold_serves_posts_index_and_json_api
         run: |
           cargo test -p autumn-cli --test generate \
@@ -522,4 +530,18 @@ jobs:
         run: |
           cargo test -p autumn-cli --test cli_tests \
             integration::generate_json_postgres::scaffolded_json_field_generates_compiles_inserts_and_selects \
+            -- --ignored --exact
+
+      # ── `autumn db scrub` boots the app it scrubbed (#1602) ──────────────
+      #
+      # The one AC the general Docker sweep cannot cover: after a scrub, the
+      # scrubbed database still MIGRATES CLEAN and a real generated app boots
+      # against it and answers GET /health with 200. Needs Docker + the diesel
+      # CLI + host pg_dump, and compiles a generated app — hence this job
+      # rather than ci.yml's sweep. Same `cli_tests` caveat as the blocks
+      # above: named explicitly here or it never runs anywhere.
+      - name: scrubbed_database_migrates_clean_and_boots_the_app
+        run: |
+          cargo test -p autumn-cli --test cli_tests \
+            integration::db_scrub::scrubbed_database_migrates_clean_and_boots_the_app \
             -- --ignored --exact
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   integrity survives. Writing refuses outside `dev`/`test` without `--force`,
   the same guard as `autumn db drop`, and every statement for one database runs
   in a single transaction. `--output` re-dumps the scrubbed database as a fresh
-  artifact, closing the backup → scrub → restore loop. See the new
+  artifact, closing the backup → scrub → restore loop.
+
+  The safety work is where most of the substance is. `#[encrypted]` columns are
+  **re-encrypted** under the target's own key rather than overwritten with a
+  plain string (which would make every later read of that row fail as malformed
+  ciphertext). PII is refused on either side of any foreign key — composite keys
+  included — so a natural key another table references is protected, not just
+  the referencing column; on `CHECK`-constrained columns, where no fabricated
+  value can satisfy the predicate; and on generated columns, which Postgres
+  refuses to update at all. Uniqueness is read from every unique index, partial
+  and composite included, and a unique column gets a wider `sha256` token so a
+  narrow `varchar(n)` cannot truncate into collisions. Statements are
+  `public`-qualified with a pinned `search_path` (a tenant `search_path` cannot
+  redirect a write), the planned tables are locked for the transaction, row-level
+  security and non-`public` schemas are refused rather than silently
+  under-scrubbed, materialized views are refreshed in dependency order, and the
+  framework-owned tables that keep verbatim copies of app rows — the ledger, the
+  version history, the search index, `api_tokens` — are reported and can be
+  emptied with `[framework] purge`. See the new
   [Data Scrubbing guide](docs/guide/data-scrubbing.md).
 
 - **`autumn upgrade` now reconciles framework-owned scaffold files, not just
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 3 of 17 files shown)


madmax983:
## Review rounds: standing down at the agreed budget

This PR was given a budget of up to 5 automated review rounds (3, extended to 5 because genuine P1s kept surfacing). That budget is now spent, and every finding from rounds 1–5 has been fixed and pushed:

| Round | Findings | Outcome |
|---|---|---|
| 1 | 1×P1, 3×P2 | All fixed — `#[encrypted]` re-encryption, FK both sides, plan-time type gates, unique-token entropy |
| 2 | 3×P1 | All fixed — unknown-provenance guard, purge-table locking, in-transaction matview refresh |
| 3 | 2×P1, 1×P2 | All fixed — feature-flag identity tables, privilege-hidden tables, UUID cast width |
| 4 | 4×P1, 1×P2 | All fixed — 3 more payload tables, foreign tables, column-level privileges, unique expression indexes, unknowable encrypted metadata |
| 5 | 2×P1, 1×P2 | All fixed — purge targets under RLS, partial-index predicate columns, non-`public` materialized views |

Round 6 raised three further items. They are **real** — I verified each against the source rather than taking them at face value — but they fall past the budget, so they are filed as autumn-foundation/autumn#2366 instead of being folded in here:

1. `autumn_runtime_config_values` / `_changes` missing from the framework payload list (P1)
2. Materialized-view refresh ordering misses dependencies routed through an ordinary view (P1)
3. Partition-key columns rewritten through the parent (P2)

### Why stopping here is the right call

The findings have converged on one theme: things that sit *outside* the classified universe — framework-owned tables, other schemas, other relkinds, privilege gaps, index shapes. The classification engine and the statement generator have been stable for several rounds; what keeps turning up is another member of that set. That is a bounded, enumerable class better handled as its own change with its own tests than as an ever-growing tail on this PR.

### Verification beyond the reviews

Because the catalog probes were new, I ran them against a real PostgreSQL 16 rather than relying on review reasoning alone:

- every catalog query executed and returned exactly the expected rows (the FK probe finds **both** sides; the `pg_depend` join finds expression- and predicate-index inputs the `indkey` join misses);
- every generated replacement expression executed, including the `(substr(token, 1, 32))::uuid` cast whose 64-character form genuinely fails;
- the pre-fix `CASE WHEN c IS NULL THEN NULL ELSE NULL END` form was confirmed to error on a non-character column, and the FK-referenced-side rewrite was confirmed to abort — i.e. those P1s reproduce;
- the full drill ran through the real binary: seed → `db backup` → `db scrub --artifact` into a separate staging database → **zero occurrences** of all seven seeded PII values swept over every column of every table, uniqueness and foreign keys intact, `NOT NULL` respected, `NULL`s preserved, generated column untouched, purge applied, source database unmodified, and a re-run byte-identical.

---

madmax983:
## CI status: the two blocking gates pass; two failures are in code this PR does not touch

### The gates this PR exists to satisfy are green on `c01ad03`

**`Scaffold Postgres e2e gate` — success on both `stable` and `1.88.0`.** The job log confirms the test genuinely ran rather than being skipped:

```
running 1 test
test integration::db_scrub::scrubbed_database_migrates_clean_and_boots_the_app ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 898 filtered out; finished in 140.91s
```

That single line closes out both earlier failures on this PR, which were mine and real:

| Failure | Cause | Fix |
|---|---|---|
| `column att.attgenerated does not exist` | CI runs `postgres:11-alpine`; that catalog column is PG 12+ | `36b1461` — version-tolerant catalog probes |
| `no encryption key could be resolved for the "dev" profile` | The fixture app never provisioned a key for its `#[encrypted]` column | `c3e1207` — provision via `autumn credentials edit` |

A third defect surfaced in review: `ServerGuard::drain` read the app's pipes with `read_to_string`, which blocks until EOF. A healthy server never closes its pipes, so the **success path would have hung until the 90-minute timeout** — the test had only ever completed by failing early. Killing the child first is not sufficient, since `cargo run` makes the server a grandchild that inherits the handles; `f6e7b21` redirects to files instead. The 140.91s completion above is that fix working.

`Lint` is also green, along with all 18 app-variant conformance jobs, five fuzz targets, MSRV, SQLite, edge capsule, feature combinations, both build-and-boot gates, downstream smoke, SemVer and docs.

### Two failures are outside this PR's diff

**`Test (ubuntu-latest)` — `pg::tests::sanitize_does_not_let_jdbc_ssl_true_override_an_explicit_sslmode`**

- `git diff origin/trunk-dev...HEAD -- autumn-cli/src/pg.rs` is **empty**; the test predates this branch.
- Passes locally in isolation and in the full suite the job runs: **5865 passed, 0 failed**.
- The path is deterministic — every `Err` in `filter_ssl_params` depends only on the input pairs. No env reads, no I/O, no `HashMap`/`HashSet` iteration, no `cfg(feature)` branches in `pg.rs`.
- Not a merge interaction: `trunk-dev` has not moved past this PR's base (`8f94e60`).
- Not feature unification: `cargo test -p autumn-cli -p autumn-web --bins` passes, so unifying features with `autumn-web` does not reproduce it.

**`Test (macos-latest)` — `upgrades_in_place_under_load_without_dropping_a_connection_or_the_state`**

- Lives in `examples/hot-upgrade/tests/live_upgrade.rs`, also untouched by this diff.
- A load test that spawns a process, has it `exec` a successor, and waits on a 30-second `recv_timeout` — timing-sensitive by construction.

Two *different* unrelated tests failing on two different platforms, neither in the diff, is the signature of the base or the runners rather than of this change.

### What I have not done

**Re-run the failed jobs.** Three attempts returned `403 This workflow is already running`, because the Windows job in that run is still going. I am not calling these flakes on that basis alone — "flake" is not a root cause — so this comment records the state rather than asserting a verdict. I will re-run once the workflow completes and report what comes back. If either fails again it is real, and I will treat it as such.

Nothing here is skipped, disabled, or quarantined.

---

madmax983:
## Update on `fe5060a`: the pg failure was transient; a hot-upgrade flake is now blocking verification

Two changes since [my earlier comment](https://github.com/autumn-foundation/autumn/pull/2365#issuecomment-5466374478).

### The `pg::tests::sanitize_…` failure was transient — confirmed

It passed on the ubuntu re-run and again on `fe5060a`. Together with passing locally in isolation, locally in the full suite, and on Windows in CI, that is four passes to one failure in code this PR does not touch. Closed out.

### `upgrades_in_place_under_load_…` is failing across platforms, and it is not this PR's

| Commit | Platform | Result |
|---|---|---|
| `c01ad03` | macOS | failed |
| `fe5060a` | ubuntu | failed |
| local (this branch) | — | **passes in 5.05s** |

The test is `examples/hot-upgrade/tests/live_upgrade.rs:257` — `wait_for_exit(&mut child, Duration::from_secs(30))`, asserting the predecessor drains and exits 0 after handing over. It spawns a process, has it `exec` a successor, and drives load across the handover: timing-sensitive by construction, and it fails under CI runner contention while passing on a quiet machine.

This PR's diff is 17 files and **none of them are under `examples/`** (`git diff origin/trunk-dev...HEAD -- examples/` is empty). There is no fix to port — the test is flaky on the base, not broken by this change.

### Why this matters beyond the flake

It fails at step 6 (`Run tests`), which **skips step 12 (`Run Docker-dependent tests`)** — the step that runs this PR's `db_scrub` testcontainer suite. So the two fixes in `fe5060a` are still unverified in CI, not because they failed but because the job never reaches them.

That step has now been reached exactly once across every run on this PR. Each time something upstream fails first, the actual subject of the change goes untested. I am re-running the failed job to get past it, and will report what the Docker step says — that result is the one I still owe.

Nothing is skipped, disabled, or quarantined, and I am not touching the hot-upgrade test.

---

Issue #2111, opened 2026-07-21
Job timestamps read Utc::now() directly, bypassing the injected Clock seam


madmax983:
### Summary
The job runtime records its timestamps by calling `chrono::Utc::now()` **directly** rather than through the injected `ClockSource` seam (`AppState.clock` / the `Clock` extractor). So a job's *firing semantics* stay virtual under the sim harness (backoff/retry sleeps go through `tokio::time` and are driven by the paused virtual clock), but the *recorded timestamps* are real wall-clock time.

### Where
In `autumn/src/job.rs`, job lifecycle timestamps and the backoff-delay computation read `chrono::Utc::now()` off-seam — `enqueued_at`, `started_at`, `finished_at`, `due_at`, and the delay math (e.g. `let delay = (due - chrono::Utc::now())`). (Grep `job.rs` for `Utc::now()` to enumerate the exact sites/lines before finalizing — cite a couple of representative `file:line`s in the issue.) Contrast: the framework seam is `AppState.clock: Arc<dyn ClockSource>`, read via `state.clock()` / the `Clock` extractor; only ~a dozen framework sites use it today, while ~115 `Utc::now()` + ~110 `Instant::now()` calls are off-seam.

### Why it matters
- **Determinism gap for sim testing (#1797):** under `#[sim_test]`, `Sim::run_to_idle`/`advance` govern the `Clock` extractor and `tokio::time` sleeps, but **not** the job runtime's recorded timestamps — those still read wall time, so a seeded run's job timestamps are not virtual/deterministic even though the ordering/firing is.
- **Product-on-real-clock drift:** application logic that keys off recorded job timestamps (dunning/retry windows, "last attempt at", scheduled-charge due times) inherits real-clock behavior that can't be exercised deterministically or time-traveled in tests.

### Proposed direction
Route the job runtime's time reads through the injected `ClockSource` (the same seam-migration pattern used for high-value ID sites via the seeded `Entropy` seam in the sim-testing waves) so job timestamps honor `with_clock` / the virtual sim clock. This is a later-wave seam-treatment candidate, not a blocker for the current W2 clock-wiring PR — filed to track the gap explicitly.

### Context
Surfaced during the sim-testing #1797 W2 work (virtual-clock wiring + run-to-idle drain). Related to the sim-testing determinism boundary.

PR #2379, opened 2026-08-31
🔌 One-command plugin install: `autumn plugin add` / `autumn plugin list` (#1606)

### LINKED ISSUE #1606: Add `autumn plugin add` for one-command plugin discovery and install
## Problem (the "So What?")

Autumn has a real plugin ecosystem seam — the `Plugin` trait, first-party plugins (`autumn-admin-plugin`, `autumn-cache-redis`, `autumn-storage-s3`), a crates.io naming convention "so the ecosystem is easy to search" (`docs/plugins.md`), and an author-facing conformance gate (`autumn-cli/src/plugin_check.rs`) — but zero consumer-facing tooling. To use a plugin today a framework user must find it themselves, hand-edit `Cargo.toml`, hand-write the `.plugin(AdminPlugin::new()...)` mount in the app builder (`autumn-admin-plugin/README.md`), and hand-wire config — and only then can they run the generators that assume the plugin is mounted (`autumn-cli/src/generate/admin.rs`). Every manual step is a place to pick an incompatible version or misconfigure the mount, and it caps the payoff of the whole plugin architecture: capabilities we already ship stay undiscovered and uninstalled.

## User Story

As a framework user, I want to discover available plugins and install one into my app with a single command, so that adding a shipped capability (admin panel, Redis cache, S3 storage, or a community plugin) takes seconds instead of a docs-and-hand-editing session.

## Acceptance Criteria

- [ ] `autumn plugin list` shows installable plugins with name, one-line description, and the version compatible with the app's `autumn-web` version — covering all first-party plugins in this workspace, plus community crates discoverable via the documented `autumn-plugin-` naming convention.
- [ ] On a fresh `autumn new` app, `autumn plugin add <name>` for each first-party plugin (`autumn-admin-plugin`, `autumn-cache-redis`, `autumn-storage-s3`) adds the dependency at a compatible version, mounts the plugin in the app builder with a sensible default configuration, and prints any required post-install steps (config keys, migrations, follow-up generators like `autumn generate admin`) — and the app compiles on the first try.
- [ ] Version safety: attempting to add a plugin whose supported `autumn-web` range excludes the app's version fails before any file is modified, with a diagnostic naming both versions (composes with the #1601 stability contract).
- [ ] Idempotent: running `autumn plugin add` twice for the same plugin changes nothing the second time (no duplicate dependency, no duplicate mount) and says so.
- [ ] Safe degradation: if the app's builder chain can't be edited confidently (heavily customized `main.rs`), the command makes no code change and instead prints the exact dependency line and mount snippet to apply by hand — it never leaves the app in a non-compiling state.
- [ ] `docs/plugins.md` and each first-party plugin README lead with the one-command install; the manual path remains documented as the fallback.

## Success Metric

Installing `autumn-admin-plugin` into a fresh scaffolded app drops from 4+ manual steps (find crate, edit `Cargo.toml`, edit builder, read config docs) to 1 command, verified by a CI job that runs `autumn plugin add` for every first-party plugin against a fresh scaffold and requires a green `cargo check` — 100% pass rate on every release.

## Out of Scope

- A hosted registry, marketplace, ratings, or curation service — discovery rides crates.io plus the existing naming convention in this slice.
- `autumn plugin remove` / uninstall reconciliation (natural follow-up composing with the existing `autumn destroy` machinery).
- Dynamic/runtime plugin loading or sandboxing — plugins remain compile-time Rust dependencies.
- Any change to the `Plugin` trait or plugin runtime semantics — this is CLI + metadata + docs only.
- Scaffold-time plugin selection flags on `autumn new` (follow-up once the install path exists).

## Gap Analysis

Phoenix's Igniter (`mix igniter.install`) is the modern benchmark: one command patches deps and wires code, and it measurably drove ecosystem adoption. Rails approximates it with `bundle add` + per-engine `rails generate <engine>:install` generators; Laravel's composer package auto-discovery removes the mount step entirely; Django is the cautionary tale — manual `INSTALLED_APPS` editing is its most common beginner failure. No Rust web framework (axum, actix-web, rocket, loco.rs) has any install story: users hand-wire crates, which is exactly autumn's status quo. The autumn-shaped answer is strong because the hard parts already exist: the CLI already generates and reverses code edits (`generate`/`destroy`), the naming convention gives a searchable namespace, `plugin-check` gives a conformance signal worth surfacing in listings, and Rust's compiler makes a botched install loud at `cargo check` rather than at runtime. Implementation work targets `trunk-dev`.

## Complexity Tier

M — almost entirely `autumn-cli` (new `plugin list`/`plugin add` subcommands plus a small per-plugin install-metadata convention), additive with no public API change to `autumn-web` and no migration burden for existing apps.

madmax983:
Closes #1606.

## The problem

Autumn has a real plugin seam — the `Plugin` trait, five first-party plugin crates, a crates.io naming convention, an author-facing `autumn plugin-check` — and no consumer-facing tooling at all. Using a shipped capability meant finding the crate, hand-editing `Cargo.toml`, hand-writing the `.plugin(...)` mount, and reading config docs. Four places to pick an incompatible version or misconfigure the mount.

## What this adds

```bash
autumn plugin list                      # what can I install, at what version?
autumn plugin add autumn-admin-plugin   # dependency + mount + next steps
```

`plugin list` shows every installable plugin with a one-line description and the version compatible with the app's `autumn-web` — the five first-party crates, plus community crates found on crates.io through the documented `autumn-plugin-<name>` convention. `--json` for machine-readable output, `--offline` to skip the lookup.

`plugin add <name>` performs the whole install: the dependency at a compatible version, the mount spliced into the `autumn_web::app()` builder chain, and the post-install steps printed (config keys, follow-up generators like `autumn generate admin`). `--dry-run` shows the edits without applying them.

## Every refusal is total, never partial

- **Version gate runs before a single filesystem action exists**, so installing a plugin whose supported `autumn-web` range excludes the app fails naming both versions with the app byte-identical. It resolves `{ workspace = true }` by walking up, so a workspace member is gated too, and treats range requirements (`>=0.6`) as unknown rather than misreading them as exact versions.
- **Idempotent**: a second `add` reports "already installed" and changes nothing. Mount detection accepts both the qualified path this command writes and the bare call a hand-written mount uses after a `use` import — so `add` never splices a second, default-constructed mount over a user's configured one, and never mistakes a bare import for a mount.
- **Safe degradation**: when the builder chain cannot be edited confidently the command writes *nothing* and prints the exact dependency line and mount snippet to apply by hand, on stderr with exit 2 so a script cannot read it as success. The anchor is scoped to `main`'s own body and refuses on ambiguity, so a builder factored into a helper, a `#[cfg(test)]` harness, or a quick-start snippet inside a raw string all degrade rather than getting mis-spliced.
- **Community crates** get their dependency written but never an automatic mount: the `<Name>Plugin` is derived from the convention and printed, because nothing outside that crate can verify it.

## One mount mechanism, including S3

Four of the five plugins implement `Plugin` and mount with `.plugin(...)`. `autumn-storage-s3` exposes a `BlobStore`, not a `Plugin`, and `S3BlobStore::from_config` is `async` — but `.await` is legal inside a method-call argument in an `async fn`, so its mount is still one splice into the same chain: one anchor, one probe, one degradation path, no statement rewriting.

## Hardening

The crates.io path is a trust boundary — anyone can publish `autumn-plugin-<x>`. Descriptions have control characters neutralised before reaching a terminal (`ESC` is not Unicode whitespace, so collapsing whitespace left ANSI sequences intact); crate names must be legal crate names before being interpolated into a URL or written as a bare TOML key; the client caps redirects and reads bodies through a 4 MiB limit; and an implausible version is refused rather than written into `Cargo.toml`.

## Verification

- `plugin_add_first_party_scaffolds_cargo_check` — the issue's Success Metric — installs **every** first-party plugin into its own fresh `autumn new` scaffold, asserts the second `add` is a no-op, and requires a green `cargo check`. Run locally: green (285s). Wired into `generator-conformance.yml` as a new `plugin-install` job, triggered by the CLI plugin module, the shared source scanner, and every plugin crate.
- 73 unit tests + 9 e2e tests, built red-first.
- `cargo clippy -p autumn-cli --all-targets -- -D warnings` clean; `./scripts/pre-push-check.sh` green.

Reviewed from four angles before opening (correctness, AC compliance, repo conventions/UX, security); the second commit is the fixes.

## Scope

No file under `autumn/` is touched — the `Plugin` trait and runtime semantics are untouched, as the issue requires. No `plugin remove`, no registry, no `autumn new` flags. `autumn plugin-check` is preserved as a distinct command with a regression test. The comment-aware source scanner `generate/pwa.rs` used moves to `autumn-cli/src/rust_source.rs` and is now shared, so the anchor scan and the "already mounted?" check cannot disagree about what a comment is.

https://claude.ai/code/session_014fBXU86FX8zKwqS8XZwKfC

---
_Generated by [Claude Code](https://claude.ai/code/session_014fBXU86FX8zKwqS8XZwKfC)_

### DIFF
--- a/.github/workflows/generator-conformance.yml
+++ b/.github/workflows/generator-conformance.yml
@@ -19,6 +19,8 @@ on:
       - ".github/workflows/generator-conformance.yml"
       - "autumn-cli/src/generate/**"
       - "autumn-cli/src/plugin/**"
+      # The comment-aware scanner `autumn plugin add` anchors its mount on.
+      - "autumn-cli/src/rust_source.rs"
       - "autumn-cli/src/templates/**"
       - "autumn-cli/src/new.rs"
       - "autumn-cli/src/console.rs"
@@ -54,6 +56,8 @@ on:
       - ".github/workflows/generator-conformance.yml"
       - "autumn-cli/src/generate/**"
       - "autumn-cli/src/plugin/**"
+      # The comment-aware scanner `autumn plugin add` anchors its mount on.
+      - "autumn-cli/src/rust_source.rs"
       - "autumn-cli/src/templates/**"
       - "autumn-cli/src/new.rs"
       - "autumn-cli/src/console.rs"
--- a/.github/workflows/generator-conformance.yml
+++ b/.github/workflows/generator-conformance.yml
@@ -18,6 +18,7 @@ on:
     paths:
       - ".github/workflows/generator-conformance.yml"
       - "autumn-cli/src/generate/**"
+      - "autumn-cli/src/plugin/**"
       - "autumn-cli/src/templates/**"
       - "autumn-cli/src/new.rs"
       - "autumn-cli/src/console.rs"
@@ -38,12 +39,21 @@ on:
       - "autumn/src/**"
       - "autumn/Cargo.toml"
       - "autumn-macros/**"
+      # The `autumn plugin add` install gate (#1606) compiles each
+      # first-party plugin into a fresh scaffold, so a change to any of
+      # them can break the mount this workflow is the only proof of.
+      - "autumn-admin-plugin/**"
+      - "autumn-cache-redis/**"
+      - "autumn-media-plugin/**"
+      - "autumn-search/**"
+      - "autumn-storage-s3/**"
       - "Cargo.toml"
   pull_request:
     branches: [trunk, trunk-dev]
     paths:
       - ".github/workflows/generator-conformance.yml"
       - "autumn-cli/src/generate/**"
+      - "autumn-cli/src/plugin/**"
       - "autumn-cli/src/templates/**"
       - "autumn-cli/src/new.rs"
       - "autumn-cli/src/console.rs"
@@ -64,6 +74,14 @@ on:
       - "autumn/src/**"
       - "autumn/Cargo.toml"
       - "autumn-macros/**"
+      # The `autumn plugin add` install gate (#1606) compiles each
+      # first-party plugin into a fresh scaffold, so a change to any of
+      # them can break the mount this workflow is the only proof of.
+      - "autumn-admin-plugin/**"
+      - "autumn-cache-redis/**"
+      - "autumn-media-plugin/**"
+      - "autumn-search/**"
+      - "autumn-storage-s3/**"
       - "Cargo.toml"
   schedule:
     # Weekly on Monday at 04:00 UTC — catches rot from transitive dep updates
@@ -565,3 +583,41 @@ jobs:
           cargo test -p autumn-cli --test cli_tests \
             integration::db_scrub::scrubbed_database_migrates_clean_and_boots_the_app \
             -- --ignored --exact
+
+  # ── `autumn plugin add` install gate (#1606) ────────────────────────────────
+  #
+  # The Success Metric for issue #1606: installing every first-party plugin
+  # into a fresh `autumn new` scaffold must produce a project that compiles on
+  # the first try. `plugin_add_first_party_scaffolds_cargo_check` runs
+  # `autumn plugin add` for each of the five plugin crates against its own
+  # scaffold, asserts the second `add` is a no-op, and then `cargo check`s the
+  # result — the only place the generated mount snippets are compiled against
+  # the real crates.
+  #
+  # It lives in its own job (not the compile-and-serve matrix) because it
+  # builds the aws-sdk-s3 / redis / media dependency graphs, which nothing else
+  # in this workflow pulls in. The five projects share one CARGO_TARGET_DIR
+  # inside the test, so the framework is compiled once.
+  plugin-install:
+    name: autumn plugin add install gate
+    needs: meta
+    runs-on: ${{ fromJSON(needs.meta.outputs.heavy_runs_on) }}
+    timeout-minutes: 90
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Free disk space
+        run: |
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android \
+                      /opt/ghc /opt/hostedtoolcache/CodeQL
+          df -h
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2.7.8
+        continue-on-error: true
+
+      - name: plugin_add_first_party_scaffolds_cargo_check
+        run: |
+          cargo test -p autumn-cli --test generate \
+            plugin_add_first_party_scaffolds_cargo_check -- --ignored --exact
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **One-command plugin install — `autumn plugin add` / `autumn plugin list`
+  (#1606):** Autumn had a real plugin seam (the `Plugin` trait, first-party
+  plugin crates, a crates.io naming convention, an author-facing
+  `autumn plugin-check`) and no consumer-facing tooling at all: using a shipped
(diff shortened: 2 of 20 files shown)


madmax983:
Automated-review rounds 1–3 are addressed in this PR — nine findings, each verified against the code before fixing, each with a regression test:

| Round | Findings | Commit |
|---|---|---|
| 1 | probe hidden in a string literal; dev/build-dependency counted as installed; repeated community `add` claimed a complete install | `4d5cb75` |
| 2 | type name read as a mount; target-gated dependency counted as installed; `async fn main_loop` matched as `main`; `add` could pin a different version than `list` showed | `394aadf` |
| 3 | existing pin at another framework series treated as installed; crates.io plugin against an unpatched local `autumn-web` links a second framework | `d82d187` |

Round 4's three findings are **not** fixed here. This PR's review budget was three rounds, so they are recorded in #2381 with analysis and suggested fixes rather than growing the diff further.

One of them is worth flagging as more than a backlog item: the mount-detection probe has now been reshaped three times (rounds 1, 2 and 4 are the same question from different sides), and each fix trades one false direction for the other. It is a dataflow question answered lexically, so it will keep producing findings until the approach changes. #2381 sets out the two viable directions and the trade-off each accepts. The other two are small and well-specified.

Every review thread on this PR is resolved. CI on `d82d187` is green or running, including the new `autumn plugin add install gate`, which installs all five first-party plugins into fresh `autumn new` scaffolds and requires a green `cargo check`.

---

madmax983:
Round 5 landed three more findings; they are recorded in #2381 alongside round 4's, and its three open threads here are deliberately left unresolved — I have not addressed them.

Round 5 is also the evidence for the concern in my previous comment: all three are narrowing variants of classes already fixed. A module-scoped `async fn main` after round 2 taught the check a name boundary; `br#"…"#` after round 1 taught the scanner raw strings; a `{ workspace = true }` plugin pin after round 3 taught the gate to read a declared version. Six findings across rounds 4–5 reduce to three underlying problems — mount detection, entry-point detection, and dependency resolution — each answering a structural question lexically. #2381 groups them that way and recommends which to fix directly and which need a design decision first.

---

madmax983:
**`Test (macos-latest)` is red on `d82d187`, and it is not this PR's failure.**

Failing test: `integration::sim_job_clock::recorded_job_timestamps_are_virtual_under_the_sim`, in `autumn-web`'s consolidated `integration_tests` binary (1859 passed, 1 failed).

Why it cannot be this change:

- **The diff touches no file under `autumn/`.** `git diff trunk-dev --stat` on this branch covers `autumn-cli/`, docs, READMEs, the workflow, CHANGELOG, CONTRIBUTING and `skills/` — nothing else. The test file `autumn/tests/integration/sim_job_clock.rs` is untouched here and unchanged since #2236.
- **There is no mechanism.** `autumn-cli` depends on `autumn-web`, not the reverse, so nothing in `autumn-cli` can reach that test binary. No `Cargo.toml` and no feature declaration is modified anywhere in the diff, so feature unification across the workspace build is identical too.
- **The same test passed on Linux at this exact SHA.** `Test (ubuntu-latest)`'s "Run tests" step completed green (16:49 → 17:23) on `d82d187`, as did `Sim sweep (sim-testing)` and the SQLite job's two sim DoD suites. The sim is seeded and deterministic (`AUTUMN_SIM_SEED=0x0`), so a genuine logic break would fail on every platform, not one.

The test serialises on the process-global job-runtime lock and calls `job::clear_global_job_client()` — shared-global state whose interleaving is scheduler-dependent, which fits a macOS-runner-only failure.

No fix exists to port: the only related tracking item is #2111, the open feature issue this test was written for, and there is no PR addressing a macOS flake in it. I am not widening this PR to chase it.

Re-running the failed job once, which is the single re-run I will spend here. If it fails again on macOS with the same test I will report that rather than re-running further, since a second identical failure would be evidence about that test, not about this PR.

---

PR #2390, opened 2026-08-31
🗃️ Ledger: batch admin bulk-delete token revoke (statements/req N→1)

madmax983:
## Summary

`AdminModel::execute_action`'s default `"delete"` branch (`autumn-admin-plugin/src/traits.rs`) — the dispatcher behind every scaffolded model's `POST /admin/{slug}/actions` bulk-action endpoint — looped over the submitted ids and called `self.delete(&pool, id)` once per id. For `TokenAdminModel` (the built-in admin model for scoped API tokens), that's one `UPDATE api_tokens SET revoked_at = ... WHERE id = $1 AND revoked_at IS NULL` **per token**, so an operator revoking a batch of tokens (e.g. incident response: revoke every token for a compromised principal) drove one DB round trip per selected row.

- Added `AdminModel::delete_many`, defaulting to the exact per-id loop `execute_action` used to run inline (moved verbatim — every existing implementor keeps byte-identical behavior, abort-on-first-failure included).
- `TokenAdminModel` overrides it with a single batched `WHERE id = ANY($1)` statement, same predicate as `delete()`, returning `ids.len()` to keep the exact same "count of ids asked for" contract (and the same flash message).
- No migration, no new index, no lock beyond each `UPDATE`'s own row locks.

## Profile / Measurement

`pg_stat_statements`, same session, before/after, 20,000-row production-shaped `api_tokens` fixture (skewed principal cardinality, 30% pre-revoked, real dead tuples):

| tier | ids | calls before | calls after | buffers before | buffers after | WAL before | WAL after |
|---|---:|---:|---:|---:|---:|---:|---:|
| small | 100 | 100 | **1** | 526 | 997 | 13,386 | 29,192 |
| medium | 500 | 500 | **1** | 2,703 | 4,267 | 67,928 | 153,615 |
| large | 2,000 | 2,000 | **1** | 10,815 | 16,350 | 272,708 | 634,799 |

**Statement count drops N→1 at every tier** — a 2,000-id bulk action collapses from 2,000 sequential blocking round trips to 1. This clears the impact floor on its own (N+1 elimination).

**Disclosed, not hidden: buffers and WAL are honestly higher for the batched statement** (~1.5–1.9× buffers, ~2.0–2.3× WAL). I checked the obvious structural cause — lost cross-statement HOT-update pruning — via `pg_stat_user_tables.n_tup_hot_upd`/`n_tup_upd` deltas, and it's ruled out: the batched path has equal-or-fewer non-HOT updates (0/0/0 vs 0/6/26). The root cause of the extra buffer/WAL cost isn't further isolated in this pass. Full numbers, EXPLAIN output, and the HOT-update deltas are in the report and the committed `baseline/`/`after/` artifacts.

Full report: `docs/reports/2026-08-30-ledger-token-admin-bulk-delete-batch/README.md`

## Equivalence

`token_bulk_delete_batch_result_equivalence` covers: empty id list, duplicate ids, an already-revoked id (its timestamp is asserted unchanged — not touched twice), a nonexistent id (no error, row count unchanged), all mixed into one call. All 244 existing `autumn-admin-plugin` unit tests and the pre-existing `token_admin_db.rs` Docker tests pass **unchanged**.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo clippy -p autumn-admin-plugin --all-targets --all-features -- -D warnings`
- [x] `cargo test -p autumn-admin-plugin --all-features` (244 unit tests + doctests, all green)
- [x] `cargo test -p autumn-admin-plugin --test token_admin_db -- --ignored token_bulk_delete_batch --nocapture --test-threads=1` against a real `postgres:16-alpine` testcontainer (profile + equivalence tests, both pass)
- [x] Already swept by CI's existing `cargo test -p autumn-admin-plugin --test token_admin_db -- --ignored` — no workflow edit needed
- [ ] `cargo check --workspace` (running; change is isolated to `autumn-admin-plugin`)

---
_Generated by [Claude Code](https://claude.ai/code/session_01TRYCc6T3xrsd48Hw2DEkqp)_


madmax983:
Closing — this duplicates #2384, which merged to `trunk-dev` (commit `ad8357d`) while this PR was open and already fixes the exact same bug: `AdminModel::execute_action`'s per-id "delete" loop on `TokenAdminModel`, batched into a single `WHERE id = ANY($1)` statement, with the same measured trade-off disclosed (statement count N→1, buffers not reduced).

#2384 solves it more narrowly — an `execute_action` override scoped entirely to `TokenAdminModel`, versus this PR's shared `AdminModel::delete_many` trait addition — so there's nothing left here worth carrying forward on top of it. No action needed from anyone else; flagging for the record per the "check for duplicate work" step so it's clear why this is closing rather than being merged through the conflict.

---

PR #2411, opened 2026-09-01
📐 Derive and enforce a proven per-deploy capacity contract (#1733)

### LINKED ISSUE #1733: Derive and enforce a proven per-deploy capacity contract
## Problem (the "So What?")
Autumn protects against overload reactively — bounded-concurrency load shedding (`autumn/src/middleware/load_shed.rs`, `docs/adr/0009-adopt-overload-protection-load-shedding.md`) sheds requests once a live limit is hit — but that limit is a hand-tuned guess held in `autumn/src/runtime_config.rs`, and nothing tells an operator what a given binary can actually sustain before they deploy. Every autumn team still capacity-plans by superstition: pick a box, run ad-hoc load tests, over-provision, hope. There is no artifact that says "this build sustains X req/s at P99 < Y ms on this hardware," and no gate that catches a code change quietly blowing that envelope.

## User Story
As a platform operator shipping an autumn app, I want each deploy to carry a proven capacity contract that the binary self-enforces, so that I can size infrastructure from a guarantee instead of a guess and catch capacity regressions before they page me.

## Vision (R&D)
Over 2-3 years, capacity stops being a spreadsheet and becomes a typed, versioned contract that travels with the binary. Autumn statically reads the handler/route graph (`autumn/src/router.rs`, `autumn/src/route.rs`) to derive each route's resource *shape* — which pools it touches, its bounded allocation profile, the concurrency it can demand — then folds in a one-time calibration run to emit a signed capacity artifact: "on 2 vCPU / 4 GB, this build sustains N req/s at P99 ≤ M ms; beyond that it admission-controls rather than degrades." The runtime enforces that contract (turning `load_shed` from a hand-set limit into a contract-derived one), CI gates it so a change that shrinks headroom fails the build with a diff, and the artifact drives autoscaling and right-sizing directly. Only Rust makes the envelope *derivable and honorable*: no GC pauses, no JIT warmup, visible allocation, and deterministic execution mean worst-case bounds are real rather than statistical — the exact property a GC/dynamic runtime cannot promise. This unlocks a category no framework offers: apps that ship with a provable, self-governing SLO.

- [ ] A new `autumn calibrate` command runs a reproducible, seeded load profile against the built binary and emits a `capacity.lock` artifact recording sustained req/s and P99 latency at the point of saturation, tagged with the build's git provenance and the host profile.
- [ ] The contract records per-route resource shape derived statically from the route graph (pools touched, DB-bound vs compute-bound), not just an aggregate number.
- [ ] A `autumn calibrate --check` mode (CI-runnable) fails with a human-readable diff when a rebuild's contract regresses sustained throughput or P99 beyond a configured tolerance versus the committed `capacity.lock`.
- [ ] The runtime load-shedding admission limit can be sourced from the committed contract instead of a hand-set `runtime_config` value, so the binary admits/sheds against its proven envelope.
- [ ] One guide documents the flow: calibrate → commit contract → CI gate → runtime enforcement, stating the guarantee and its assumptions.

## Success Metric
A representative reference app ships a `capacity.lock` whose predicted sustained req/s is within ±15% of independently measured throughput on the same host class, and a seeded 30%-throughput-regressing change is caught by `--check` in CI with zero false positives across 20 no-op rebuilds.

## Out of Scope
- Multi-node / cluster-wide capacity aggregation and autoscaler integration (single-binary, single-host contract first).
- Automatic remediation or self-provisioning; the first slice enforces admission and gates CI, it does not scale infrastructure.
- Pure symbolic worst-case bounds with no calibration run; the first slice is static-shape-derivation plus one calibration.
- Per-tenant capacity carving.

## Gap Analysis
axum/actix/rocket give raw throughput but no capacity artifact, no admission contract, no CI gate — capacity is left entirely to the operator. loco and leptos add conventions but nothing about proven resource envelopes. Rails/Phoenix/Django/Next.js structurally cannot derive an honorable envelope: GC pauses, JIT warmup, and dynamic dispatch make worst-case latency and memory statistical rather than bounded, so any capacity contract they printed would be a lie under load. The autumn-shaped answer treats capacity like a lockfile: derive resource shape from the already-typed route graph (`autumn/src/router.rs`), calibrate once, commit the contract, enforce it through the load-shedding path that already exists (`autumn/src/middleware/load_shed.rs`, ADR-0009), and gate it in CI — capacity as a verified, versioned build output instead of tribal knowledge.

## Complexity Tier
L — spans a new calibration harness, static route-shape analysis, a versioned contract artifact plus CI diff tooling, and wiring the contract into the existing load-shed admission path.

madmax983:
Closes #1733.

## The problem

Autumn protected against overload *reactively* — bounded-concurrency load shedding (ADR-0009) sheds once a live limit is hit — but that limit was a hand-tuned guess in `autumn.toml`, and nothing told an operator what a given binary could actually sustain before they deployed it. Capacity planning was a spreadsheet and a hope.

This makes capacity a lockfile.

## The flow

```sh
autumn calibrate          # measure → capacity.lock
git add capacity.lock     # the contract travels with the build
autumn calibrate --check  # CI gate: fails with a diff when a rebuild leaves the envelope
```

```toml
[server]
capacity_contract = "capacity.lock"   # the binary admits against its proven envelope
```

`autumn calibrate` builds the app in **release** mode, reads its route graph back through the same `AUTUMN_DUMP_ROUTES` pipeline `autumn routes` uses, boots it on a reserved port **with admission control switched off** — so the run measures the app, not a ceiling it was already carrying — and walks a seeded concurrency ladder, taking the **median of `--runs` repeats** (default 3) at each rung. The **saturation knee** (the last rung where more concurrency still bought materially more throughput) becomes the recorded envelope:

```toml
[envelope]
sustained_rps = 4210.5
p99_latency_ms = 18.42
saturation_concurrency = 64
admission_limit = 64

[[routes]]
method = "GET"
path = "/posts"
shape = "db-bound"
pools = ["db"]
```

## Measured, not assumed

The issue's success metric has two halves, and the second one is the hard one: a regression must be caught, **and an unchanged rebuild must not trip the gate**. Running the real command against `examples/hello` is what settled it — and the first implementation failed:

| | run-to-run spread | no-op rebuilds passing |
|---|---|---|
| one sample per rung | up to 20% | 1 of 3 |
| median of three (`--runs 3`, default) | up to 8% | **5 of 5** |

One sample per rung produced a gate noisier than its own tolerance — precisely the cry-wolf gate this design set out not to ship. The sibling dev-loop and cold-start benchmarks already take repeated runs and a median; calibration now does too.

With that in place, a simulated 30% throughput regression is still caught cleanly:

```
  sustained req/s   8876.0 → 5731.0  (-35.4%, tolerance -20.0%)
✗ this build no longer meets the committed capacity.lock:
    - sustained throughput regressed 35.4% (8876.0 → 5731.0 req/s), beyond the 20.0% tolerance
```

The remaining margin is honest but not generous — ~8% noise against a 20% tolerance — which is why the live gate is scheduled/manual rather than per-PR, and why `docs/guide/capacity-contracts.md` states the assumptions rather than implying a guarantee it cannot make.

## Design decisions worth reviewing

**Failing open is load bearing.** Every failure along the contract path — missing file, malformed TOML, a newer schema version, a contract measured on another host class, a recorded limit of `0` — degrades to *unlimited* with a warning, never to a ceiling. Failing closed would mean a typo'd path or a stale lockfile sheds every request on the way up, turning a capacity feature into an outage. An explicit `server.max_concurrent_requests` always wins, **including an explicit `0`**: an operator who turned shedding off by hand must not have it switched back on by a file committed months ago.

**The gate is deliberately narrow.** `--check` compares two numbers, and only on a matching host class. It never compares timestamps, git provenance, or the route digest, because a no-op rebuild changes all three — and a capacity gate that cries wolf gets switched off, which is strictly worse than not having one. Route-graph drift is reported *beside* the numbers but never fails the gate.

**`--check` replays the committed workload.** Profile, feature selection, seed, ladder, rung duration, repeat count and explicitly named `--target`s all live in the contract's `[calibration]` section and are replayed by the gate. Comparing a rebuild measured under this invocation's defaults against an envelope produced by different ones would make the diff meaningless.

**Per-route shape is a provable subset.** It is read off the handler's declared extractors at macro-expansion time, so a pool reached through an application-held `State` is invisible and the route reads `compute-bound`. That means "no pool proven", never "no pool touched" — the same caveat the security dimensions of `routes audit` carry. Repository auto-API routes are `db-bound` by construction, since their generated handlers own the pool checkout internally.

**Calibration drives plain HTTP over TCP.** `server.unix_socket` is cleared for the calibration child; in-process TLS is deliberately *not* cleared, because the env override materializes `[server.tls]` from variable **presence**, so clearing it that way would switch TLS on — with empty cert paths — for every app that configures none. A TLS-terminating profile is reported by the readiness timeout, which names `--profile` as the way out.

**The live gate does not run on PRs.** Per-PR only the pure logic runs; the release build + ladder lives in a scheduled/manual workflow, matching the cold-start and dev-loop-scaling gates.

## Acceptance criteria

| AC | Evidence |
|---|---|
| `autumn calibrate` emits `capacity.lock` from a reproducible seeded profile, with git provenance + host profile | `autumn-cli/src/capacity_driver.rs`, `autumn/src/capacity.rs`; `the_request_plan_is_reproducible_for_a_seed`, `contract_round_trips_through_toml`; verified end-to-end against `examples/hello` |
| Per-route resource shape derived statically from the route graph | `autumn-macros/src/route.rs` → `ApiDoc::pools` → `route_listing::RouteInfo`; `route_macro_records_*` (5), `a_route_declaring_a_database_pool_is_db_bound` + 3 |
| `--check` fails with a human-readable diff beyond tolerance | `check_contract` / `render_check_report`; `a_thirty_percent_throughput_regression_fails`, `a_tail_latency_blowout_fails_even_when_throughput_holds`, `run_to_run_noise_inside_the_tolerance_passes`, `the_rendered_diff_names_the_metric_and_both_numbers`; plus the empirical 5/5 + −35.4% runs above |
| Runtime sources the load-shed limit from the contract | `router::build_load_shed_layer` → `capacity::resolve_configured_admission_limit`; end-to-end `ceiling_is_sourced_from_the_committed_capacity_contract`, `a_contract_from_another_host_class_does_not_throttle_this_process` |
| One guide documents calibrate → commit → gate → enforce, with the guarantee and its assumptions | `docs/guide/capacity-contracts.md` |

## Testing

Built TDD red→green throughout; each cycle's red was observed before implementing.

- `cargo test -p autumn-web --lib capacity::` — 17 passed
- `cargo test -p autumn-web --lib capacity_contract` (config) — 3 passed
- `cargo test -p autumn-web --lib route_listing::` — 60 passed
- `cargo test -p autumn-macros --lib route::` — 87 passed
- `cargo test -p autumn-web --test integration_tests --features test-support load_shed` — 6 passed (2 new + 4 pre-existing, no regression)
- `cargo test -p autumn-cli --bin autumn capacity` — 22 passed
- `cargo test -p autumn-cli --bin autumn manifest_uploads` — 16 passed
- `cargo test -p autumn-cli --bin autumn parse_calibrate` — 2 passed

The AC-4 e2e is bounded by a 10s timeout: were the ceiling ever *not* sourced from the contract, the request would be admitted and block on the test gate forever, turning a regression into a hung CI job instead of a failing assertion.

## Follow-ups

#2418 tracks two gaps found by a later review round: the runtime does not yet check which **profile** a contract was calibrated under (a `capacity.lock` shared between staging and prod could impose the wrong envelope), and `package_manifest_dir` requires an explicit `-p` in a virtual workspace. Neither is a regression; an explicit `-p` and single-profile deployments are unaffected.

## Out of scope (per the issue)

Multi-node/cluster aggregation and autoscaler integration, automatic remediation, pure symbolic worst-case bounds, and per-tenant capacity carving.

https://claude.ai/code/session_01SXckJzNnQ2pVSYsffzUUy6


madmax983:
The two findings from the review round on `4355d3a` are tracked in #2418 rather than folded in here.

This PR has been through six review rounds; the agreed budget for this change was five (extended from three because every round through five turned up genuine P1s, all of which were fixed here). Both new findings are real — I verified them against the code rather than taking them on trust — but neither is a regression and neither blocks the first slice:

- **Runtime does not check `calibration.profile`** — the contract records which profile it was calibrated under and `--check` replays it, but `admission_limit_for_host` gates on host class alone. A `capacity.lock` shared between staging and prod could impose the wrong envelope. Fails toward *enforcing something*, so it wants the same fail-open treatment every other contract mismatch gets.
- **`package_manifest_dir` needs an explicit `-p`** — in a virtual workspace with one unambiguous binary member, `calibrate` run from the root boots the right binary in the wrong directory. An explicit `-p` (and any single-package repo) is unaffected.

Continuing to iterate here has hit diminishing returns relative to review cost; #2418 is the better place for both.

---

Issue #2398, opened 2026-09-01
In-place upgrade RSTs connections queued in the predecessor's accept backlog (live_upgrade flakes on macOS)


madmax983:
`upgrades_in_place_under_load_without_dropping_a_connection_or_the_state` (`examples/hot-upgrade/tests/live_upgrade.rs:268`) has failed twice on `Test (macos-latest)`, on two different commits of autumn-foundation/autumn#2391:

| Head | Errors | Adopt → listener close |
|---|---|---|
| `ecc89f7` | 1 | `14.749` → `14.754` (5 ms) |
| `cf749f9` | 3 | `32.941` → `32.956` (15 ms) |

Both are the same assertion and the same errno:

```
assertion `left == right` failed: no connection may be refused across the cutover
  left: 3
 right: 0

connection error against 127.0.0.1:49888/: Connection reset by peer (os error 54)
```

Note the errno: `ECONNRESET`, not `ECONNREFUSED`. The connection was *accepted by the kernel* and then reset — it is not a connection that arrived after the socket was gone.

## Mechanism

The predecessor's cutover path skips the readiness flip and the prestop grace (correctly — the successor already holds the same listening socket), then closes its listener immediately:

```rust
// autumn/src/app.rs — the upgrade_cutover branch
tracing::info!(phase = "upgrade_cutover", … "shutdown: successor is serving; draining without a readiness flip");
// …
tracing::info!(phase = "listener_stopping", "shutdown: stopping listener");
// Phase 5: stop listener and signal jobs/scheduler to stop dequeuing.
shutdown_signal_token.cancel();
```

In-flight requests the predecessor has already accepted are drained. What is not drained is the **kernel accept queue**: connections completed by the TCP stack but not yet returned by `accept()`. Closing a listening socket with a non-empty backlog sends `RST` to exactly those. The predecessor's own log confirms it then exits cleanly — `shutdown: all phases completed cleanly, exit_code: 0` — so nothing surfaces the dropped connections at runtime either.

Under the test's 8 concurrent load tasks at 5 ms intervals the backlog is rarely empty at that instant, which is why it fires intermittently rather than never.

The window is the gap between the successor adopting the socket and the predecessor closing its copy. Both are in the table above; the run with the wider gap dropped more connections, which is the shape you would expect if the backlog is the culprit.

## Why this matters beyond the test

[...]

Issue #2392, opened 2026-08-31
`upgrades_in_place_under_load_without_dropping_a_connection_or_the_state` is load-sensitive and flakes CI


madmax983:
`examples/hot-upgrade/tests/live_upgrade.rs`'s live-upgrade test (added in #2345 for #1674) fails intermittently in CI. Surfaced while driving #2377 to green; it is unrelated to that PR and is filed here rather than worked around there.

## Evidence

The test asserts **zero** loss across a live exec-based socket handover while under load. Two of those assertions have fired:

| Where | Line | Assertion |
|---|---|---|
| CI, macos-latest (`1760b5e`) | 268 | `assert_eq!(connect_errors, 0)` — *"no connection may be refused across the cutover"*; got `left: 1, right: 0` |
| Local, under CPU contention | 283 | `assert!(failed.is_empty())` — *"every read must be served across the cutover"* |

Same test, two different zero-tolerance assertions, so this is not one specific race — it is the cutover losing a single request when the machine is busy.

**It is not deterministic per platform.** On PR #2377 it failed on `ubuntu-latest` on one head and then **passed on `ubuntu-latest` on the very next head**, failing instead on `macos-latest`. It follows whichever runner happens to be contended.

**Reproduced locally**, which is the useful part:

```bash
cargo test -p hot-upgrade --test live_upgrade      # passes, ~5s (30s timeout)

# now hold 3 of 4 CPUs busy and run the built binary repeatedly:
for i in 1 2 3; do (bash -c 'while :; do :; done') & done
for run in $(seq 1 8); do target/debug/deps/live_upgrade-* --test-threads=1; done
# → ~1 failure in 8
```

Idle it passes every time. Contended it fails roughly one run in eight. GitHub's shared runners are exactly that environment.

## Why this matters beyond the noise

It is a **cross-cutting** flake: `live_upgrade` runs inside `cargo test --workspace`, so it can redden the `Test` job of any PR in the repo regardless of what that PR touches. #2377's diff is 25 files, none under `examples/hot-upgrade/`, and it was hit twice.

The risk is not the lost minutes — it is that a genuinely broken PR gets waved through as "oh, that's the hot-upgrade flake again".

## Suggested directions

Deliberately not prescribing one; the right call belongs to whoever owns #1674's guarantee:

[...]

Issue #2372, opened 2026-08-30
🐛 SIGUSR2 in-place upgrade resets an accepted connection during drain (live_upgrade flake is a real defect)


madmax983:
## Summary

`examples/hot-upgrade/tests/live_upgrade.rs::upgrades_in_place_under_load_without_dropping_a_connection_or_the_state` fails intermittently in CI. It looks like a timing flake, and I initially treated it as one. It is not: the test is correctly catching a rare violation of the zero-downtime guarantee in the `SIGUSR2` upgrade path.

Filing separately from #2365, whose diff does not touch `examples/` or `autumn/src/app.rs` — it just happens to be the PR that kept tripping over this.

## Occurrences

| Commit | Platform | Result |
|---|---|---|
| `c01ad03` | macOS | failed |
| `fe5060a` (attempt 1) | ubuntu | failed |
| `fe5060a` (attempt 2, re-run) | ubuntu | failed |
| local, quiet machine | linux | passes in ~5s |

Three failures across two platforms. It reproduces only under load, which is why it reads as flaky.

## What actually fails

Not the 30s `wait_for_exit` — the test fails in ~5.3s total, well inside that. The failing assertion is `live_upgrade.rs:268`:

```rust
assert_eq!(connect_errors.load(Ordering::Relaxed), 0,
    "no connection may be refused across the cutover");
```

`left: 1, right: 0`. The load loop logged exactly one error:

```
connection error against 127.0.0.1:35195/: Connection reset by peer (os error 104)
```

## Why this is a real defect, not a client artifact

**ECONNRESET (104), not ECONNREFUSED (111).** The listener was never absent — an *established* connection was torn down.

The usual benign explanation is a pooled keep-alive connection being reused after the server closed it. That is ruled out here: `get()` opens a **fresh `TcpStream` per request** and sends `Connection: close`. There is no pool. So the sequence is:

1. `TcpStream::connect` — **succeeds**
2. `write_all(request)` — **succeeds**
3. `read_to_end` — **fails, ECONNRESET**

A connection was accepted and then dropped without a response.

It is also not a gap in listener coverage. `app.rs:4678` duplicates the listening socket (`HandoffSocket::from_listener`) and hands the dup to the successor while the predecessor keeps serving through the original — both descriptors refer to one socket with a shared accept queue, so there is no window where nothing is listening, and the predecessor's eventual `close()` cannot reset queued connections while the successor holds a dup.

[...]

PR #2412, opened 2026-09-01
🔐 ACME config: close the two doctor↔runtime validation-parity nits (#1874)

### LINKED ISSUE #1874: ACME/TLS config validation parity — residual low-severity doctor↔runtime nits
## Context

Deferred from #1608 / PR #1858 (merged). These are two low-severity validation-parity gaps between the doctor/`validate()` checks and the runtime typed deserialization.

Both are **low-severity**: each is caught fail-fast at boot / first issuance today; the only gap is that `autumn doctor --strict` / `validate()` don't reject them as early as the runtime typed-deserialization does. No security, DoS, or CA-rate-limit impact. Batched here rather than blocking the merge.

## Item 1 — doctor: string-typed `renew_before_days` silently defaults

A string-typed `renew_before_days` (e.g. `renew_before_days = "30"`) is treated as *absent* by the `as_integer()` chain in `resolve_acme_doctor_config` (`autumn-cli/src/doctor.rs`) and silently defaults to 30, while the runtime rejects it as a non-`u32`.

**Fix:** preserve a parse error (as `http_challenge_port` / `directory` already do) so `acme_config` reports **Fail**.

## Item 2 — validate(): whitespace-padded domain passes but is stored untrimmed

A whitespace-padded domain (e.g. `" app.example.com "`) passes `AcmeConfig::validate()` (`autumn/src/config.rs`) because it trims only for the blank/wildcard checks but stores the *untrimmed* value, which then reaches the placeholder/CSR + ACME `Identifier::Dns` order paths.

**Fix:** reject `domain != domain.trim()` (or normalize domains before use).

## Notes

Both are ~1-line fixes with a doctor/validate test each — good first-issue-sized cleanup.

madmax983:
Closes #1874.

Two low-severity parity gaps between `autumn doctor` / `AcmeConfig::validate()` and the runtime's typed deserialization. Built red → green → refactor, then reviewed by three agents from different angles (correctness/parity, test quality, scope/docs) with every confirmed finding fixed.

## Item 1 — doctor: a non-integer `renew_before_days` silently defaulted

`resolve_acme_doctor_config`'s `as_integer()` chain collapsed "present but not an integer" into "absent", so `renew_before_days = "30"` (and a float or a bool) silently became 30 and `acme_config` reported **Pass** — while the runtime's typed `AcmeConfig` refuses to deserialize the same file and the server exits at boot.

Doctor now records the rendered value the operator actually wrote in a new `renew_before_days_error` and reports an `acme_config` **Fail** naming it, exactly as it already did for `http_challenge_port` and `directory`.

One correction to the issue's framing: a *negative* or out-of-`u32`-range integer was **not** a `--strict` false pass. Those did parse, were clamped to `u32::MAX`, and already failed via the `>= 90` rule — for them this change improves only the message (it now describes the malformed value rather than the renewal window). The changelog and field docs say so rather than overclaiming.

## Item 2 — `validate()`: a whitespace-padded domain passed, then was used untrimmed

`AcmeConfig::validate()` trims only for its blank and wildcard checks but **stores** the entry untrimmed — and the untrimmed string is what `CertificateParams::new` turns into the certificate's SAN (`autumn/src/acme/renewal.rs:679`) and what becomes the ACME order's `Identifier::Dns` (`renewal.rs:598`). So `domains = [" app.example.com "]` was requested verbatim and failed mid-issuance with an opaque CA error.

`validate()` now rejects `domain != domain.trim()`, naming the entry, its index, and the trimmed spelling to use. Rejecting rather than silently normalising is deliberate: quietly trimming would issue a certificate for a name the operator did not literally write.

The **doctor grader mirrors the same rule** — the issue is about parity, and leaving doctor lenient would just re-open the gap in the other direction.

Per-entry rule order is blank → wildcard → untrimmed on both sides, so `" *.example.com "` still reports the wildcard (the deeper problem, and the message operators get today) rather than the padding.

## Tests

TDD, with each red observed before its green:

| Phase | Evidence |
|---|---|
| 🔴 | `validate_acme_whitespace_padded_domain_rejected` failed — `validate()` returned `Ok(())` for every padded form |
| 🟢 | Rule added; all 10 `validate_acme_*` pass |
| 🔴 | `acme_config_fail_when_renew_before_days_malformed` and `..._domain_entry_whitespace_padded` failed at assertion level (the parse recorded the bad value, the grader still passed it) |
| 🟢 | Grader branches added; all 41 `doctor::tests::acme_*` pass |

New coverage beyond the issue's "a doctor/validate test each":

- **`acme_config_grader_agrees_with_runtime_validate`** — the doctor grader is a hand-copy of `AcmeConfig::validate()` and the two have now drifted three times (#1608, plus both items here). Every other test pins one side against a hardcoded expectation; nothing pinned the two sides against *each other*. This asserts equal verdicts across 15 configs, and a new `AcmeConfig` field breaks its compile.
- **`acme_numeric_fields_reject_non_integer_toml_values`** — pins the premise the entire doctor fix rests on: that runtime serde really does reject these spellings. Without it, a future lenient `deserialize_with` would make doctor fail a file that boots fine, with nothing to catch it.
- Assertions strengthened where a wrong-but-plausible fix would have passed: the malformed-`renew_before_days` detail must carry the **quoted** `"30"` and the deserialize-tier wording (not the `>= 90` wording), and the padded-domain message must echo the operator's **raw** entry, not only the trimmed spelling it suggests.

## Refactor (no behavior change)

`check_acme_config_impl` takes `&AcmeDoctorConfig` instead of 7 positional args — an 8th would trip `clippy::too_many_arguments` — and its "the runtime cannot even parse this" tier moved into `check_acme_deserialize_errors`, clearing a real `too_many_lines` (102/100). The two near-identical scalar parsers collapse onto a generic `parse_acme_scalar`, and the thrice-duplicated `fail` closure onto one `const acme_config_fail`.

## Known gap, deliberately out of scope

`cache_dir` still uses the same lenient `and_then(as_str)` pattern one binding below the one fixed here, so `cache_dir = 42` passes `doctor --strict` while the runtime refuses to boot (and doctor then inspects an invented directory). Same bug class, and now nearly a one-liner with `parse_acme_scalar` — but it is not in this issue's scope, which was explicitly batched to these two items. Filed as a follow-up rather than widening this PR.

## Verification

- `cargo test -p autumn-web --lib config::tests::{acme,validate_acme}` — 15 pass
- `cargo test -p autumn-cli --bin autumn doctor::tests::acme_` — 41 pass
- `cargo clippy -p autumn-cli --all-targets -- -D warnings` — clean
- `./scripts/pre-push-check.sh` — an earlier run caught `missing_const_for_fn` + `unnecessary_wraps` on the new helper under `-D warnings` (a scoped clippy run without `-D warnings` reports them as warnings, which is how they were missed); both fixed, full gate re-running.
- No example, starter, fixture, doc, or deploy script in the repo uses a padded `domains` entry, so the new startup rejection breaks nothing.

https://claude.ai/code/session_01Fzuq1humVRgrY8xqK23sG5

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fzuq1humVRgrY8xqK23sG5)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -524,6 +524,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **ACME config: `autumn doctor` and the runtime now reject the same two
+  spellings (#1874):** two low-severity parity gaps let `autumn doctor
+  --strict` bless an `autumn.toml` the server refuses to boot on, and let one
+  malformed value through both. A **string-typed `renew_before_days`**
+  (`renew_before_days = "30"`, and likewise a float, a bool, a negative, or an
+  out-of-`u32`-range integer) was read by doctor's `as_integer()` chain as
+  *absent* and silently defaulted to 30, while the runtime's typed
+  deserialization rejects it — doctor now records the value the operator
+  actually wrote and reports an `acme_config` **Fail** naming it, exactly as it
+  already did for `http_challenge_port` and `directory`. A
+  **whitespace-padded domain** (`domains = [" app.example.com "]`) passed
+  `AcmeConfig::validate()`, which trims only for its blank and wildcard checks
+  but stores the entry untrimmed — and the untrimmed string is what becomes the
+  certificate's SAN and the ACME order's `Identifier::Dns`, so the padded name
+  was requested as-is and failed mid-issuance with an opaque CA error. Both
+  `validate()` and doctor now reject it at startup with a message naming the
+  entry, its index, and the trimmed spelling to use. Neither gap was a security,
+  denial-of-service, or CA-rate-limit issue: each was already caught fail-fast
+  at boot or at first issuance, just later and less legibly than it should have
+  been.
+
 - **CI now rejects colliding migration versions:** app, framework and plugin
   migrations are applied into one shared version space — diesel keys
   `__diesel_schema_migrations` on the 14-digit version and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -526,14 +526,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **ACME config: `autumn doctor` and the runtime now reject the same two
   spellings (#1874):** two low-severity parity gaps let `autumn doctor
-  --strict` bless an `autumn.toml` the server refuses to boot on, and let one
-  malformed value through both. A **string-typed `renew_before_days`**
-  (`renew_before_days = "30"`, and likewise a float, a bool, a negative, or an
-  out-of-`u32`-range integer) was read by doctor's `as_integer()` chain as
-  *absent* and silently defaulted to 30, while the runtime's typed
-  deserialization rejects it — doctor now records the value the operator
-  actually wrote and reports an `acme_config` **Fail** naming it, exactly as it
-  already did for `http_challenge_port` and `directory`. A
+  --strict` bless an `autumn.toml` the server refuses to boot on. A
+  **non-integer `renew_before_days`** — a quoted `renew_before_days = "30"`, a
+  float, or a bool — was read by doctor's `as_integer()` chain as *absent* and
+  silently defaulted to 30, so the check passed while the runtime's typed
+  deserialization rejects the file at boot. Doctor now records the value the
+  operator actually wrote and reports an `acme_config` **Fail** naming it,
+  exactly as it already did for `http_challenge_port` and `directory`. (A
+  negative or out-of-`u32`-range integer already failed, having been clamped to
+  `u32::MAX` and caught by the `>= 90` rule; it now fails with a message about
+  the value rather than about the renewal window.) A
   **whitespace-padded domain** (`domains = [" app.example.com "]`) passed
   `AcmeConfig::validate()`, which trims only for its blank and wildcard checks
   but stores the entry untrimmed — and the untrimmed string is what becomes the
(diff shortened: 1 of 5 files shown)


madmax983:
## `Test (ubuntu-latest)` is red on the known `live_upgrade` flake, not on this PR

**Failing check:** [`Test (ubuntu-latest)`](https://github.com/autumn-foundation/autumn/actions/runs/33554063084/job/100017410798) on `7655f14`.

**What failed:** `examples/hot-upgrade/tests/live_upgrade.rs:268` — `upgrades_in_place_under_load_without_dropping_a_connection_or_the_state`:

```
assert_eq!(connect_errors.load(Ordering::Relaxed), 0,
           "no connection may be refused across the cutover")
  left: 1
 right: 0
```

**Why it isn't this PR's.** This is #2392, matching on every detail — same test, same line 268, same assertion, same `left: 1, right: 0`. That issue states the case directly:

> It is a **cross-cutting** flake: `live_upgrade` runs inside `cargo test --workspace`, so it can redden the `Test` job of any PR in the repo regardless of what that PR touches.

and reports it reproducing locally about **1 run in 8** with 3 of 4 CPUs held busy, following whichever runner happens to be contended (it hit `ubuntu-latest` on one head of #2377 and `macos-latest` on the next). #2372 and #2398 track the same cutover behaviour.

There is also no causal path from this diff. The change is confined to ACME config validation — `AcmeConfig::validate()`, `autumn doctor`'s ACME grader, the changelog and two doc tables. `examples/hot-upgrade` configures no TLS or ACME at all, and the assertion that fired counts refused TCP connections during a `SIGUSR2` socket handover. Every other check on this head is green, including `Lint`, `MSRV (1.88.0)`, `SemVer check`, all 20 app-variant conformance matrices, and all 5 fuzz targets.

**No fix ported, because none exists yet.** #2392, #2372 and #2398 are all open with no linked PR. #2392 also deliberately leaves the remedy to whoever owns #1674's guarantee, listing three options with different consequences for what the test promises — and says it was filed rather than worked around in the PR that hit it. Weakening or `#[ignore]`-ing the assertion here would erode exactly the guarantee it exists to defend, and picking between those options is not this PR's call.

Re-running the failed job once, per the one-re-run rule. If it comes back red on the same assertion I'll say so rather than re-running again.

---

PR #2417, opened 2026-09-01
🧷 Pin worker tiers from the CLI and make `[jobs.fleet]` first-class (#1623)

### LINKED ISSUE #1623: Add per-queue worker pools and queue pinning for worker processes
## Problem (the "So What?")
Every Autumn worker loop drains every queue from one shared pool: `jobs.workers` (default 1 — `autumn/src/config.rs:1703-1705`) spawns identical workers on all three backends (`autumn/src/job.rs:3083`, `:5735`, `:7367`) that each claim from the full queue list (`autumn/src/job.rs:6334-6360`). Queue priority ordering (`docs/guide/jobs.md:200-244`) fixes head-of-line *ordering*, not slot contention — a burst of slow bulk jobs (thumbnails, re-indexing) can still occupy every worker slot and starve latency-sensitive jobs like password-reset emails. The jobs guide already promises "per-queue concurrency caps / dedicated worker pools" as a follow-up (`docs/guide/jobs.md:255-256`), and parent #1613 explicitly defers both dedicated pools and queue pinning to this slice.

## User Story
As a scaling operator, I want to reserve dedicated worker capacity for named queues and pin worker processes to queue subsets, so that a flood of slow low-priority jobs can never starve my latency-sensitive queues.

## Acceptance Criteria
- [ ] An operator can declare, in config alone, dedicated worker capacity for a named queue — N slots that no other queue can consume — with no application code changes.
- [ ] An operator can cap a queue's concurrency below the process total, so a bulk queue can never occupy more than its configured share of slots.
- [ ] A worker-role process (#1613) can be pinned to a subset of queues via config/flags; a pinned process never claims jobs from queues outside its subset, on both Postgres and Redis backends.
- [ ] Existing apps that configure nothing new keep today's exact behavior (single shared pool, strict/weighted drain order) with zero config changes.
- [ ] With one queue saturated by slow jobs, a job enqueued to a queue with dedicated capacity is claimed promptly instead of waiting behind the flood — demonstrated by an integration test.
- [ ] If pinning across the declared topology leaves a configured queue with zero worker coverage, startup and `autumn doctor --strict` diagnose it loudly instead of letting jobs silently accumulate.
- [ ] Per-queue queue depth and oldest-waiting-job age are observable via the actuator jobs endpoint (today's gauges are keyed per job type only — `autumn/src/actuator.rs:500-518`).
- [ ] The jobs guide documents pools, caps, and pinning, replacing the "out of scope" note at `docs/guide/jobs.md:255-256`.

## Success Metric
With one queue fully saturated by slow jobs, p95 enqueue-to-start latency for jobs on a queue with dedicated capacity stays within 2× its unloaded baseline (today it is unbounded — starved until the flood drains).

## Out of Scope
- Worker autoscaling or queue-depth-based scaling signals (deferred in #1613; composes with #1378 metrics).
- Per-job-instance dynamic priority at enqueue time (the other follow-up named at `docs/guide/jobs.md:255`).
- Fleet-level placement of pinned worker processes across hosts — #1621 Phase 2 carries per-host role/config; this slice defines the per-process knob it would carry.
- Changes to the shipped per-job-type `concurrency = N` / `concurrency_key` caps — they stay as-is and compose with per-queue caps.
- The split-topology guard for the `local` backend (owned by #1613).

## Gap Analysis
Sidekiq pins dedicated processes to queue lists with optional weights; Solid Queue's YAML declares worker blocks each with their own `queues:` and `threads:` — the closest shape to what Autumn needs; Celery routes tasks to queues and pins workers with `-Q`; Oban configures per-node `queues: [critical: 10, default: 5]` with per-queue concurrency and lets operators zero-out queues on web nodes — #1613's gap analysis already names that knob as missing. The Autumn-shaped answer: the `[jobs] queues` config (`autumn/src/config.rs:1712-1888`) already names every queue, so per-queue capacity and per-process queue pinning should be declared in the same place, config-only and convention-first — same binary, no new process types beyond #1613's roles.

## Complexity Tier
M — job runtime, config, actuator, and doctor surfaces in the `autumn` crate (plus docs); additive public config only, no schema migration, default behavior unchanged.

---
Parent: #1613 (web/worker process roles). Lands on `trunk-dev`.

madmax983:
Closes #1623.

## What was already there

Most of #1623 had already landed on `trunk-dev` under other PRs (#1621 / #1752 / #1756), none of which name #1623 in their subject: per-queue `reserved` / `concurrency` pools, `jobs.pin`, `QueueSlots` / `queue_may_claim` wired into all three backends, the runtime zero-coverage guard, the actuator per-queue gauges, the doctor coverage check, and the jobs-guide section.

So this PR closes the acceptance criteria that were still genuinely open, rather than rebuilding what exists.

## What this adds

**`autumn serve --pin` (AC3 — "via config/**flags**").** Pinning could only be spelled in `autumn.toml` or `AUTUMN_JOBS__PIN`, while the sibling `--role` knob had a flag. `--pin` is repeatable and comma-separated, forwards the existing `AUTUMN_JOBS__PIN` spelling rather than inventing a second one, and is persisted into `serve.mode` so a bare `serve restart` cannot silently relaunch a pinned worker tier as an unpinned one that drains everything.

```bash
autumn serve --role worker --pin critical
autumn serve --role worker --pin bulk --pin default
```

**`[jobs.fleet]` as a validated config section (AC6).** `autumn doctor` can only *prove* a zero-coverage gap once the operator declares every tier's pin there — but `JobConfig` had no `fleet` field, so declaring the topology that makes the check work was an unknown config key: warned about today, and fatal under `server.strict_config_enforce_all`. It is now a first-class section (`tiers`, `manifest`, `declared_queues`), purely declarative — no process acts on it at runtime. A cross-crate test parses one document through both doctor's raw-TOML reader and the typed config so the two spellings cannot drift.

**End-to-end coverage the ACs asked for.** The existing unit tests cover `retain_pinned` as a pure function, which proves the schedule is filtered — not that a running worker honors it.

- **AC3, both durable backends:** a worker pinned to `critical` leaves the waiting `bulk` job untouched, and an unpinned worker against the same store then drains it, so the negative assertion cannot pass merely because the job was undeliverable. The Postgres test additionally asserts the row is still `enqueued`, ruling out "claimed then lost".
- **AC2:** a `concurrency` cap bounds in-flight jobs below the worker count while still draining the whole backlog, with an uncapped control queue proving the pool really is larger than the cap.
- **Success metric:** p95 enqueue-to-start latency on a dedicated-capacity queue stays within 2× its unloaded baseline under flood. The baseline is measured in-run, so the bound calibrates to whatever machine CI provides instead of encoding one box's timings.

**Docs (AC8).** The guide now covers `[jobs.fleet]` and `--pin`, and the "doctor does not fail on coverage" paragraph — stale since #1756 — is corrected.

## Verification

Docker isn't available in this environment, so rather than ship five `#[ignore]`d tests unexecuted, I stood up local Redis and Postgres 16 and ran them for real. All five pass; each was then re-run with its pin / reservation / cap removed and fails on exactly its intended assertion.

That mutation check earned its keep twice. The Postgres fixture originally hand-picked three migrations, omitted `add_pending_unique_key_to_jobs`, and produced a table the claim query errored against on every poll — and `pg_claim_next_job` swallows that error, so the worker looked like it was "correctly not claiming". The test would have passed for entirely the wrong reason. It now applies the crate's own `FRAMEWORK_MIGRATIONS`.

`./scripts/pre-push-check.sh` passes; `cargo test -p autumn-cli` is green (6116 tests); the schema-snapshot guard passes.

## Review findings fixed in the second commit

Four independent reviews (correctness, config compatibility, test quality, docs accuracy) turned up real defects, all fixed:

- **`serve restart` ignored the restart shell's own `AUTUMN_JOBS__PIN`** and let the recorded pin win — then re-recorded it, so re-pointing a tier by exporting a new pin was overridden *permanently*.
- **A present-but-empty `AUTUMN_JOBS__PIN` is a deliberate unpin** (the app keys off presence), which `Vec<String>` couldn't represent — a bare `restart` re-pinned an intentionally-unpinned daemon.
- **`schema_keys.snapshot` needed the four new `jobs.fleet` leaves.** The pre-push gate is compile-only, so this would have failed CI's `cargo test --workspace` with nothing local catching it.
- **`tiers = ["critical", "bulk"]`** — the flat shape `jobs.pin` uses a few lines above it — was silently dropped as "no topology", quietly switching the AC6 hard-fail *off*. Now reported as a failure.
- **The p95 test's detection band was ~50 ms wide**, and a self-calibrating budget could grow past the signal and pass with the reservation removed. Now 3 s flood jobs, a 500 ms floor, discarded warm-up samples, and a guard that refuses to conclude anything from a too-noisy baseline. Re-verified: 3.0 s loaded p95 against a 500 ms budget.
- **The cap test asserted `peak == 2`**, which a 2-worker process also satisfies.
- **My own `[jobs.fleet]` doc example failed `doctor` if copied verbatim.** Fixed in both copies, and the guide's fenced TOML is now extracted and run through the real check so it cannot regress.

## Out of scope

Unchanged, per the issue: worker autoscaling, per-job-instance dynamic priority at enqueue time, fleet-level placement of pinned processes, the per-job-type `concurrency` / `concurrency_key` caps, and the `local`-backend split-topology guard (#1613).

https://claude.ai/code/session_01ErhnrjNhhdwM3nUHda5shK

---
_Generated by [Claude Code](https://claude.ai/code/session_01ErhnrjNhhdwM3nUHda5shK)_


madmax983:
Codex round 4's finding is real and I've filed it as #2419 rather than fixing it here.

Reasoning, so the choice is auditable:

- This PR was given a budget of **3 Codex review rounds** (5 only if a genuine P1 turned up), with anything beyond that to be captured as an issue. Rounds 1–3 produced six findings, all P2, all fixed. This is round 4, and it is P2 — no P1 has appeared at any point.
- The headline case Codex names — `queues = ["critical", 1]` in a jobs manifest — is **pre-existing** and unchanged by this PR. Only the all-non-string `queues = [1]` sub-case shifted behavior here (previously fell through to `declared_queues`, now returns an authoritative empty set); both the old and new behavior fail to report the corruption, so neither is right, and this PR is not what made it wrong.
- The sound fix is not a fourth one-line guard. `jobs::validate_manifest` already encodes the correct rule for the *emit* path; doctor's *read* path never calls it. The duplication is what let the two drift, so #2419 proposes a shared strict parser instead of a second copy of the check.

Worth recording that this is the fourth instance of one class: a malformed value degrading to `None`/a shorter list, which the caller cannot tell apart from "nothing declared" — and "nothing declared" is exactly the input that switches the zero-coverage hard-fail off. The three instances inside `[jobs.fleet] tiers` are fixed in this PR; #2419 covers the manifest reader and argues for closing the class rather than the instances.

---

PR #2421, opened 2026-09-02
🪞 Echo: single AppState::test_default() for test fixtures (instances 17→1, mass -530 lines)

madmax983:
## Summary

🎯 **Clone class** — 17 instances of the ~35-field `AppState` struct literal used to build a default test fixture, across `autumn/src/app.rs` (7 sites, including its own `test_router`/`test_router_with_config` helpers), `autumn/src/session.rs` (2 sites, including a local `test_state()` helper), `autumn/src/router.rs` (1 site, a local `test_state()` helper), `autumn/src/auth.rs` (1 site, a local `test_app_state()` helper), and `autumn/src/prelude.rs` (2 cfg-split sites in one test). Token mass ≈300-450 tokens / 35-40 lines per instance. Detector: `jscpd` (`cpd 5.1.1`), `--pattern "**/*.rs" --min-tokens 50 --min-lines 10`, config in the PR description below (not committed, per the repo's "one clone class per PR" — see Reproduce). `autumn/src/test.rs`'s similarly-shaped `TestApp::build()` was investigated and excluded: it derives its values from `self.config`/injected clock/entropy rather than hardcoding them, and carries comments explaining deliberate divergence (e.g. why `job_registry` is built from the resolved clock) — textually similar, semantically distinct.

📜 **History** — Three separate historical commits each added one `AppState` field and, in the same commit, hand-edited every one of these files to add that field to their local copy of the literal:
- `df9e7ebd` "Add injectable Clock extractor for deterministic time testing (#1014)" — added `clock`; touched `app.rs`, `auth.rs`, `prelude.rs`, `router.rs`, `session.rs`, `test.rs`, `state.rs`.
- `23563534` "sim: seeded Entropy seam + deterministic IDs (W3, #1797) (#2107)" — added `entropy`; touched the same file set.
- `8e1cb386` "Jobs/roles: expose ProcessRole on AppState (#1726) ... (#1746)" — added `role`; touched the same file set again.

Three occasions of the same files moving in lockstep for the same reason is the co-change signature of one duplicated decision, not five people independently writing similar-looking setup code. No missed-fix defect was found (every site was in fact updated correctly each time), but the risk was real: `test_router_with_config` silently ignored the config it was handed for `forbidden_response`/`auth_session_key` (it always used the same hardcoded defaults as `test_router`), which no current caller happens to exercise, but would have been a source of confusing test failures for the first one that did. That inconsistency now can't recur — there's one fixture, not two independently-maintained ones.

💡 **Hypothesis** — All 17 sites encode the same decision: "here is a default `AppState` suitable for a unit test" (no DB, no real background workers, `info` logging, `ready_for_test()` probes, `user_id` session key, etc.). The three co-change commits are the mechanism: each is one team member's PR threading a new cross-cutting field through what should have been *one* place.

🔧 **Change** — Added `AppState::test_default()` (`#[cfg(test)]`, in `state.rs`, next to the struct's other test-only code) holding the canonical values. Every call site now either calls it directly, or uses Rust struct-update syntax (`AppState { field: ..., ..AppState::test_default() }`) to override only the field(s) that genuinely differ per caller: `health_detailed` (bool, varies true/false across sites), `profile` (`router.rs`'s helper tags itself `"test"`), and `auth_session_key` (`auth.rs`'s helper takes it as a parameter, now passed straight through). None of these are mode/caller-identity flags — they're the same kind of per-test data the original literals already varied field-by-field. Two app.rs inline test bodies (`build_router_post_route`'s state+config+build_router, and `build_router_mounts_health_check_at_custom_path`) were byte-for-byte what `test_router`/`test_router_with_config` already do, so those now just call the existing helper instead of repeating it. `prelude.rs`'s `#[cfg(feature = "db")]`/`#[cfg(not(feature = "db"))]` split of the same literal collapsed to one unconditional call, since `test_default()`'s own field-level `#[cfg(feature = "db")]` attributes already handle both configurations.

📊 **Measurement**
- Instances: 17 → 1 (+16 thin call sites)
- Diff: `+67 / -532` across 6 files (`git diff --stat`)
- jscpd duplicated lines (whole-repo, same config): 23738 → 23242 (-496 lines); clone pairs: 1687 → 1671
- Concept count: zero new caller-identity branches, zero mode/boolean parameters added; every call site is a call or a 1-3-field struct-update, not a wrapper shim
- API surface: no change (all touched code is `#[cfg(test)]`/test-module only; no public API affected)

🔬 **Reproduce**
```
npx --yes jscpd --pattern "**/*.rs" \
  --ignore "**/target/**,**/tests/**,fuzz/**,examples/**,example-e2e/**,benchmarks/**" \
  --min-tokens 50 --min-lines 10 --reporters console --silent .

git show --stat df9e7ebd -- autumn/src/app.rs autumn/src/auth.rs autumn/src/prelude.rs autumn/src/router.rs autumn/src/session.rs autumn/src/test.rs autumn/src/state.rs
git show --stat 23563534 -- autumn/src/app.rs autumn/src/auth.rs autumn/src/prelude.rs autumn/src/router.rs autumn/src/session.rs autumn/src/test.rs autumn/src/state.rs
git show --stat 8e1cb386 -- autumn/src/app.rs autumn/src/auth.rs autumn/src/prelude.rs autumn/src/router.rs autumn/src/session.rs autumn/src/test.rs autumn/src/state.rs
```

## Verification

- `cargo check -p autumn-web --lib --tests --features "test-support,offline-sync,ws,mail,redis,i18n,presence"` — clean
- `cargo clippy -p autumn-web --lib --tests --features "test-support,offline-sync,ws,mail,redis,i18n,presence" --all-targets -- -D warnings` — clean
- `cargo fmt --all` — clean
- `cargo test -p autumn-web --lib --features "ws,presence,test-support" -- app::tests session::tests router::tests auth::tests prelude::tests state::tests` — 285 passed, 0 failed (all pre-existing test bodies unchanged; only their fixture construction moved)

Behavior is unchanged everywhere: every call site still gets the exact same field values it had before this PR, characterized by the pre-existing (unmodified) test assertions passing unchanged.

## Note on size

This touches 6 files, `+67/-532`. Per this task's own size guideline ("any refactor over ~200 changed lines: split it or escalate it"), that's worth flagging explicitly rather than passing silently — I judged it safe to keep as one PR because the diff is a single, mechanical clone class (not several), overwhelmingly deletions, has strong co-change evidence, and every edit is either a plain call or a struct-update with no new abstraction surface. Happy to split per-file if a reviewer would prefer that.

https://claude.ai/code/session_01EmvdK1e2gNT7yF9vfpEJrs

---
_Generated by [Claude Code](https://claude.ai/code/session_01EmvdK1e2gNT7yF9vfpEJrs)_

### DIFF
--- a/autumn/src/app.rs
+++ b/autumn/src/app.rs
@@ -12012,42 +12012,8 @@ mod tests {
     pub fn test_router(routes: Vec<Route>) -> axum::Router {
         let config = AutumnConfig::default();
         let state = AppState {
-            extensions: std::sync::Arc::new(std::sync::RwLock::new(
-                std::collections::HashMap::new(),
-            )),
-            #[cfg(feature = "db")]
-            pool: None,
-            #[cfg(feature = "db")]
-            replica_pool: None,
-            #[cfg(feature = "db")]
-            shards: None,
-            #[cfg(all(feature = "db", feature = "reporting"))]
-            db_capture_gap: None,
-            profile: None,
-            role: crate::config::ProcessRole::Combined,
-            started_at: crate::time::monotonic_now(),
             health_detailed: true,
-            probes: crate::probe::ProbeState::ready_for_test(),
-            metrics: crate::middleware::MetricsCollector::new(),
-            log_levels: crate::actuator::LogLevels::new("info"),
-            task_registry: crate::actuator::TaskRegistry::new(),
-            job_registry: crate::actuator::JobRegistry::new(),
-            config_props: crate::actuator::ConfigProperties::default(),
-            metrics_source_registry: crate::actuator::MetricsSourceRegistry::new(),
-            health_indicator_registry: crate::actuator::HealthIndicatorRegistry::new(),
-            #[cfg(feature = "ws")]
-            channels: crate::channels::Channels::new(32),
-            #[cfg(feature = "presence")]
-            presence: crate::presence::Presence::new(crate::channels::Channels::new(32)),
-            #[cfg(feature = "ws")]
-            shutdown: tokio_util::sync::CancellationToken::new(),
-            policy_registry: crate::authorization::PolicyRegistry::default(),
-            forbidden_response: crate::authorization::ForbiddenResponse::default(),
-            auth_session_key: "user_id".into(),
-            shared_cache: None,
-            clock: std::sync::Arc::new(crate::time::SystemClock),
-            entropy: std::sync::Arc::new(crate::entropy::OsEntropy),
-            app_id: AppState::next_app_id(),
+            ..AppState::test_default()
         };
         crate::router::build_router(routes, &config, state)
     }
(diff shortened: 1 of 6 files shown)


madmax983:
`Test (macos-latest)` failed on `6c878ec` in `tests/live_upgrade.rs::upgrades_in_place_under_load_without_dropping_a_connection_or_the_state` (`assert_eq!` on dropped-connection count: `left: 4, right: 0`, at `tests/live_upgrade.rs:268`).

This isn't this PR's failure: the diff here only touches `#[cfg(test)]` `AppState` fixture construction in `app.rs`/`session.rs`/`router.rs`/`auth.rs`/`prelude.rs`/`state.rs` — nothing in the live-upgrade/connection-draining path `live_upgrade.rs` exercises. The same job passed on `trunk-dev` at this PR's exact base commit (`515f8153`, [run 33545485422](https://github.com/autumn-foundation/autumn/actions/runs/33545485422/job/99989193982)), and this is a load/timing test (concurrent request traffic against a real server during an in-place binary swap) — a classic flake shape on a shared macOS runner, not a root cause in this diff.

I attempted a re-run of just the failed job, but GitHub refused it while the rest of this workflow run is still in progress ("This workflow is already running"). I'll re-run it once the full run completes.

---

madmax983:
`Test (ubuntu-latest)` also failed on `6c878ec`, in the Docker-dependent Postgres suite: `integration::job_tracking_stores_integration::postgres_backend_persists_tracked_job_and_expires_it` panicked with "record should be past its configured TTL" (`autumn/tests/integration/job_tracking_stores_integration.rs:260`) — 301 passed, 1 failed.

Same verdict as the macOS failure above: this diff doesn't touch job tracking, TTL handling, or anything Postgres-related — only `#[cfg(test)]` `AppState` fixture construction. The assertion is a real-clock TTL-expiry check against a testcontainer-backed Postgres row, which is exactly the shape of a CI-timing flake (not enough wall-clock time elapsed under runner contention before the TTL check ran). The same `Test (ubuntu-latest)` job, including its Docker-dependent-tests step, passed on `trunk-dev` at this PR's exact base commit ([run 33545485422](https://github.com/autumn-foundation/autumn/actions/runs/33545485422/job/99989194009)).

The overall workflow run is still in progress, so `rerun_failed_jobs` is still refused ("workflow already running"); I'll re-run both failed jobs together once it finishes.

---

PR #2453, opened 2026-09-02
🔁 Continuous SQLite replication with point-in-time restore (#1628)

### LINKED ISSUE #1628: Add continuous SQLite replication with point-in-time restore
## Problem (the "So What?")
#1614 makes SQLite a first-class production backend for the zero-ops tier, but its durability story is #1595's snapshot backups — its Out of Scope explicitly defers "Continuous streaming replication (Litestream-style)", and #1595/#1619 likewise defer point-in-time recovery. A solo self-hoster on one VPS can therefore lose everything since the last snapshot (hours of writes) when the disk dies. Today the codebase has no WAL/checkpoint handling anywhere and no backup code (`autumn-cli/src/db.rs` ships only `create|drop|reset`), while the pieces a replication story needs already exist: an S3-compatible destination surface with env-indirected credentials (`autumn-storage-s3/src/lib.rs`, `autumn/src/storage/config.rs`), cron scheduling with single-runner coordination (`autumn/src/task.rs`, `autumn/src/scheduler.rs`), and readiness/alerting seams (`autumn/src/probe.rs`, #1610). #1619's offsite upload narrows the loss window only to the backup schedule; this spec closes it to seconds.

## User Story
As a solo self-hoster running my app on the zero-ops SQLite tier, I want the framework to continuously replicate my database to offsite storage and restore it to a point in time with one command, so that a dead VPS costs me seconds of data instead of hours, and recovery doesn't require being a database expert.

## Acceptance Criteria
Phase 1 — continuous replication (first slice):
- [ ] With replication enabled in config, a running app continuously ships SQLite changes to a single offsite S3-compatible destination, reusing #1619's destination conventions (own config section, profile overlays, `AUTUMN_*` env overrides, env-var-indirected credentials, distinct-from-blob-storage default) — no sidecar process, no external tools.
- [ ] The steady-state recovery point objective is configurable, defaults to at most 10 seconds of potential data loss under normal write load, and the effective replication lag is observable by the operator.
- [ ] Replication is safe against a live app: it never corrupts, blocks, or measurably degrades the single-writer workload (#1614's single-host contract).

Phase 2 — restore and PITR:
- [ ] From a fresh machine with only the destination credentials, one command restores the latest replicated state; an optional timestamp flag restores to a chosen point in time within the configured retention window.
- [ ] Restore is gated by the same confirmation/`--force` and integrity-refusal protocol as #1595, and refuses to restore a replica that fails verification.

Phase 3 — trust:
- [ ] Replicas are periodically verified restorable (not merely uploaded); a verification failure or replication lag beyond a configurable threshold is an alertable condition on #1610's channels and never fails silently.
- [ ] CI proves the loop end to end: seed → run app with replication → destroy the local database → point-in-time restore on a clean environment → row equality at the chosen point.
- [ ] The "SQLite in production" guide (#1614) gains a durability section stating the RPO contract, retention window, and the fresh-box recovery runbook.

## Success Metric
An operator whose VPS disk is destroyed recovers on a fresh box, to within 10 seconds of the last committed write, in under 10 minutes with a single restore command — versus a worst case of one full backup interval (hours) with #1595/#1619 snapshots alone.

## Out of Scope
- Read replicas, multi-writer clustering, or networked SQLite (LiteFS, rqlite) — #1614 keeps this tier single-host, single-writer.
- Replacing #1595/#1619 snapshot backups — snapshots remain the coarse-grained, cross-backend story; replication composes with them.
- Postgres continuous archiving/PITR — Postgres deployments already have mature external ecosystems.
- Multi-destination fan-out or cross-region replication policy (same boundary as #1619).
- Client-side encryption of replicated data (follows #1619's posture: TLS in transit, provider server-side encryption at rest, client-side as a named follow-up).
- Admin UI for replication management — config, CLI, and health surfaces only in this slice.

## Gap Analysis
Litestream defined the category: a sidecar streaming SQLite WAL segments to S3 with snapshot+WAL generations and timestamp-addressed restore — but it is a separate binary to install, supervise, and monitor, which contradicts autumn's single-binary pitch. LiteFS and rqlite solve multi-node topologies #1614 explicitly avoids. Rails 8's "SQLite in production" push shipped Solid Queue/Cache but leaves continuous durability to external tools; Turso/libSQL offer it as a hosted service. Postgres operators reach for WAL-G/pgBackRest. The autumn-shaped answer is Litestream's replication model built into the process the operator already runs: enabled by config, targeting the destination surface #1619 defines, surfacing lag and verification through the existing health/alerting seams (#1610), and restoring through the same CLI contract as #1595 — zero extra processes, zero new credential conventions.

## Complexity Tier
L — continuous replication touches the SQLite write path, storage destinations, restore CLI, verification, and alerting, and it is correctness-critical durability machinery.

madmax983:
Closes #1628.

A running app on the zero-ops SQLite tier now ships its write-ahead log to
offsite storage continuously, from inside the process it already runs — no
sidecar to install and supervise, no external tools — and rebuilds the database
on a fresh machine with one command, optionally at a chosen instant.

```toml
[replication]
enabled = true

[replication.s3]
bucket = "myapp-replicas"
region = "auto"
endpoint = "https://<account-id>.r2.cloudflarestorage.com"
access_key_id_env = "AUTUMN_REPLICA_ACCESS_KEY_ID"
secret_access_key_env = "AUTUMN_REPLICA_SECRET_ACCESS_KEY"
force_path_style = true
```

```bash
autumn db replica status --json                 # how fresh is the replica?
autumn db replica restore --force               # fresh box, latest state
autumn db replica restore --timestamp 2026-09-02T14:29:00Z --force --overwrite
```

## The invariant it rests on

In WAL journal mode SQLite writes the main database file **only during a
checkpoint**. So if exactly one component is allowed to checkpoint, that
component can also take a byte-faithful copy of the main file at any other
moment and know it is a consistent base for the WAL frames that follow.

That component is the replicator. When `[replication]` is enabled the SQLite
pool sets `PRAGMA wal_autocheckpoint = 0`, and the replicator holds one
connection open for the process's lifetime — which also stops SQLite's
"last connection closing" checkpoint from firing behind its back.

Two orderings carry the durability story:

- a segment always ends on a **commit boundary**, so a replica is never half a
  transaction;
- a checkpoint is attempted **only once nothing is un-shipped**, and is bracketed
  by `PRAGMA data_version` so a write that slips into the gap retires the
  generation rather than being silently folded away. A dead destination
  therefore stalls checkpointing and grows the `-wal`: it costs disk, never data.

Steady-state upload is the size of your *writes*, not of your database. A
checkpoint opens the next WAL **index** inside the current generation; a full
base snapshot happens once per `snapshot_interval_secs` (hourly by default),
which is also what bounds how much WAL a restore replays.

## Restore refuses rather than best-efforts

Handing SQLite a damaged WAL is the dangerous failure: recovery stops at the
first bad frame and *reports success*, so a truncated replica looks like a clean
restore missing only the last few minutes. Every check here is a refusal — a
hole in the segment sequence, a payload whose SHA-256 does not match, a segment
that does not continue the previous one or carries a foreign salt, a generation
without its commit marker, a replay that stopped short of the page count its
last commit recorded, or a rebuilt database that fails `PRAGMA integrity_check`.
Nothing is published until all of them pass.

## Acceptance criteria

| AC | Evidence |
| --- | --- |
| **1** — config-driven continuous shipping to one S3-compatible destination, #1619 conventions, no sidecar | `[replication]` in `autumn/src/config.rs` + `AUTUMN_REPLICATION__*`; `config::tests::replication_parses_from_toml`, `env_override_replication_fields`, `replication_env_does_not_materialize_a_section_from_optional_keys_alone`, `a_path_destination_from_env_never_grows_an_empty_s3_section`, `replication_env_overlays_a_toml_section`; `replication::tests::a_shared_blob_storage_bucket_needs_an_opt_in`, `endpoint_spellings_of_one_destination_compare_equal`; in-process loop on a dedicated thread wired in `app.rs`, proved by `the_loop_ships_on_its_own_and_flushes_what_is_left_on_shutdown`; real S3 by `sqlite_replication_s3::replicates_to_and_restores_from_a_real_s3_endpoint` (MinIO) |
| **2** — RPO configurable, ≤10 s default, lag observable | `rpo_secs` default 10 (`replication_defaults_to_absent_and_carries_the_documented_rpo`), ship interval = rpo/2 (`sync_interval_defaults_to_half_the_rpo_with_a_one_second_floor`); `lag_and_generation_are_observable_to_the_operator`, the `status` unit suite, `autumn db replica status --json` (`a_fresh_box_restores_the_latest_state_with_one_command`) |
| **3** — safe against a live app | `a_writer_on_another_thread_is_not_blocked_by_a_slow_upload` (40 writes on another thread against a 120 ms/upload destination finish in < 3 s), `replication_never_blocks_or_corrupts_a_live_writer`, `a_dead_destination_stalls_the_checkpoint_rather_than_dropping_data`, `db::tests::replication_adds_the_auto_checkpoint_lock_to_the_pooled_pragmas`, `sqlite::tests::a_second_open_connection_does_not_break_the_checkpoint_report` |
| **4** — one-command fresh-box restore + `--timestamp` | `a_fresh_box_restores_the_latest_state_with_one_command`, `a_timestamp_flag_restores_to_a_chosen_point_in_time`, `an_unconfigured_destination_says_how_to_configure_one` (all drive the real binary in a directory holding only `autumn.toml`); `restores_to_a_chosen_point_in_time_within_the_window`, `a_point_in_time_restore_picks_the_right_generation_out_of_several`, `a_restore_plan_reports_the_instant_it_actually_lands_on`, `a_point_in_time_before_the_window_is_refused_rather_than_rounded` |
| **5** — #1595 guard + integrity refusal | `a_production_profile_restore_is_refused_without_force`, `overwriting_an_existing_database_needs_force` (and `--force` does **not** imply `--overwrite`); `a_missing_segment_is_refused_instead_of_silently_truncating`, `a_tampered_segment_is_refused_by_its_digest`, `a_generation_without_its_commit_marker_is_ignored`, `a_snapshot_corrupted_together_with_its_digest_is_still_refused`, `a_target_outside_the_retention_window_is_refused_not_rounded`, `segment::tests::decode_refuses_*` |
| **6** — periodic verification, alertable, never silent | Verification **is a restore**: `verification_actually_restores_and_fails_loudly_on_a_damaged_replica`, `a_periodic_verification_records_its_outcome_in_the_operator_status`, `a_failed_verification_reaches_the_status_the_health_indicator_reads`, `the_replication_indicator_registers_and_reports_down_through_the_registry` (registered under `sqlite-replication`, `HealthOnly`, DOWN and recovering); escalation is #1610's existing health-indicator alert path, which sweeps that registry |
| **7** — CI proves the loop end to end | `ships_a_generation_and_restores_the_latest_state_on_a_fresh_path` and `restores_to_a_chosen_point_in_time_within_the_window` (in-process) plus the two CLI fresh-box tests — all default-feature and non-`#[ignore]`d, so they run in CI's `cargo test --workspace` with no workflow edit; the MinIO S3 leg is swept by the existing Docker step |
| **8** — guide durability section | `docs/guide/sqlite-in-production.md` § *Durability: continuous replication and point-in-time restore* — RPO contract table, retention window, fresh-box runbook, alerting, what makes it safe against a live app, limits; the three stale "streaming replication is out of scope" statements elsewhere on the page are corrected |

## Review

Four independent review passes (durability/correctness, security and
credentials, API and conventions, test coverage against the ACs) ran against the
first cut; the second commit closes their findings. The ones worth naming:

- the checkpoint interlock compared a `wal_len` measured *before* the tick's
  upload — a transaction committed during that upload could be checkpointed away
  unshipped;
- a generation published its commit marker while its base snapshot was still
  behind the WAL, so a restore could regress past what the previous generation
  had already shipped;
- `snapshot_interval_secs` and `retention_hours` were unreachable on a database
  that never filled the WAL;
- `autumn db replica verify` staged a full copy of production data in a
  world-readable `/tmp` directory;
- unbounded gzip and response-body reads, reachable from the in-process verifier.

## Known follow-ups (deliberately not in this slice)

- No test boots a full `AppBuilder::run` with `[replication] enabled = true`:
  a SQLite runtime pool only exists in the `--features sqlite` lane, so an
  app-level replication test belongs there. The loop, the thread wiring and the
  indicator registration are each covered separately.
- No multipart upload, so a base snapshot above S3's 5 GiB single-`PutObject`
  limit is refused with an actionable message rather than split.
- The signing key is not zeroized (would add a `zeroize` dependency), and
  `ReplicationS3Config` is a third copy of the S3 connection shape that
  `[backup.offsite]` and `[storage.s3]` also carry.
- `autumn doctor` has no dedicated replication check; `AutumnConfig::validate`
  now runs the section's validation, so `autumn check` and boot both catch it.

## Breaking change

`AutumnConfig` gains a public `replication` field, so a struct-literal
construction needs `..AutumnConfig::default()`. Recorded in
`docs/migrations/next.md`.

## Verification

`cargo fmt --all --check`, `./scripts/check-panic-gate.sh` (61 gated modules),
`./scripts/check-determinism-gate.sh`, `cargo clippy --workspace --all-targets
-- -D warnings`, and the full replication suite: 72 unit tests, 27 `autumn-web`
integration tests, 8 `autumn-cli` fresh-box tests. The MinIO test is
`#[ignore]`d and runs in CI's Docker sweep.

https://claude.ai/code/session_0111Vyu2rs9xjiPGWmgwNiiY

---
_Generated by [Claude Code](https://claude.ai/code/session_0111Vyu2rs9xjiPGWmgwNiiY)_


madmax983:
Three checks failed on `26f5e9b`. Two were this branch's and are fixed in `828b026`; one is not this branch's.

**Fixed — `Test (windows-latest)`.** All seven `replica_restore` tests failed parsing machine B's `autumn.toml` with `too few unicode value digits`. The harness interpolated the database path into a *basic* TOML string, where the backslashes of `C:\Users\...` are escape sequences and `\U` opens a unicode escape. Now a literal string, which interpolates a path verbatim. The regression test builds the config from a Windows-shaped path and parses it, so it reproduces on every platform — backslashes are ordinary filename characters off Windows — rather than only in the lane that happened to catch it.

**Fixed — `Test (ubuntu-latest)`.** `sqlite_replication_s3::replicates_to_and_restores_from_a_real_s3_endpoint` asserted the opening tick ships one segment; it ships none. That tick checkpoints before it snapshots, so everything written so far is already inside the database file it just copied. The local suite has pinned exactly that in its `prime` helper since the checkpoint-first fix landed; this test is `#[ignore]`d, so it only runs in the Docker sweep and never caught up. A stale expectation, not a behaviour change — verified by running the test against a real MinIO container locally, where it now passes.

**Not this branch's — `Test (macos-latest)`.** `upgrades_in_place_under_load_without_dropping_a_connection_or_the_state`, in `examples/hot-upgrade/tests/live_upgrade.rs`. This PR's diff is 34 files and touches none of them. The test arrived in fec5221 (#2345), which is already in this branch's merge base, and that is the only commit touching `examples/hot-upgrade` on `trunk-dev` — so there is no fix upstream to port. It is timing-marginal and fails intermittently on macOS.

I have not spent a re-run on it: the push of `828b026` starts a fresh run that re-runs macOS anyway, which is strictly better than re-running the superseded head. If it fails again there I will say so rather than repeat this comment.

---

madmax983:
CI on `a7d35ad`: 12 of 14 jobs green, including `Test (ubuntu-latest)` with the Docker sweep. The two red jobs are both pre-existing flakes in test infrastructure this PR does not touch, and I'm spending the one allowed re-run on them.

**`Test (windows-latest)` — `request_timeout::timeout_fires_cleanly_during_graceful_drain`** (`request_timeout.rs:280`, *"hung request must be released by the deadline during drain: Elapsed(())"*). 2020 passed, 1 failed. This head's entire diff is a single file — `autumn/tests/integration/sqlite_replication.rs` — which cannot touch request handling or graceful drain, and this same test passed on the previous Windows run at `5939523`. It is a deadline-based test on a runner where the job took 36 minutes.

**`Test (macos-latest)` — `sim_delayed_enqueue::delayed_enqueue_becomes_due_under_virtual_time` and `sim_fault_plan::a_test_app_without_a_plan_is_unchanged`** (2020 passed, 2 failed). Both fail with the same error: `job runtime is not initialized; register jobs with AppBuilder::jobs()`. That is a race on the process-global job runtime, first seen on this branch at `ec323af` on a *different* test in the same family; those tests pass in isolation, and the scenarios are seeded, so a genuine defect would fail deterministically rather than picking a new victim each run.

To be straight about the one connection that does exist: this PR adds tests to the same consolidated `integration_tests` binary, which changes scheduling and can make an existing global-state race fire more often. The defect itself is not this PR's, and the real fix is the one `CLAUDE.md` already prescribes — process-global-state tests belong in their own `[[test]]` binary — which is well outside this slice.

On the replication work itself, every lane agrees: all 32 replication tests pass on Windows, macOS and Ubuntu, and the previous head's Windows harness failure (13 failures, `destroy_database` unable to unlink a database the replicator still held open) is fully resolved in `a7d35ad` — 0 occurrences now.

---

PR #2484, opened 2026-09-03
Fix OpenAPI response schema loss when body guards expand before route macro (#1677)

### LINKED ISSUE #1677: Body-guard macros drop OpenAPI response schema when expanded before the route attribute
## Context

A Codex review on PR #1662 (per-route `#[throttle]` rate limiting) flagged that when a body-guard macro is written *above* the route method attribute, the generated OpenAPI document loses the handler's response schema.

Reference: PR #1662 — https://github.com/madmax983/autumn/pull/1662

This is the sibling concern to #1668. That issue tracks the *pre-body-extraction gating* redesign (running the guard before the body extractor). This issue tracks a separate, purely documentation-vs-metadata problem: OpenAPI response-schema loss caused by attribute-expansion ordering.

## Root cause

All four body-guard macros — `#[secured]`, `#[step_up]`, `#[authorize]`, and `#[throttle]` — rewrite the handler's return type to `Response` when they expand:

- `autumn-macros/src/throttle.rs:404-406` sets `input_fn.sig.output` to `-> ::autumn_web::reexports::axum::response::Response`.

The route macro infers the OpenAPI response body from the handler's signature output:

- `autumn-macros/src/route.rs:147` calls `api_doc::infer_response_body(&input_fn)`.
- `autumn-macros/src/api_doc.rs:419-425` reads `input_fn.sig.output`, unwraps `Result::Ok`, and looks for `Json<T>` (directly or inside a tuple) to emit the response schema.

Attribute macros expand outermost-first. So when a guard attribute is written *above* the method attribute (guard outermost, method inner), the guard expands first and rewrites `sig.output` to `Response`. By the time the route macro runs, the real `Json<T>` return type is gone and `infer_response_body` returns `None`.

## User-visible symptom

A route like:

```rust
#[throttle(limit = 5, per = "1m", key = "ip")]  // guard outermost
#[post("/users")]
async fn create() -> Json<Created> { /* … */ }
```

produces an OpenAPI operation with **no response schema**, whereas the opposite ordering (method attribute outermost) documents `Created` correctly. Throttling (and idempotency-replay accounting) is enforced correctly in both orderings — only the generated OpenAPI response schema differs.

## Current workaround

Place the route method attribute (`#[get]` / `#[post]` / …) *outermost*, above the guard attribute, so the route macro sees the real return type before any guard rewrites it. This is now documented for `#[throttle]` in the macro rustdoc and in `docs/guide/rate-limiting.md`.

## Proposed fix direction

Make response-schema inference independent of expansion ordering, applied consistently across all four body guards (`#[secured]`, `#[step_up]`, `#[authorize]`, `#[throttle]`):

- Have each guard preserve the original return type as recoverable metadata (e.g. an emitted marker attribute or a hidden type alias) that the route macro can read instead of the rewritten `sig.output`; or
- Have the route macro recover the original return type from the generated wrapper (e.g. the `let __autumn_inner: #ty = …;` binding the guards emit).

Either way the fix should live in one shared place so all four guards benefit and stay consistent.

## Scope

- `autumn-macros` only (macro-expansion metadata + route-macro inference). No runtime behavior change.
- Non-breaking preferred.

madmax983:
Fixes #1677.

## Problem

`#[secured]`, `#[step_up]`, `#[authorize]`, and `#[throttle]` all rewrite a handler's return type to `Response` when they expand as proc-macros. Rust expands attribute macros outermost-first, so when one of these guards is written *above* the route method attribute (`#[get]`/`#[post]`/…), it expands first — by the time the route macro runs and calls `api_doc::infer_response_body`, the real `Json<T>` return type is already gone from `sig.output`, and the generated OpenAPI document silently loses the route's response schema. The opposite ordering (route attribute outermost) worked correctly, which is why this previously needed a documented workaround.

## Fix

Every guard already binds the pre-rewrite type as `let __autumn_inner: T = (async move { … }).await;` around the guarded body before rewriting `sig.output`. `infer_response_body` now recovers the original type from that binding instead of trusting `sig.output` alone — recursing to the *innermost* binding when guards stack, since an outer guard's own binding necessarily reads back `Response` once an inner guard has already rewritten the signature.

The recovery is matched by the guard's **exact structural shape** — a `Local` binding sitting as the second-to-last statement of the block, immediately followed by the generated `IntoResponse::into_response(__autumn_inner)` tail (a new shared helper, `idempotency_guard::generated_inner_response_binding`) — not merely by variable name, so an unrelated handler-local that happens to share the name can never be mistaken for a guard's marker. This lives in one shared place (`autumn-macros/src/api_doc.rs` + the shared `idempotency_guard.rs` helper) so all four guards benefit identically.

**Scope**: macro-expansion metadata only. No authorization, rate-limiting, or idempotency-replay runtime behavior changes in either ordering. One legitimate, existing runtime reader of `ApiDoc::response` does change output for the specific fixed case: `mount_mcp`'s MCP tool-catalog eligibility gate (`should_expose`) previously excluded a guard-above-route JSON handler explicitly opted in via `#[api_doc(mcp)]` because it looked like it had no response schema; it's now correctly listed — fixing an availability gap, not changing what a call is authorized to do (every MCP call still dispatches through the identical, fully-guarded HTTP pipeline). Covered by a dedicated end-to-end test.

## Process

Built test-first (red/green/refactor), then reviewed from four independent angles (two correctness-focused, two security-focused agents) plus a simplicity/reuse pass. All four converged on the same real gap in the first cut — the recovery matched *any* same-named local anywhere in the body, so an unguarded handler with a coincidentally-named local could have its correct schema silently overridden. Fixed by tightening recovery to the guard's exact structural shape (described above), with a regression test proving the fix.

### Acceptance criteria (from the issue) and evidence

| AC | Evidence |
|---|---|
| Response-schema inference is independent of expansion order, for all four guards | `autumn-macros/src/api_doc.rs` unit tests (`infer_response_body_recovers_*`, `_ignores_unrelated_local_named_autumn_inner`, `_recovers_innermost_type_under_triple_stacked_guards`); `autumn-macros/src/route.rs` integration tests using the real macro codegen for each guard (`route_macro_infers_response_schema_when_{throttle,secured,step_up,authorize}_expands_first`, `_under_stacked_guards_above_route`); end-to-end tests with real compiled+run handlers in `autumn/tests/integration/openapi.rs` and `authorization_integration.rs` |
| Fix lives in one shared place, consistent across all four guards | Single implementation in `api_doc::infer_response_body`/`recover_guarded_return_type`, built on one shared structural-match helper (`idempotency_guard::generated_inner_response_binding`) all four guards' identical generated shape satisfies — no per-guard special-casing |
| Scope: `autumn-macros` only, no runtime behavior change | Diff confined to `autumn-macros/src/{api_doc.rs,idempotency_guard.rs,route.rs,lib.rs}` plus tests/docs; four independent review passes confirmed no security-relevant `ApiDoc` field (`secured`/`required_roles`/`required_scopes`/`authorize_bindings`/`public`/`agent_authority`) is touched; the one real runtime-observable consequence (MCP tool-catalog exposure) is explicitly documented and tested (`mcp_endpoint.rs::tools_list_includes_a_body_guard_written_above_the_route_attribute`) |
| Non-breaking | No public API signature changes; all 1054 pre-existing `autumn-macros` tests plus the full `openapi`/`authorization`/`mcp` integration suites pass unmodified alongside the new tests (1056 + 137 total, all green) |
| The issue's literal repro (`#[throttle]` above `#[post]` returning `Json<Created>`) now documents a response schema | `route_macro_infers_response_schema_when_throttle_expands_first` (exact repro) + `throttle_above_route_attribute_preserves_response_schema` (end-to-end) |
| Guard enforcement itself (throttling, idempotency-replay) stays correct, unregressed | Full pre-existing replay-layer/throttle-ordering suite passes unchanged; the fix's only write path is the `response:` field of the emitted `ApiDoc` literal |
| The now-unnecessary "route-attribute-outermost" workaround is removed from docs | `docs/guide/rate-limiting.md` and `#[throttle]`'s rustdoc (`autumn-macros/src/lib.rs`) updated to state both orderings produce identical documentation |

### Testing

- `cargo test -p autumn-macros --lib` — 1056/1056 passed
- `cargo fmt --all -- --check` — clean
- `cargo clippy -p autumn-macros --all-targets -- -D warnings` — clean
- `cargo test -p autumn-web --test integration_tests --features test-support,openapi,mcp -- authorization_integration:: openapi:: mcp_endpoint::` — 137/137 passed

---

https://claude.ai/code/session_01RthtGux1BKsZLMuQ8iAspE

---
_Generated by [Claude Code](https://claude.ai/code/session_01RthtGux1BKsZLMuQ8iAspE)_


madmax983:
**CI failure triage: `Test (macos-latest)` — not this PR's.**

`upgrades_in_place_under_load_without_dropping_a_connection_or_the_state` (`autumn/tests/live_upgrade.rs:257`) failed. This PR's diff touches only `autumn-macros/src/{api_doc.rs,idempotency_guard.rs,route.rs,lib.rs}` (OpenAPI response-schema recovery), `autumn/tests/integration/{openapi.rs,authorization_integration.rs,mcp_endpoint.rs}`, `CHANGELOG.md`, and `docs/guide/rate-limiting.md` — none of which `live_upgrade.rs` or its hot-upgrade/connection-handling code paths reference. `git diff` against the merge-base confirms `live_upgrade.rs` is byte-identical to `trunk-dev`.

The failing test is a real-load, real-timing hot-upgrade test (drops-a-connection-under-load) on the macOS runner, a combination with the highest flake surface in this suite. No panic message reached the captured job log (the log jumps straight from request-serving `INFO` lines to the stack backtrace), which is consistent with output interleaving under load rather than a deterministic assertion failure.

I'll re-run just this failed job once the workflow run finishes (GitHub currently refuses `rerun-failed-jobs` while sibling jobs — `Test (ubuntu-latest)`, `Test (windows-latest)` — are still in progress) to confirm the flake. If it fails identically on re-run, I'll root-cause it as this PR's regardless of the diff scope above.

---

madmax983:
**Follow-up on `Test (macos-latest)`: reproduced on re-run, confirmed unrelated, tracked separately.**

The one allowed re-run failed identically (same test, same ~5.4s duration, same assertion site). Per the drive-to-green policy, a second identical failure is treated as real rather than a flake, so I root-caused it rather than re-running again:

- The failure is entirely in `examples/hot-upgrade/tests/live_upgrade.rs` (`#[cfg(unix)]`-gated, so it never runs on Windows — consistent with `Test (windows-latest)` passing), a completely separate example crate this PR's diff never touches.
- `examples/hot-upgrade` and this test were introduced together in #1674 / PR #2345 (`fec5221`), before this PR's branch point, and are untouched by both this PR's diff and the `trunk-dev` merge commit here — so it's neither a regression from this PR nor from picking up `trunk-dev`.
- Both failures bottom out at `live_upgrade.rs:257`'s assertion that the predecessor process drains and exits `0` after an in-place `SIGUSR2` upgrade handoff, in ~5.4s total — too short for that assertion's own 30s timeout to have been reached, meaning the predecessor already exited non-zero shortly after the test's load-drain window. I couldn't extract the actual exit code or captured child-process logs from the CI job log (the consolidated test binary interleaves heavy tracing output from concurrent tests, and truncated fetches never reached the panic message), so this needs a local macOS repro or full log access to pin down further.

Opened autumn-foundation/autumn#2492 to track it, with the full analysis and reproduction pointers.

**This PR's own state**: `Test (ubuntu-latest)` and `Test (windows-latest)` both pass; all other 61 checks are green; Codex review clean across 3 rounds (PR opened, a lint fix, and the `trunk-dev` merge); no merge conflicts. `Test (macos-latest)` is the only red check, and it's pre-existing, unrelated breakage now tracked in #2492 rather than something this PR should carry a fix for.

---

PR #2512, opened 2026-09-04
Fix: guard lockout re-stamp against concurrent successful login (#2500)

### LINKED ISSUE #2500: 🪝 Snag: a concurrent successful login can be silently re-locked by a racing failed attempt (medium, repro 12/100)
🎯 **Charter**: Account-lockout state tour on `autumn generate auth`'s generated `/login` handler — a returning user logs in with the correct password at the same moment a wrong-password attempt (attacker credential-stuffing, or the user's own retried tab) pushes `failed_attempts` to the lockout threshold. Follow-up to the `docs/reports/2026-09-03-snag-reddit-clone-session.md` session, whose proposed next charter #2 was "no shipped example uses the generated lockout feature — needs a fresh fixture."

🐛 **Repro** — 12/100 in the cleanest run (a second run: 7/60; a third: 3/100 — intermittent, see conditions below), from a clean checkout:

1. Scaffold a fresh app and generate auth, patched to build against this checkout's `autumn-web`:
   ```bash
   autumn new lockoutapp && cd lockoutapp
   autumn generate auth User
   cat >> Cargo.toml <<'EOF'
   [patch.crates-io]
   autumn-web = { path = "/path/to/autumn/autumn" }
   EOF
   ```
2. `autumn.toml`: point `[database]` at a Postgres db, and set
   ```toml
   [auth.lockout]
   enabled = true
   threshold = 3
   cooloff_secs = 900
   ```
3. Run the app (migrates itself in dev), sign up `racer@example.com` / `CorrectHorse123`, and mark it confirmed (`update users set email_confirmed_at = now() where email = 'racer@example.com';` — sidesteps only the mail step, not the bug).
4. Before each trial, set `failed_attempts = 2, locked_at = NULL` (one below threshold).
5. Fire, truly concurrently (thread-barrier-synchronized start, not sequential shell backgrounding): one `POST /login` with the *wrong* password and one `POST /login` with the *correct* password.
6. Inspect the row after both complete.

Script (self-contained, `pip install requests`):
```python
import subprocess, sys, threading, requests
BASE, EMAIL, GOOD_PW, BAD_PW = "http://127.0.0.1:3000", "racer@example.com", "CorrectHorse123", "WrongPassword999"
PSQL = ["psql", "postgresql://postgres:postgres@127.0.0.1:5432/lockoutapp_dev", "-t", "-A", "-c"]
THRESHOLD = 3
def db(sql): return subprocess.run(PSQL+[sql], capture_output=True, text=True, check=True).stdout.strip()
def reset(): db(f"update users set failed_attempts={THRESHOLD-1}, locked_at=NULL where email='{EMAIL}';")
def state():
    fa, l = db(f"select failed_attempts, locked_at is not null from users where email='{EMAIL}';").split("|")
    return int(fa), l == "t"
def wrong(b, o):
    s = requests.Session(); b.wait()
    s.post(BASE+"/login", data={"email": EMAIL, "password": BAD_PW}, allow_redirects=False, timeout=10)
def good(b, o):
    s = requests.Session(); b.wait()
    r = s.post(BASE+"/login", data={"email": EMAIL, "password": GOOD_PW}, allow_redirects=False, timeout=10)
    o["status"], o["loc"] = r.status_code, r.headers.get("location", "")
for i in range(100):
    reset(); b = threading.Barrier(2); o = {}
    t1 = threading.Thread(target=wrong, args=(b, o)); t2 = threading.Thread(target=good, args=(b, o))
    t1.start(); t2.start(); t1.join(10); t2.join(10)
    fa, locked = state()
    logged_in = o.get("status") == 303 and "/account" in o.get("loc", "")
    if locked and fa < THRESHOLD and logged_in:
        print(f"trial {i}: BUG — fa={fa} locked_at set, but login {i} just succeeded (303 -> /account)")
```

On a hit: `failed_attempts = 0` (the successful login's own reset ran), `locked_at` is **set** anyway, and — confirmed by immediately issuing one more `POST /login` with the correct password on a fresh session — that follow-up is rejected `422 "Invalid email or password."` In every one of the 12 hits in the cited run, the follow-up was denied (12/12).

**Root cause** (`autumn-cli/src/generate/auth.rs`, generated into `src/routes/auth.rs`'s `login`, and duplicated verbatim in `reauth` — lines ~3913-4098 and ~4998-5079 in this checkout): the wrong-password branch does the failure-counting increment and the `locked_at` stamp as two *separate*, sequential `UPDATE`s (needed because Diesel can't express the conditional in one `.set(...)` — the code comment at line 3977 says so), gated only by `current_locked_at.is_none()` — a value captured from the request's *own initial SELECT*, never re-read. If a concurrent request's successful-login reset (the guarded `UPDATE … SET failed_attempts = 0, locked_at = NULL WHERE locked_at IS NULL OR locked_at <= expiry`, lines 4079-4098) commits in the gap between the wrong request's two statements, the second statement — `UPDATE users SET locked_at = now() WHERE id = ?`, unconditional on `failed_attempts` — reapplies the lock on top of the counter the successful login just cleared.

⚖️ **Oracle**: Two independent documented claims, both violated by this outcome:
1. `docs/guide/authentication.md` §Account lockout, point 4: *"A successful login clears the counter and the lock."* Here the lock does **not** stay cleared — it comes back after the login already completed.
2. `docs/reports/auth-pipeline-security-audit-2026-08.md`'s "Notable strengths" section states the lockout mechanism is *"success-path guarded against concurrent locking."* That's true in one direction only (the guarded reset UPDATE correctly refuses to clear an *already*-locked account, per `rows_cleared == 0` at line 4095) — it is not true in this direction: nothing guards the *lock-stamping* write against a login that already succeeded and reset the counter. This is a distinct race from L15 in that report (which is about concurrent *failures* collapsing at the cool-off boundary) — this one involves a concurrent *success*.

💥 **Impact**: Medium. Not a crash or data loss, and it self-heals after `cooloff_secs` (900s / 15 min by default) or an operator `/auth/admin/unlock`. But it is a genuine availability bug on the primary login path: a user who submits the *correct* password and is told they're logged in (303 to `/account`, session cookie issued) can be locked out of their own account for the full cool-off window by nothing more than one badly-timed concurrent wrong-password request — plausible from an attacker credential-stuffing the same account the real user happens to be logging into, or even from the user's own second browser tab/retry. Because the block is duplicated in `reauth` (the step-up re-auth handler gating sensitive actions like account deletion), the same race can strand a user mid-step-up too. Every app generated with `autumn generate auth` and default (or any) `[auth.lockout]` settings carries this — no shipped example currently exercises the generated lockout feature (confirmed in the prior reddit-clone session), so this had never been driven live before.

🔬 **Reproduce**: run the script above against a fresh scaffold per steps 1-4 (Postgres running locally, lockout `threshold = 3`). 12/100 in the cited run; reran twice more from a clean reset each time (7/60, 3/100) — consistently reproducible, intermittent because it depends on the successful login's DB round trip landing inside the ~sub-millisecond window between the wrong request's two `UPDATE`s, which widens under concurrent load (the reproduction above fires exactly one competing wrong-password request; a busier server, or an attacker firing several, only makes the window easier to hit, not harder).

📌 **Environment**: commit `e21283ca932c0ddb36316be05cfef69e778a6907` (branch `trunk-dev`), scaffolded via this checkout's `autumn new` + `autumn generate auth User`, `autumn-web` patched to the in-tree crate (`[patch.crates-io]`) so the generated code matches this commit exactly. PostgreSQL 16.13 (native service, no Docker available in this sandbox), rustc/cargo 1.94.1, Ubuntu 24.04 container. `[auth.lockout] threshold = 3, cooloff_secs = 900` (behavior is threshold-independent; 3 was chosen only to make trials fast). Driven directly over HTTP (Python `requests` + `threading.Barrier` for genuine concurrent start, not sequential shell backgrounding).

Dedup: searched issues/PRs for "lockout", "failed_attempts", "locked_at", "race" — found #1096 (the original lockout feature), #1796 (an unrelated `locked_at` re-check fix for magic-link verify), and the `auth-pipeline-security-audit-2026-08.md` report (L4, L13c, L15, L16 and the "success-path guarded" strength claim above) — none describe this specific successful-login-vs-concurrent-lock-stamp interleaving. Not a duplicate of L15 (that's concurrent *failures* colliding at the cool-off boundary; this is a concurrent *success* getting overwritten).

https://claude.ai/code/session_01Vj5hsfG615BwGfCT9WWit9

madmax983:
Closes https://github.com/autumn-foundation/autumn/issues/2500

## Summary

`autumn generate auth`'s generated `login` handler (and the duplicated `reauth` step-up block) counted a wrong-password attempt with an atomic `failed_attempts + 1` `UPDATE`, then — if the new count crossed `[auth.lockout].threshold` — stamped `locked_at` with a **second**, separate `UPDATE`. That second write was **unconditional** (`WHERE id = ?` only), gated only by a stale in-memory `current_locked_at` value read at the top of the request, never re-checked against the database.

If a concurrent request with the **correct** password committed its own "successful login resets the counter" `UPDATE` (`failed_attempts = 0, locked_at = NULL`) in the gap between the failed request's two statements, the unconditional lock stamp reapplied on top of that reset — silently re-locking an account for the full `cooloff_secs` window (default 15 min) even though it had **already** logged in successfully (303 redirect + session cookie already issued). The issue's own concurrent repro hit this 12/100 (7/60, 3/100 on reruns).

### Fix

The lock-stamp `UPDATE` in both `login` and `reauth` now filters on the row's *current* `failed_attempts` (`>= threshold`) and `locked_at` (`IS NULL`) at write time instead of trusting the in-memory read, so a concurrent successful reset makes the stamp a no-op rather than clobbering it. The `account_locked` telemetry event is now gated on the stamp having actually affected a row, so a losing race is never logged as a lock that didn't happen.

## Testing (TDD: red → green → refactor)

- **Generator meta-tests** (`autumn-cli/src/generate/auth.rs`): `routes_file_guards_lock_stamp_against_concurrent_reset` asserts both lock-stamp UPDATEs carry `failed_attempts.ge(lockout_cfg.threshold)` and `locked_at.is_null()` immediately before the `locked_at.eq(Some(now))` write (anchored to the write itself, not a bare substring count, so it can't be satisfied by an unrelated pre-existing guard elsewhere in the file); `routes_file_gates_lockout_telemetry_on_rows_actually_locked` asserts the telemetry is behind `if locked_rows > 0`. Both fail against the pre-fix generator source and pass after the fix; full `generate::auth::` suite (214 tests) still green.
- **Real Postgres regression suite** (`autumn/tests/integration/auth_lockout_race.rs`, testcontainer, `#[ignore]`d — picked up by CI's existing Docker sweep with no workflow edit): forces the exact bad interleaving deterministically (not a raw race, since the issue shows that only lands 3-12% of the time) across two genuinely separate connections:
  - `unconditional_lock_stamp_reproduces_2500_relock_after_concurrent_success` — characterizes the bug with the old pattern.
  - `guarded_lock_stamp_never_relocks_after_concurrent_success` — proves the fix.
  - `guarded_lock_stamp_winning_the_race_correctly_rejects_the_concurrent_login` — mirror ordering: when the lock genuinely wins, the concurrent login is correctly rejected, not silently granted.
  - `guarded_lock_stamp_still_locks_under_concurrent_attacker_only_race` — proves the guard can't be used to dodge a genuine lockout under pure concurrent wrong-password bursts (no legitimate login involved).
- All four scenarios were also independently hand-verified against a real local Postgres 16 instance via `psql` during development (Docker isn't available in this sandbox), matching every assertion in the automated suite.
- Reviewed the diff from three independent angles (correctness, security/attacker-bypass, test-quality) via separate review passes; the one real gap found (a vacuous string-count assertion that couldn't have caught a regression to the `locked_at.is_null()` half of the guard, plus missing coverage of the concurrent-attacker-only scenario) was fixed before this PR — see the two items above.
- `cargo fmt --all`, `cargo clippy -p autumn-cli --all-targets -- -D warnings`, `cargo clippy -p autumn-web --test integration_tests -- -D warnings`, and `cargo test -p autumn-web --test integration_tests --no-run` all clean. Scaffolded a fresh app against this checkout (`autumn new` + `autumn generate auth User` + `[patch.crates-io]`) and ran `cargo check --tests` on the generated project to confirm it still compiles.

## Acceptance criteria (derived from the issue) and evidence

| # | Criterion | Evidence |
|---|---|---|
| 1 | A successful login's reset of `failed_attempts`/`locked_at` must never be undone by a concurrent failed-login's lock-stamp write (`login`) | `guarded_lock_stamp_never_relocks_after_concurrent_success` + manual `psql` verification; meta-test `routes_file_guards_lock_stamp_against_concurrent_reset` |
| 2 | Same protection on the duplicated `reauth` step-up block (issue notes this can strand a user mid-step-up, e.g. account deletion) | Identical guard applied to `reauth`; meta-test asserts the guard appears in **both** handlers |
| 3 | `account_locked` telemetry must only fire when the account is actually locked | `if locked_rows > 0` gating; meta-test `routes_file_gates_lockout_telemetry_on_rows_actually_locked` |
| 4 | When the lock legitimately wins the race, the concurrent login must still be correctly rejected (not silently granted) | `guarded_lock_stamp_winning_the_race_correctly_rejects_the_concurrent_login` + manual verification |
| 5 | The fix must not weaken lockout against genuine concurrent credential stuffing (no new bypass) | `guarded_lock_stamp_still_locks_under_concurrent_attacker_only_race` + manual verification + independent security review (monotonicity argument: `failed_attempts`/`locked_at` can't move backwards without a password-verified reset) |
| 6 | Non-enumeration preserved (lockout response byte-identical to wrong-password response) | No change to the `return Err(auth_err())` / `reauth_form_err(...)` return paths; existing `routes_file_lockout_response_identical_to_wrong_password` unaffected; confirmed by both independent reviews |
| 7 | Generated code must still compile | Fresh scaffold + `cargo check --tests` against this checkout, clean |
| 8 | Regression must be provable deterministically, not just claimed (original bug's own repro was only 3-12% reproducible) | Explicitly-sequenced (non-flaky) testcontainer tests forcing the exact interleaving every run, cross-checked manually against a real Postgres instance |
| 9 | No regressions to existing lockout/auth behavior | Full `generate::auth::` suite (214 tests), `cargo fmt --check`, `cargo clippy -D warnings` on both touched crates, all clean |

No gaps found beyond what's listed above — everything identified during review was fixed in this PR rather than deferred.

---

https://claude.ai/code/session_01BTY1BjnsDzsy7KDsKgSXDL

---
_Generated by [Claude Code](https://claude.ai/code/session_01BTY1BjnsDzsy7KDsKgSXDL)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1604,6 +1604,39 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **🔒 `autumn generate auth`: a concurrent successful login could be silently
+  re-locked by a racing failed attempt (issue #2500):** the generated
+  `login` handler (and the duplicated `reauth` step-up block) counted a
+  wrong-password attempt with an atomic `failed_attempts + 1` `UPDATE` and
+  then, once the new count crossed `[auth.lockout].threshold`, stamped
+  `locked_at` with a *second*, unconditional `UPDATE ... SET locked_at =
+  now() WHERE id = ?` — gated only by a stale in-memory `current_locked_at`
+  value read at the top of the request, never re-checked against the
+  database. If a concurrent request with the *correct* password committed
+  its own "successful login resets the counter" `UPDATE`
+  (`failed_attempts = 0, locked_at = NULL`) in the gap between the failed
+  request's two statements, the unconditional lock stamp reapplied on top
+  of that reset — silently re-locking an account for the full `cooloff_secs`
+  window (default 15 minutes) even though it had *already* logged in
+  successfully (303 redirect and session cookie already issued). Reproduced
+  12/100 (and 7/60, 3/100 on reruns) with a genuinely concurrent
+  wrong-password/correct-password pair against a fresh `autumn generate
+  auth` scaffold.
+  Fix: the lock-stamp `UPDATE` in both handlers now filters on the row's
+  *current* `failed_attempts` (`>= threshold`) and `locked_at`
+  (`IS NULL`) at write time instead of trusting the in-memory read, so a
+  concurrent successful reset makes the stamp a no-op rather than
+  clobbering it; the `account_locked` telemetry event is now gated on the
+  stamp actually having affected a row, so a losing race is never logged
+  as a lock. New generator meta-tests (`autumn-cli/src/generate/auth.rs`)
+  assert both guards are present in the generated source, and
+  `autumn/tests/integration/auth_lockout_race.rs` (Postgres testcontainer)
+  reproduces the exact bad interleaving deterministically against a real
+  database — showing the pre-fix pattern re-locks the account and the fixed
+  pattern doesn't, plus the mirror-image ordering where a lock that
+  genuinely wins the race correctly rejects the concurrent login instead of
+  silently granting a session.
+
 - **🧭 Wayfinder: keyboard bypass-blocks link added to 6 supported example
   apps (a11y `bypass` Serious 7/8 → 0/8; `landmark-one-main` Moderate 1/8 → 0/8):**
   `autumn check --a11y` — the framework's own WCAG audit, run against each
--- a/autumn-cli/src/generate/auth.rs
+++ b/autumn-cli/src/generate/auth.rs
@@ -4004,52 +4004,72 @@ pub async fn login(
                 }};
 
                 if new_attempts >= lockout_cfg.threshold && current_locked_at.is_none() {{
-                    // Account transitions into the locked state — stamp locked_at
-                    // atomically. Propagate errors: if this write fails the account
-                    // is not locked despite the counter crossing the threshold, which
-                    // would allow a successful login to slip through.
-                    diesel::update({table}::table.find({snake_name}.id))
+                    // Account transitions into the locked state — stamp locked_at,
+                    // but only if the row still shows an over-threshold, not-yet-locked
+                    // state at write time. `current_locked_at` above is a stale
+                    // in-memory read from before the password check; without a
+                    // fresh DB-level guard, a concurrent *successful* login could
+                    // reset failed_attempts/locked_at between our increment and this
+                    // write, and this UPDATE would silently re-lock an account that
+                    // just logged in successfully (#2500). Filtering on the row's
+                    // current failed_attempts and locked_at makes the write a no-op
+                    // in that case instead of clobbering the reset. Propagate errors:
+                    // if this write fails the account is not locked despite the
+                    // counter crossing the threshold, which would allow a successful
+                    // login to slip through.
+                    let locked_rows = diesel::update(
+                        {table}::table
+                            .find({snake_name}.id)
+                            .filter({table}::failed_attempts.ge(lockout_cfg.threshold))
+                            .filter({table}::locked_at.is_null()),
+                    )
                         .set({table}::locked_at.eq(Some(now)))
                         .execute(&mut *db)
                         .await
                         .map_err(|e| AutumnError::internal_server_error_msg(&format!("Failed to lock account: {{e}}")))?;
 
-                    // Truncate to a coarse IP prefix (IPv4 /24, IPv6 /64) so
-                    // the telemetry event enables incident response without
-                    // logging a precise user identifier.
-                    let ip_prefix = match addr_ip {{
-                        std::net::IpAddr::V4(ip) => {{
-                            let [a, b, c, _] = ip.octets();
-                            format!("{{a}}.{{b}}.{{c}}.0/24")
-                        }}
-                        std::net::IpAddr::V6(ip) => {{
-                            let s = ip.segments();
-                            format!("{{:x}}:{{:x}}:{{:x}}:{{:x}}::/64", s[0], s[1], s[2], s[3])
-                        }}
-                    }};
-                    // Salt the digest with the deployment secret so the
-                    // account ID cannot be recovered by hashing small integers.
-                    let account_id_digest = {{
-                        use sha2::{{Digest, Sha256}};
-                        // Require a deployment secret for the digest salt. Operators
-                        // MUST set SECRET_KEY_BASE (already required for sessions) or
-                        // AUTUMN_ADMIN_SECRET. The static fallback prevents reversibility
-                        // only within this process; set the env var in production.
-                        let salt = std::env::var("SECRET_KEY_BASE")
-                            .or_else(|_| std::env::var("AUTUMN_ADMIN_SECRET"))
-                            .unwrap_or_else(|_| "autumn-lockout-fallback-salt".to_string());
-                        let hash = Sha256::digest(
-                            format!("{{}}:{{}}", salt, {snake_name}.id).as_bytes(),
+                    // Only emit lockout telemetry when this request's write actually
+                    // applied the lock — a concurrent successful login winning the
+                    // race above means the account never ends up locked, so it must
+                    // not be reported as such.
+                    if locked_rows > 0 {{
+                        // Truncate to a coarse IP prefix (IPv4 /24, IPv6 /64) so
+                        // the telemetry event enables incident response without
+                        // logging a precise user identifier.
+                        let ip_prefix = match addr_ip {{
+                            std::net::IpAddr::V4(ip) => {{
+                                let [a, b, c, _] = ip.octets();
+                                format!("{{a}}.{{b}}.{{c}}.0/24")
+                            }}
+                            std::net::IpAddr::V6(ip) => {{
+                                let s = ip.segments();
+                                format!("{{:x}}:{{:x}}:{{:x}}:{{:x}}::/64", s[0], s[1], s[2], s[3])
+                            }}
+                        }};
+                        // Salt the digest with the deployment secret so the
+                        // account ID cannot be recovered by hashing small integers.
+                        let account_id_digest = {{
+                            use sha2::{{Digest, Sha256}};
+                            // Require a deployment secret for the digest salt. Operators
+                            // MUST set SECRET_KEY_BASE (already required for sessions) or
+                            // AUTUMN_ADMIN_SECRET. The static fallback prevents reversibility
+                            // only within this process; set the env var in production.
+                            let salt = std::env::var("SECRET_KEY_BASE")
+                                .or_else(|_| std::env::var("AUTUMN_ADMIN_SECRET"))
+                                .unwrap_or_else(|_| "autumn-lockout-fallback-salt".to_string());
+                            let hash = Sha256::digest(
+                                format!("{{}}:{{}}", salt, {snake_name}.id).as_bytes(),
+                            );
+                            hex::encode(&hash[..8])
+                        }};
+                        tracing::warn!(
+                            event = "account_locked",
+                            account_id_digest = %account_id_digest,
+                            ip_prefix = %ip_prefix,
+                            failed_attempts = new_attempts,
+                            "account locked after repeated failed login attempts"
                         );
-                        hex::encode(&hash[..8])
-                    }};
-                    tracing::warn!(
-                        event = "account_locked",
-                        account_id_digest = %account_id_digest,
-                        ip_prefix = %ip_prefix,
-                        failed_attempts = new_attempts,
-                        "account locked after repeated failed login attempts"
-                    );
+                    }}
                 }}
                 return Err(auth_err());
             }}
@@ -5046,7 +5066,16 @@ pub async fn reauth(
                     1i32
                 }};
                 if new_attempts >= lockout_cfg.threshold && current_locked_at.is_none() {{
-                    diesel::update({table}::table.find({snake_name}.id))
+                    // See the login handler's matching guard (#2500): filter on the
+                    // row's current failed_attempts/locked_at rather than writing
+                    // unconditionally, so a concurrent successful reauth that already
+                    // reset the counter cannot be silently re-locked.
+                    diesel::update(
+                        {table}::table
+                            .find({snake_name}.id)
+                            .filter({table}::failed_attempts.ge(lockout_cfg.threshold))
+                            .filter({table}::locked_at.is_null()),
+                    )
                         .set({table}::locked_at.eq(Some(now)))
                         .execute(&mut *db)
                         .await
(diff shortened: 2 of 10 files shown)


madmax983:
**Check:** `App-variant conformance (sharded-owner, stable)` failed on `9b0641e` ([job](https://github.com/autumn-foundation/autumn/actions/runs/33931421839/job/101210637065)).

**Why it's not this PR's:** the job never reached any test body for this diff — it died during `cargo test -p autumn-cli --test generate generated_sharded_owner_scaffold_cargo_checks -- --ignored --exact`'s initial `Updating crates.io index` step, with `failed to get postgresql_archive as a dependency ... transfer too slow: failed to transfer more than 10 bytes in 30s`. That's a registry network timeout unrelated to anything this PR touches (auth lockout codegen + a new Postgres testcontainer test) — no fix to port.

I'll re-run the failed job once the rest of the workflow run finishes (GitHub rejects a re-run while the run is still in progress) and report back if it's still red.

---

madmax983:
**Check:** `Test (ubuntu-latest)` failed on `9b0641e` ([job](https://github.com/autumn-foundation/autumn/actions/runs/33931421785/job/101215337701)) — 4 tests in `autumn-macros`:
- `route::tests::route_macro_infers_response_schema_when_secured_expands_first`
- `route::tests::route_macro_infers_response_schema_under_stacked_guards_above_route`
- `route::tests::route_macro_infers_response_schema_when_throttle_expands_first`
- `route::tests::route_macro_infers_response_schema_when_step_up_expands_first`

**Why it's not this PR's:** this PR's diff touches only `autumn-cli/src/generate/auth.rs`, `autumn/tests/integration/{mod.rs,auth_lockout_race.rs}`, and `CHANGELOG.md` — nowhere near `autumn-macros`. This branch is fully up to date with `trunk-dev` (`git merge-base` == `origin/trunk-dev` HEAD), and I reproduced the identical panics running the same 4 tests locally against `trunk-dev` HEAD with none of this PR's changes applied. It's a real, deterministic regression (not a flake) on the base branch: `#2488` (merged just after `#2484`) redesigned how `#[secured]`/`#[step_up]`/`#[throttle]` emit their guard as a `FromRequestParts` gate parameter, which broke `#2484`'s `#1677` fix for recovering the original `Json<T>` response type across stacked/reordered guard expansion — `route_macro` now bails with its own "route macros can only be applied to functions" `compile_error!` on that shape.

No existing fix to port. Filed autumn-foundation/autumn#2516 with the full root-cause trace since it currently fails plain `Test (ubuntu-latest/macos-latest/windows-latest)` CI for every PR against `trunk-dev`, not just this one — out of scope to fix here. Once #2516 lands I'll merge `trunk-dev` into this branch to pick it up; keeping this PR watched until then.

---

madmax983:
Pushed `161ff96`, which fixes the `trunk-dev` regression from autumn-foundation/autumn#2516 directly (rather than waiting on a separate PR), since it was blocking this PR's CI:

- `autumn-macros/src/parse.rs`: new `split_leading_items_and_fn` parses macro input as a sequence of items and takes the trailing one as the handler, tolerating (and re-emitting) any leading items — the common single-function case is tried first so existing error messages are unchanged.
- `#[secured]`, `#[step_up]`, `#[throttle]`, `#[authorize]`, and the route macro (`#[get]`/`#[post]`/etc.) all use it now and thread the leading items through to their own output, instead of failing outright with "route macros can only be applied to functions" whenever one guard is stacked above another guard or the route attribute.
- `#[step_up]`/`#[throttle]` also stopped emitting their `__AUTUMN_STEP_UP_MAX_AGE`/`__AUTUMN_THROTTLE_ROUTE_ID` marker consts inside the handler body (only inside the new gate) after #2488 — restored a lightweight copy in the body so #1677's OpenAPI response-schema recovery still works for those two guards, mirroring what `#[secured]` already does.

Verified: all 1073 `autumn-macros` unit tests pass (including the 4 that were failing on `trunk-dev`), `cargo clippy -p autumn-macros --all-targets -- -D warnings` and `cargo fmt --check` are clean, and `cargo build -p autumn-web -p autumn-cli --tests` succeeds — that exercises every real `#[secured]`/`#[step_up]`/`#[throttle]`/`#[authorize]` route in the codebase, not just the unit tests.

I'll close autumn-foundation/autumn#2516 as fixed by this PR once it merges. Re-running CI now.

---

madmax983:
Pushed `4fa1b5f`: the previous commit's `Lint` job caught a `clippy::needless_pass_by_value` I missed locally (autumn-macros' clippy config runs `pedantic`+`nursery`, stricter than my local check) — `split_leading_items_and_fn` only ever borrows its `TokenStream`, so `authorize_macro`/`secured_macro`/`step_up_macro`/`throttle_macro`/`parse_async_handler_with_leading_items` no longer consume their own `item` param either once they call it with `&item`. Silenced with `#[allow(clippy::needless_pass_by_value)]` (documented why) rather than changing those signatures to `&TokenStream`, which would've rippled into ~80 existing by-value call sites across the proc-macro boundary and tests for a purely stylistic win.

Verified `cargo clippy -p autumn-macros --all-targets -- -D warnings` and `cargo fmt --all -- --check` are clean, and the full `autumn-macros` test suite (1073 tests) still passes.

---

madmax983:
**Check:** `Compile-and-serve gates (1.88.0)` failed on `4fa1b5f` ([job](https://github.com/autumn-foundation/autumn/actions/runs/33937835950/job/101229183545)) — `integration::scaffold_form_for::generated_attachment_scaffold_multipart_test_passes` (an `#[ignore]`d gate that scaffolds a fresh attachment/multipart project and runs its own generated `cargo test`).

**Why I judge this a flake, not a regression:** I reproduced the exact same test locally against this exact commit — `cargo test -p autumn-cli --test cli_tests -- --ignored --exact integration::scaffold_form_for::generated_attachment_scaffold_multipart_test_passes` — and it passed (`ok`, 207.97s). The sibling matrix job on the *same* commit, `Compile-and-serve gates (stable)`, ran the identical test at its step 26 and also passed. So it's 2-for-2 passing on this exact commit and 1 failure, isolated to the `1.88.0`-toolchain runner — consistent with the resource-contention flakiness this workflow's own comments already call out for these heavy scaffold-compile-and-run jobs, not a deterministic regression from this PR's diff.

Re-ran the failed job (`rerun_failed_jobs` on the run). Will report back if it's still red.

---

PR #2515, opened 2026-09-05
🧩 Derive a queryable architecture graph from the app's macros (#1747)

### LINKED ISSUE #1747: Derive a queryable architecture graph from the app's macros
## Problem (the "So What?")
Autumn already declares every architectural element — routes, models, repositories, jobs, auth requirements, tenancy scoping, plugins — through framework-owned proc-macros (`autumn-macros/src/route.rs`, `model.rs`, `repository.rs`, `job.rs`, `secured.rs`), yet none of that knowledge is retained as a queryable model of the running application. The only introspection that exists is DB-schema pull (`autumn-cli/src/generate/introspect.rs`) and a per-request dev inspector for N+1 detection (`autumn/src/inspector.rs`) — both partial and runtime-discovered. Teams reconstruct "which routes touch table X" or "what does changing this model break" by reading the whole codebase, and every architecture diagram drifts the moment it is drawn.

## User Story
As an engineer (or an AI agent) working on a large autumn app, I want to ask the running binary structural and impact questions about its own architecture, so that I can refactor, review, and onboard against ground truth instead of stale diagrams and full-codebase reads.

## Vision (R&D)
Two-to-three years out, every autumn binary is a self-modeling system: at compile time the framework derives a complete, typed graph of the application — nodes for routes, models, repositories, jobs, and plugins; edges for reads/writes, auth requirements, tenancy scope, and dependency — and embeds it in the single binary so the app can answer questions about itself at runtime. Because every element is declared through macros autumn owns, the graph is complete and cannot drift: there is no reflection gap and no external registry to keep in sync. This turns architecture into a queryable asset — impact analysis ("what breaks if I change this model") becomes O(query), security review and onboarding read the model instead of the codebase, and AI agents operate on a verified structural map rather than guessing. It is the substrate a whole category of tools — living docs, safe-refactor, agent navigation, operator forensics — can be built on.

- [ ] A build-time pass emits a versioned, machine-readable application-graph manifest for a reference app, with nodes for every `#[route]`/`#[static_get]`, `#[model]`, `#[repository]`, and `#[job]`/`#[scheduled]`/`#[task]`, plus each route's declared auth requirement.
- [ ] The manifest records edges connecting routes → repositories → models (which route reads/writes which table) and jobs → models.
- [ ] `autumn graph <query>` answers at least: "which routes/jobs touch model/table X" and "impact of changing model X" (the transitive set of affected routes and jobs) against the reference app.
- [ ] A completeness/drift test fails the build if any macro-declared route, model, repository, or job is absent from the manifest.
- [ ] The manifest is retrievable from a running single-binary build (embedded, not a side file) — e.g. via a dev/ops endpoint.

## Success Metric
On a reference app with a hand-verified ground truth, `autumn graph impact <model>` returns the affected routes and jobs with 100% recall (zero false negatives), and the drift test catches any declared element removed from the graph.

## Out of Scope
- Cross-service / distributed graphs spanning multiple deployments — single app only.
- Temporal / historical graph or diffing the graph across versions.
- Graphical visualization / diagram-rendering polish.
- Any agent or tooling that auto-modifies code based on the graph.
- Runtime/dynamic call-graph tracing — the first slice is derived statically from macro declarations.

## Gap Analysis
Rails ships `rails routes` (routes only, no dataflow) and discovers models via runtime reflection with no typed edges; Django introspection is schema/admin-oriented and partial; Phoenix and Next.js expose no unified server-side architecture model; loco mirrors Rails' gaps. Bare axum/actix/rocket have nothing to introspect — you wire routes and queries by hand, so there is no framework-owned declaration site to derive a graph from. leptos models a client reactive graph, unrelated. The autumn-shaped answer: because routes, models, repositories, jobs, auth, and tenancy are all declared through autumn's proc-macros (`autumn-macros/src/`), the framework can derive the complete graph at compile time — typed and drift-free — and, thanks to single-binary asset embedding (`autumn/src/assets.rs`), ship it inside the app so the binary can answer questions about itself. This is distinct from the auth-coverage security manifest (#1604/#1627, a proof not a queryable model), OpenAPI/MCP projection (#1117, external HTTP surface only), and the per-request dev inspector (`autumn/src/inspector.rs`).

## Complexity Tier
L — a new compile-time collection pass spanning multiple macros, a stable manifest format, a query engine, embedding into the binary, and a build-time drift gate.

madmax983:
Closes #1747.

## Problem

Autumn already declares every architectural element through proc-macros it owns — `#[route]`, `#[static_get]`, `#[model]`, `#[repository]`, `#[job]`/`#[scheduled]`/`#[task]` — but none of that survived expansion as something you could ask a question of. "Which routes touch table X" meant reading the whole codebase, and every architecture diagram was stale the moment it was drawn.

## What this does

Each declaring macro now registers a node with `inventory`; the framework assembles them into a typed graph at link time and embeds it in the binary.

```console
$ autumn graph impact Post
Changing model Post affects 1 repositor(y/ies), 12 route(s) and 2 job(s)
  repository PostRepository over Post
  route      GET /api/posts
  route      DELETE /r/{sub_slug}/posts/{post_slug}
  route      POST /posts/{post_id}/upvote
  …
  job        post_publication
  scheduled task hot-rank-calculator (every 15m)
```

* **Nodes** — one per declared element. Routes carry their mounted path and declared auth requirement (read off `ApiDoc`, so the graph *states* the posture `autumn routes audit` proves rather than deriving it a second time); models carry their table; repositories their model, table, `Pg*` type and auto-API prefix.
* **Edges** — each with an `access` (`read`/`write`/`read/write`) and a `provenance`: `declaration` (`#[repository(Post)]` names its model outright), `signature` (a handler taking `PgPostRepository`), or `body` (a model type, a `diesel` table module, or an identifier inside a raw-SQL literal).
* **`autumn graph show | touches <NAME> | impact <NAME>`**, with `--json`, `--manifest` and a `--check` drift gate that names *what* moved (`~ route GET /r/{slug} / auth secured -> none`).
* **`/actuator/graph`** — the same graph from the running single binary, sensitive-gated like `/actuator/env`.

## Design note: why the join happens at link time

A proc-macro sees one item at a time and cannot know which names are models, tables or repositories. So routes and jobs publish the *candidate* names in their own tokens — identifiers adjacent to a `::` separator, type-shaped identifiers, and identifiers inside string literals that read as SQL — and the framework resolves them against what actually got declared. An unresolvable candidate is simply dropped.

The collection is a deliberate superset with no stop-list. A missing edge is a false negative in an impact answer, which is the one failure this feature cannot afford; a spurious candidate costs nothing.

## Acceptance criteria

| AC | Evidence |
|---|---|
| Versioned manifest with nodes for every `#[route]`/`#[static_get]`, `#[model]`, `#[repository]`, `#[job]`/`#[scheduled]`/`#[task]`, plus each route's auth | `ArchitectureGraph { schema_version: 1, … }`; `every_macro_declared_element_is_in_the_graph`, `every_route_node_carries_its_declared_auth_requirement` |
| Edges routes → repositories → models, and jobs → models | `a_repository_extractor_links_the_route_to_the_repository`, `a_table_module_named_in_the_body_links_the_route_to_the_model`, `a_job_that_names_a_table_gets_a_model_edge` |
| `autumn graph <query>` answers "which routes/jobs touch X" and "impact of changing X" | `autumn graph touches <NAME>` / `impact <NAME>`; `touches_answers_for_a_table_name`, `impact_answers_for_a_model_name` |
| Completeness/drift test fails the build on a missing element | `examples/reddit-clone/tests/architecture_graph.rs` — plus `autumn graph --check` |
| Retrievable from the running single binary, embedded | `/actuator/graph`, served from the graph installed when the router is built |
| **Success metric:** 100% recall on `impact <model>` against hand-verified ground truth | `impact_of_the_post_model_has_total_recall` |

## Keeping it honest

The completeness gate is deliberately *not* derived from the registrations it checks — that would pass for a build whose graph module dropped every node it was handed. It censuses the reference app's **sources** for every declaring attribute, runs the binary's own graph dump, and fails when the two disagree. Verified non-vacuous by mutation: dropping models from the builder, and dropping one route's descriptor in the macro, each fail it.

Two real gaps that running against `examples/reddit-clone` exposed, both fixed rather than documented away:

* A GET page rendering `hx-delete=(…)` on a link was reported as a **writer**. A route's access is now its declared HTTP method alone; jobs, which declare no method, still use mutation evidence.
* `#[votable]` puts `react` on the model's **repository**, so the upvote and downvote routes reached the `votes` table without ever naming it. `#[model]` now publishes its relation tables, which took `touches votes` from 2 routes to 10.

## What the derivation cannot see

Carried in the document's own `limits` field, not just the guide: cross-function indirection is not followed (that is the dynamic call-graph tracing the issue puts out of scope), resolution is name-based so aliases and renames are not resolved, runtime-assembled SQL is invisible, `#[votable]`/`#[commentable]` relations are attributed at repository granularity, and access is declared intent rather than an executed statement. Routes mounted by other mechanisms (`#[ws]`, a raw `merge`/`nest` router) are named in `completeness.unmodelled_mounted_routes` rather than dropped.

## Testing

* 40 unit tests in `autumn/src/graph/`, 19 in `autumn-macros/src/graph.rs`, 9 in `autumn-cli/src/graph.rs`, 5 in `autumn/src/actuator.rs`, 7 in the reference app.
* `cargo test -p autumn-web --lib` — 5603 passed.
* Clippy clean on `autumn-web`, `autumn-macros`, `autumn-cli` with `--all-targets -D warnings`.
* Verified end to end against `examples/reddit-clone` via both `AUTUMN_DUMP_GRAPH=1` and the CLI.

Pre-existing on `trunk-dev` and untouched by this branch: four `route_macro_infers_response_schema_*` tests in `autumn-macros` fail on the base commit as well.

Two `#[edge]` cfg-gating tests in `route.rs` were updated to count three native companions rather than two — the graph node references `::autumn_web` like the route-info companion and path helper do.

https://claude.ai/code/session_0199pTxbHk8vqctgmFFtkiGb

---
_Generated by [Claude Code](https://claude.ai/code/session_0199pTxbHk8vqctgmFFtkiGb)_


madmax983:
## `Test` is red on the base branch, not on this PR

`Test (ubuntu-latest)` fails here with four `autumn-macros` unit tests:

```
route::tests::route_macro_infers_response_schema_under_stacked_guards_above_route
route::tests::route_macro_infers_response_schema_when_secured_expands_first
route::tests::route_macro_infers_response_schema_when_step_up_expands_first
route::tests::route_macro_infers_response_schema_when_throttle_expands_first
```

These are not this PR's. Three independent checks:

1. **The base branch's own CI is red on the same job.** Run [33909595295](https://github.com/autumn-foundation/autumn/actions/runs/33909595295) on `trunk-dev` at `66106a5c` — this PR's merge base — fails `Test` on **ubuntu, windows and macos**, with `Lint` passing. (That the same run's `Lint` passed is also why I treated the earlier lint failures here as mine: they were.)
2. **Reproduced locally against the base.** Checking out `origin/trunk-dev`'s `autumn-macros/src/route.rs` over this branch and running the four tests on rustc 1.98.1 reproduces all four failures identically.
3. **The diff is unrelated.** This PR's only change to `route.rs` is emitting an architecture-graph descriptor plus two `#[edge]` cfg-gating test updates; the failures are in OpenAPI response-schema inference under stacked body guards.

### Root cause, for whoever picks it up

Each of the four tests expands a body guard first and feeds the result to `route_macro`. `throttle_macro` (and its siblings) emit **three items** — a gate struct, its `FromRequestParts` impl, and then the handler:

```
# [doc (hidden)] # [allow (non_camel_case_types)] pub struct __AutumnThrottleGate_create ;
# [doc (hidden)] impl ... FromRequestParts <...> for __AutumnThrottleGate_create { ... }
async fn create () -> ... { ... }
```

`parse::parse_async_handler` (`autumn-macros/src/parse.rs:324`) does `syn::parse2::<ItemFn>(item)`, which requires the stream to be exactly one `fn`. It therefore fails with `expected \`fn\``, and the macro returns `compile_error!("route macros can only be applied to functions")` — which is the string each assertion reports.

### Proposed patch

`parse_async_handler` needs to accept a multi-item stream: parse it as `syn::File` (or a `Vec<syn::Item>`), take the single `Item::Fn`, and hand the remaining items back so `route_macro` can re-emit them ahead of its own output. Roughly:

```rust
pub fn parse_async_handler_with_prelude(
    item: TokenStream,
) -> Result<(Vec<syn::Item>, ItemFn), TokenStream> {
    let file: syn::File = syn::parse2(item.clone()).map_err(/* … */)?;
    let mut prelude = Vec::new();
    let mut handler = None;
    for it in file.items {
        match it {
            syn::Item::Fn(f) if handler.is_none() => handler = Some(f),
            other => prelude.push(other),
        }
    }
    // … async check, then Ok((prelude, handler.ok_or(…)?))
}
```

I have not made that change here: it touches the shared handler-parsing path for every route macro, it is unrelated to #1747, and widening this PR into it would put a cross-cutting macro change behind an architecture-graph review. Happy to open a separate PR for it if that is wanted.

I have not re-run the job either — a re-run cannot help a failure that is red on the base branch on all three platforms.

---

madmax983:
Follow-up to my earlier note on the red `Test` job: a fix for that base-branch breakage now exists, so I've ported it here rather than waiting.

`Test (ubuntu-latest / windows-latest / macos-latest)` all failed on `95eeed3` with the same four `route::tests::route_macro_infers_response_schema_*` tests (`1093 passed; 4 failed`), which are red on the base commit `trunk-dev@66106a5` too. #2518 (fixes #2516) carries the repair: #2488 moved the `#[throttle]`/`#[step_up]` checks into a sibling `FromRequestParts` gate and stopped emitting the guard's marker const into the handler body, which `api_doc::infer_response_body` needs in order to trust the `__autumn_inner` binding it recovers the pre-rewrite return type from.

`f9dcbf2` cherry-picks that commit verbatim — no edits, so it no-ops the moment `trunk-dev` carries the same change. Note it touches `autumn-macros/src/{route,step_up,throttle}.rs` and `CHANGELOG.md`, which are **not** part of the architecture-graph work; review those four files against #2518, not this PR.

Locally on rustc 1.98.1: `cargo test -p autumn-macros --lib` goes from 1093 passed / 4 failed to **1097 passed / 0 failed**.

---

madmax983:
## `Test (ubuntu-latest)` — the macro breakage is fixed; what's left is runner disk

The port in `f9dcbf2` did its job. On this head:

- **`Test (macos-latest)` — fully green**, every step success or skipped.
- **`Test (windows-latest)` — fully green**, `Run tests` passing at 93 minutes.
- **`Test (ubuntu-latest)` — steps 1–15 green**, including step 6 `Run tests` (the full `cargo test --workspace`, ~38 min) and the redis-TLS / direct-TLS / ACME / system-test lanes.

The four `route_macro_infers_response_schema_*` failures are gone on all three platforms.

### What is still red

Only **step 16, `Run Docker-dependent tests`**, and it never reaches a test body — it dies during the step's own compile:

```
##[warning]You are running out of disk space... Free space left: 23 MB
error: failed to write .../target/debug/deps/rustcCNKlyX/lib.rmeta: No space left on device (os error 28)
error: could not compile `bollard-stubs` (lib) due to 1 previous error
rustc-LLVM ERROR: IO failure on output stream: No space left on device
error: could not compile `bollard` (lib)
error: failed to write to `.../rmeta1M0MPZ/full.rmeta`: No space left on device (os error 28)
error: could not compile `autumn-web` (lib) due to 1 previous error
```

I re-ran the job once (attempt 2). It failed the same way at the same step, with 23 MB free instead of 41 MB — so this is deterministic, not a flake, and I have not re-run it again.

### Why this is the job's shape rather than the diff

Step 16 is the sixteenth consecutive step writing into one `target/` on one runner. By the time it starts, fifteen prior steps — the workspace suite plus eight feature-flipped rebuilds — have already filled the disk; it then has to compile the testcontainers stack (`bollard`, `bollard-stubs`, `tonic`, `opentelemetry-*`) and re-link `autumn-web` on top of that. The crates it dies on are dependencies this PR does not touch, and the same commit's macOS and Windows Test jobs are fully green.

#2519 is the fix already in flight: it moves this sweep into its own `test-docker` job on a clean runner, precisely because the sequential single-job shape is CI's critical path. Porting that here would mean restructuring the workflow (and repointing branch protection) from a feature PR, which is well outside this change.

In fairness I can't claim this PR adds *nothing* to the pressure — it adds ~5.6k lines to `autumn-web` and a test target to the reference app, and the headroom at that step is now measured in tens of megabytes. But at that margin nearly any recent commit is equally "the" tipping one, and nothing in this diff is the reason a sixteen-step job runs out of disk on its last step. If it matters before #2519 lands, the cheap check is whether `trunk-dev` clears step 16 on its own once its macro breakage is fixed.

Everything else on `f9dcbf2` is green: Lint, MSRV 1.88.0, cargo-deny, SQLite runtime, edge capsule conformance, sim sweep, plugin API contract, plugin freshness, security posture, example fleet e2e, migration/CLI-drift/markdown-link gates, all 24 app-variant conformance jobs, all 7 fuzz jobs, feature-combination compile gate, Windows Tier 1 journey, both compile-and-serve gates, both scaffold Postgres e2e gates, all three build+boot jobs, SemVer and the documentation build.

---

madmax983:
## Standing down on Codex round 18 — filed as #2520 item 20

Round 18's finding ("Preserve auth for declared plugin mounts") is **accurate**, and I verified it against the code rather than taking it at face value: `graph_declared_route_summary` builds a complete `RouteAuth` for every `declare_plugin_routes` route, then `add_auto_api_nodes` finds no owning repository (`repository_api: None`, correctly) and reduces the route to a bare `"METHOD /path"` string in `completeness.unmodelled_mounted_routes`. Flip such a route gated → public and the document is byte-identical, so `--check` passes.

I am **not** fixing it in this PR, and the reason is not that it is unimportant:

- **Nothing published is false.** The guide states the boundary outright — those mounts are *named*, not modelled, and "the document says what it could not see rather than reading as a complete account of the served surface." This is a capability the data supports and the schema does not yet express, not a misstatement to correct.
- **The fix is a schema decision.** `unmodelled_mounted_routes` is a `Vec<String>`. Carrying auth means changing that element's shape (or adding a structured sibling) at `schema_version` 1, and moving the drift-comparison and completeness-census readers with it. That belongs to a deliberate change, not to round 18 of an existing PR.

Filed with the full analysis and a proposed shape at #2520 (item 20), including why the two populations currently conflated in that field — mounts the graph *cannot* see versus mounts it *can* see but does not model — should be distinguished when it is done.

For the record on what this PR did absorb: rounds 15, 16 and 17 each found a genuine recall loss in `sql_shape` and all three were fixed here, because a dropped edge is the one failure this feature cannot afford. That class is now closed as far as conditional grammars go (`TRUNCATE`, `MERGE`, `WITH`), with the remaining closed-list gaps tracked as #2520 item 19. Round 18 is the first finding since that does not lose an edge, and it is where the line falls.

---

Issue #2516, opened 2026-09-05
trunk-dev is red: #2488 broke #1677's stacked-guard OpenAPI response-schema recovery in route_macro


madmax983:
## Summary

`Test (ubuntu-latest / macos-latest / windows-latest)` currently fails on `trunk-dev` HEAD (confirmed on commit `66106a5`, and reproduces identically and deterministically — not a flake — both in CI and locally with no other changes applied). 4 tests in `autumn-macros` fail:

- `route::tests::route_macro_infers_response_schema_when_secured_expands_first`
- `route::tests::route_macro_infers_response_schema_under_stacked_guards_above_route`
- `route::tests::route_macro_infers_response_schema_when_throttle_expands_first`
- `route::tests::route_macro_infers_response_schema_when_step_up_expands_first`

Each panics with:
```
a #[secured]-above-#[post] route must still document its Json<Created> response: :: core :: compile_error ! { "route macros can only be applied to functions" }
```

i.e. `route_macro` is bailing out with its own "not a function" `compile_error!` when fed the token stream produced by `secured_macro`/`throttle_macro`/`step_up_macro` — meaning it can no longer parse those guards' output as an `ItemFn` at all, so the `#1677` response-schema recovery this exact test suite was written to prove never even gets a chance to run.

## How I found this

While driving PR #2512 (an unrelated `autumn-cli` auth-lockout fix — no changes anywhere near `autumn-macros`) to green, `Test (ubuntu-latest)` failed on that PR's head. I confirmed it's not that PR's fault:
- The PR's branch is fully up to date with `trunk-dev` (`git merge-base` == `origin/trunk-dev` HEAD) and touches only `autumn-cli/src/generate/auth.rs`, `autumn/tests/integration/{mod.rs,auth_lockout_race.rs}`, and `CHANGELOG.md`.
- Running the same 4 tests locally against `trunk-dev` HEAD (no PR diff applied) reproduces the identical panics.

## Root cause (from bisecting recent history on `autumn-macros/src/route.rs`)

- `#2484` ("Fix OpenAPI response schema loss when body guards expand before route macro (#1677)") added these 4 tests and taught `route_macro` to recover the original `Json<T>` return type from the `__autumn_inner: T` binding a guard macro leaves behind when it rewrites `sig.output` to `Response`.
[...]

PR #2517, opened 2026-09-05
Fix #1668 guard-stacking regression: route macros can only be applied to functions

### LINKED ISSUE #1668: Evaluate pre-body-extraction gating for body-guard macros (#[secured], #[step_up], #[throttle])
## Context

A Codex review on PR #1662 (per-route `#[throttle]` rate limiting) flagged that the throttle check runs *inside the generated handler body* — that is, after Axum's `FromRequestParts` extractors have run, but crucially **after** the body extractor (`Json` / `Form` / `Multipart`) has already parsed and validated the request body.

This is not specific to `#[throttle]`: it is the shared pattern for all of our body-guard macros — `#[secured]`, `#[step_up]`, and now `#[throttle]`. The guard logic is emitted into the handler body, so it only runs once every extractor (including the body) has succeeded.

Reference: PR #1662 — https://github.com/madmax983/autumn/pull/1662

## Problem / why it matters

Running the guard after body extraction has two downsides:

- **DoS-shaped cost (especially for the rate limiter).** An over-limit client can still force the server to parse the entire request body and buffer uploads before it ever receives its `429`. The rate limiter is supposed to shed load cheaply, but in the current shape the expensive work (body parsing, multipart/upload buffering) happens *before* the limit decision is reached.
- **Correctness wrinkle.** If the body is malformed, extraction fails first and the client gets an extractor error (e.g. `400`/`422`) instead of the guard's intended response (`401`/`403`/`429`). The guard's outcome is effectively masked by body-extraction failures, which is surprising and inconsistent.

## Proposal

Evaluate reimplementing these guards — or offering an alternative form of them — as a `FromRequestParts` gate or a tower `Layer` that runs **before** the body extractor, so the guard decision is reached without parsing the body.

This should be a deliberate, **framework-wide** change applied consistently across all body-guard macros (`#[secured]`, `#[step_up]`, `#[throttle]`), not a one-off divergence for `#[throttle]`.

Tradeoffs to weigh:

- **Extractor-ordering guarantees in Axum.** All `FromRequestParts` extractors run before the single `FromRequest` body extractor, so a parts-based (or layer-based) gate can reliably run first. Confirm the ordering guarantee holds for our generated code and any nested/composed extractors.
- **Interaction with idempotency-replay body buffering (`body_guarded_replay`).** The replay path buffers and re-reads the body; a pre-body gate needs to compose cleanly with that mechanism rather than fighting it.
- **Macro ergonomics / signature stability.** Keep the macro invocation surface and generated handler signatures stable so this is non-breaking for existing call sites.

## Scope

- Framework-wide: `autumn-macros` + runtime.
- Non-breaking preferred.
- Note: PR #1662 intentionally keeps the existing body-guard pattern for consistency with `#[secured]` / `#[step_up]` and documents the limitation in place. This issue tracks the follow-up redesign rather than blocking that PR.

madmax983:
## Summary

`trunk-dev` is currently red on all three `Test` platforms (ubuntu/windows/macos-latest) — see [run 33909595295](https://github.com/autumn-foundation/autumn/actions/runs/33909595295) on 66106a5c, the tip at the time this was found. This blocks CI on every open PR against `trunk-dev`, unrelated to whatever that PR itself changes (confirmed on #2510, which only touches `hot-upgrade`).

Root cause: [#2488](https://github.com/autumn-foundation/autumn/pull/2488) (#1668) moved `#[throttle]`/`#[secured]`/`#[step_up]`'s runtime checks out of the handler body into each guard's own handler-unique `FromRequestParts` gate — a `struct` + `impl`, emitted as a sibling item ahead of the (still single) handler function — instead of a single modified function. Every route/guard macro's own `item` parser (`parse::parse_async_handler`, plus the three guard macros' own parsers, needed when one guard expands above another) only ever accepted a lone `ItemFn`, so a guard macro receiving that two-item output as `item` — any of `#[throttle]`/`#[secured]`/`#[step_up]`/`#[route]` written *above* another guard already using the #1668 gate shape — hit a spurious `route macros can only be applied to functions` compile error.

## Fix

- `parse::parse_async_handler_with_preamble` now accepts zero or more leading item definitions ahead of the trailing function, returning them separately so the route/guard macro re-emits that preamble (the gate type the function's new first parameter names) verbatim instead of choking on it.
- Fixing the parse gap surfaced a second, previously-masked bug: with parsing no longer failing first, `#[throttle]`/`#[step_up]` above `#[route]` still resolved to no OpenAPI response schema, because `api_doc::infer_response_body`'s `__autumn_inner`-binding recovery (#1677) only trusts that binding when one of `RESPONSE_REWRITING_GUARD_MARKERS`'s marker consts sits earlier in the *same* handler-body block — and #1668 moved `#[throttle]`'s/`#[step_up]`'s marker consts into their gate's `impl` block, out of the body, while `#[secured]`'s marker consts stayed in-body for exactly this reason. Both guards now also leave a dead-code copy of their marker const in the handler body (mirroring `#[secured]`'s existing `role_scope_consts` pattern), restoring the invariant `infer_response_body` relies on — including through stacked guards, where only the innermost binding carries the handler's real return type.

This surfaced via `route.rs`'s existing `route_macro_infers_response_schema_when_throttle_expands_first` / `..._when_step_up_expands_first` / `..._under_stacked_guards_above_route` tests, which predate this fix but were failing on `trunk-dev` and had never previously reached the code path they exercise (the parse error masked the schema-recovery bug underneath it).

## Test plan

- [x] `cargo test -p autumn-macros --lib` — 1073 passed, 0 failed
- [x] `cargo test -p autumn-web --test integration_tests -- secured_route:: step_up_route:: throttle_route:: authorization_integration::` — 47 passed, 0 failed
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p autumn-macros --all-targets -- -D warnings`

https://claude.ai/code/session_01KJy5FusFsz4hLGr5BFZzBb

---
_Generated by [Claude Code](https://claude.ai/code/session_01KJy5FusFsz4hLGr5BFZzBb)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3186,6 +3186,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   function that quietly delegated to its borrowed twin would fail even though
   its output is correct.
 
+### Fixed
+
+- **macros: `#[throttle]`/`#[secured]`/`#[step_up]` above the route macro no
+  longer fail to compile, or silently drop the OpenAPI response schema, once
+  stacked (#1668 regression):** #1668 moved these three guards' runtime
+  checks out of the handler body into each guard's own handler-unique
+  `FromRequestParts` gate — `struct` + `impl`, emitted as a sibling item
+  ahead of the (still single) handler function. Every route macro's own
+  `item` parser (`parse::parse_async_handler`, plus the three guard macros'
+  own parsers, needed when one guard expands above another) only ever
+  accepted a lone `ItemFn`, so a guard macro receiving that two-item output
+  as `item` — any of `#[throttle]`/`#[secured]`/`#[step_up]`/`#[route]`
+  written *above* another guard already using the #1668 gate shape — hit a
+  spurious `route macros can only be applied to functions` compile error.
+  `parse::parse_async_handler_with_preamble` now accepts zero or more
+  leading item definitions ahead of the trailing function, returning them
+  separately so the route/guard macro re-emits that preamble (the gate type
+  the function's new first parameter names) verbatim instead of choking on
+  it.
+
+  Fixing the parse gap surfaced a second, previously-masked bug: with
+  parsing no longer failing first, `#[throttle]`/`#[step_up]` above
+  `#[route]` still resolved to no OpenAPI response schema, because
+  `api_doc::infer_response_body`'s `__autumn_inner`-binding recovery (#1677)
+  only trusts that binding when one of `RESPONSE_REWRITING_GUARD_MARKERS`'s
+  marker consts sits earlier in the *same* handler-body block — and #1668
+  moved `#[throttle]`'s/`#[step_up]`'s marker consts into their gate's
+  `impl` block, out of the body, while `#[secured]`'s marker consts stayed
+  in-body for exactly this reason. Both guards now also leave a dead-code
+  copy of their marker const in the handler body (mirroring `#[secured]`'s
+  existing `role_scope_consts` pattern), restoring the invariant
+  `infer_response_body` relies on — including through stacked guards, where
+  only the innermost binding carries the handler's real return type.
+
+  Regression-proven at both the macro-expansion level (`route.rs`'s
+  existing `route_macro_infers_response_schema_when_throttle_expands_first`
+  / `..._when_step_up_expands_first` / `..._under_stacked_guards_above_route`
+  tests, which predate this fix but never previously reached the code path
+  they exercise) and end to end: `cargo test -p autumn-macros --lib` (1073
+  passed), `cargo fmt --all -- --check`, `cargo clippy -p autumn-macros
+  --all-targets -- -D warnings` all clean.
+
 ## [0.7.0] - 2026-08-23
 
 For a narrative tour of this release, see the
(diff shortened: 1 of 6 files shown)


madmax983:
`Test (macos-latest)` failed on d6ed1c8, but not from this PR's change: `integration::cache_stampede::swr_serves_stale_and_refreshes_in_background` failed ("background refresh must publish the new value after it finishes computing") — a stale-while-revalidate cache timing test with no relation to `autumn-macros`/route guards. This PR doesn't touch caching code. Re-running once the workflow finishes.

---

madmax983:
Closing as superseded: `trunk-dev` already carries an independently-authored, functionally-equivalent fix for this same #1668 guard-stacking regression, merged via commit `2d1863e` (bundled inside #2512 alongside an unrelated lockout-race fix, with the actual macro fix itself apparently landing as #2516). That's the source of this PR's current merge conflicts — both PRs fixed the identical root cause (parse failure + OpenAPI schema-recovery gap) with different implementations.

Since `trunk-dev`'s version is already reviewed and merged, there's nothing left for this PR to contribute. The equivalent fix has also been carried onto #2510's branch by merging `trunk-dev` in and taking its implementation for the four macro files (verified via `cargo test -p autumn-macros --lib` — 1073 passed — plus `cargo test -p hot-upgrade --test live_upgrade` to confirm #2510's own substance is unaffected).

---

PR #2527, opened 2026-09-05
🚦 Semaphore: CI health follow-up (escape found: #2488 merged 4 commits stale)

madmax983:
Follow-up to `docs/reports/2026-09-04-semaphore-ci-health-census.md` (#2504). That census closed with two open items: a macOS-vs-ubuntu timing cluster pending a load-faithful rerun campaign, and "escape mining: not run this pass." This PR runs the escape mining (free — GitHub API + git ancestry only, no new CI spend) and ships the rerun campaign as a dispatchable harness instead of only a description. No fix PR is included — nothing here clears this role's own hard gate for one.

## 🎯 Verdict path

Unchanged from the 2026-09-04 census, with one correction: **`trunk-dev`'s own tip (`66106a5c`) is currently red** — all three `Test` matrix legs fail 4 `autumn-macros` unit tests (`compile_error!{"route macros can only be applied to functions"}`), confirmed via the Actions API. Two fix PRs are already open (#2517, #2518) — this PR does not duplicate a third and this PR's own `Test` checks will show that same pre-existing failure, unrelated to this diff (docs + a new opt-in workflow only).

## 🌡️ Symptom — an escape found by git ancestry, not sampling

```
$ git merge-base --is-ancestor f29d4b4a 02177caf && echo ancestor
ancestor
$ git log --oneline f29d4b4a..02177caf | wc -l
4
```

PR #2488 ("Gate `#[secured]`/`#[step_up]`/`#[throttle]` before body extraction", #1668) merged into `trunk-dev` at `2026-09-04T18:24:15Z`. Its recorded `base.sha` is `f29d4b4a` — 4 commits behind PR #2484 ("Fix OpenAPI response schema loss when body guards expand before route macro"), which merged over 9 hours *earlier*. #2488's branch was opened before #2484 even existed and, per its `base.sha`, was never updated against `trunk-dev` before merging. Its CI therefore validated #2488 against a `trunk-dev` state that never included #2484's change — not the state the merge actually produced. Both PRs are individually correct and each passed its own tests; they don't compose, and nothing in the pipeline ever tested them together before merge. `trunk-dev` went red the moment #2488 landed and stayed red for hours — PR #2517 opens by saying exactly that, and drew two independent, non-duplicate-aware fix PRs 87 seconds apart (#2517, #2518).

This is Law 1's shape: #2488's green was true of a `trunk-dev` that never existed. I could not find a branch-protection read in the tools available to this session to confirm "require branches up to date before merging" directly — the `base.sha` staleness is the evidence, not a settings read — so this is a strong inference, not a confirmed setting.

No new organic evidence on the 2026-09-04 macOS cluster (`live_upgrade`/`cache_stampede`/`sim_fault_plan`) since the last census's window closed — but `trunk-dev` being red for large stretches since then means many `test` jobs aborted on the unrelated macro error before reaching those tests, which suppresses the sample rather than confirming absence.

## 🔍 Diagnosis

**Escape root cause**: merge-queue/branch-protection gap — required checks can pass against a stale base and still gate the merge. Pipeline-configuration defect, not a test defect.

**`trunk-dev` breakage itself** (already correctly diagnosed elsewhere, not this PR's fix): #2488 moved guard checks into sibling `FromRequestParts` gate items instead of the handler body; #2484's schema-recovery logic and every guard macro's item-parser still assumed the old single-`ItemFn` shape. Both #2517 and #2518 fix the macro codegen itself rather than touching any test's tolerance — consistent with this role's "product bug, not test tolerance" discipline even where Semaphore isn't the one triaging it.

## 🔧 Treatment

No fix PR: the escape is a process/config gap outside this role's authority to change unilaterally (ask-before: branch protection); `trunk-dev`'s breakage already has two adequate fix PRs in flight; the macOS cluster still has no rerun-campaign evidence to act on.

What ships instead (Acceptable outcome #4, a harness):

- **`.github/workflows/manual-macos-contention-check.yml`** — the Tier 1 load-faithful rerun protocol the 2026-09-04 census described but didn't commit: a `workflow_dispatch` job, 10 (or 5/20) independent fresh `macos-latest` VMs, each running unfiltered `cargo test --workspace --no-fail-fast` against a pinned commit, classifying each of the three target tests per sample. **Not triggered by this PR** — 10 macOS runners is real spend (this role's own ask-before item), and it needs a green `trunk-dev` commit to pin to (current tip won't get past the macro failure).
- **`docs/ci-health/quarantine-ledger.md`** — the formal intake-form ledger (owner, diagnose-by date, rerun baseline, mechanism, linked issue) the 2026-09-04 census flagged as this repo's one open admissions gap. Opens at zero open entries (the one pre-existing quasi-quarantine, `cancelled_release_does_not_leak_lock`, was already fixed in #2479); backfills that closure and tracks the three under-investigation macOS tests without quarantining any of them.

## 📊 Measurement

No before/after rerun stats this pass (nothing touched). Ledger delta: 0→0 open, 0→1 closed (backfilled), 0→3 under-investigation.

## 🔬 Reproduce

Escape finding: `git merge-base --is-ancestor f29d4b4a 02177caf` (from a checkout with both PRs' commits fetched).

Contention campaign: dispatch `manual-macos-contention-check.yml` with `sha` pinned to a post-fix `trunk-dev` commit — **needs sign-off first** (new CI spend).

## Recommendation for a human

Confirm and, if unset, enable "require branches up to date before merging" (or an equivalent always-re-test-the-merge-result policy) on `trunk-dev`. That single change would have turned this incident into "#2488's own CI catches the conflict before merge" instead of hours of red plus duplicate fixes.

---

https://claude.ai/code/session_01R3rAPRpdeRiSwwVmALHZSV

---
_Generated by [Claude Code](https://claude.ai/code/session_01R3rAPRpdeRiSwwVmALHZSV)_

### DIFF
--- a/.github/workflows/manual-macos-contention-check.yml
+++ b/.github/workflows/manual-macos-contention-check.yml
@@ -1,0 +1,123 @@
+name: Manual macOS contention check (Semaphore rerun harness)
+
+# Not part of the merge gate and never runs on push/pull_request. This is the
+# Tier 1 load-faithful rerun protocol from
+# docs/reports/2026-09-04-semaphore-ci-health-census.md's Reproduce section,
+# committed so a human can dispatch it with one click instead of hand-rolling
+# it each time the macOS timing-cluster hypothesis needs re-checking.
+#
+# It exists to answer one question: does `cargo test --workspace` on a real,
+# freshly-provisioned GitHub-hosted `macos-latest` runner reproduce the
+# `live_upgrade` / `cache_stampede` / `sim_fault_plan` failures the 2026-09-04
+# census clustered on macOS (3/17, 1/17, 1/17) against 0/16-17 on
+# `ubuntu-latest`? Ten independent fresh-VM samples, each running the exact
+# CI command unfiltered, is real GitHub Actions spend (macOS is the most
+# expensive hosted runner class) -- ask before dispatching this, per
+# Semaphore's own "Ask before: New CI spend" rule. This commit only adds the
+# capability; it does not run it.
+#
+# Two things this harness must get right or the samples don't mean anything
+# (both called out explicitly in the census this harness implements):
+#   - each sample needs its OWN fresh runner/VM, not a loop inside one job,
+#     since a shared target/ and warm build cache would let samples 2-10
+#     inherit sample 1's state instead of each reproducing the same cold,
+#     ~25-minute build-up the macOS-vs-ubuntu comparison depends on -- a
+#     `strategy.matrix` dispatch gives that for free;
+#   - it must run on actual `macos-latest`, not "some Mac", since the
+#     hypothesis under test is contention specific to that shared, hosted
+#     runner class.
+on:
+  workflow_dispatch:
+    inputs:
+      sha:
+        description: >-
+          Commit to reproduce. Pin this explicitly -- do not leave it to
+          whatever the dispatching branch's tip happens to be, or the ten
+          samples are not a same-commit baseline. As of 2026-09-05, trunk-dev's
+          own tip is red (autumn-macros guard-stacking regression, #2516/#1668
+          -- see docs/reports/2026-09-05-semaphore-ci-health-followup.md), so
+          pin to a commit where `cargo test --workspace` can actually complete
+          -- e.g. trunk-dev's tip once #2517 or #2518 merges -- or every
+          sample will abort at the first failing binary before ever reaching
+          the three target tests.
+        required: true
+      samples:
+        description: "Number of independent fresh-VM samples to run"
+        required: false
+        default: "10"
+        type: choice
+        options: ["5", "10", "20"]
+
+env:
+  # Copied from ci.yml's top-level `env:` block, not reinvented:
+  # RUST_MIN_STACK in particular changes how close a deep call frame sits to
+  # overflowing the default thread stack, which is exactly the kind of
+  # platform-dependent resource-margin difference this campaign needs to hold
+  # constant against the runs it's reproducing.
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+  RUST_MIN_STACK: 8388608
+
+jobs:
+  # Named `test`, not `sample`: Swatinem/rust-cache keys its cache entry on
+  # the job id by default, so a job called `sample` would never hit the cache
+  # ci.yml's own `test` job has been warming on this branch, and every sample
+  # would cold-build the entire dependency graph -- a different resource
+  # profile from the runs actually being reproduced. Reusing the job id `test`
+  # hits that same cache entry (best-effort: `continue-on-error: true` below,
+  # matching ci.yml:346, means a cache-service miss falls through to a cold
+  # build rather than failing the sample).
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        # Static 20-wide matrix with a `sample_count` filter, not a dynamic
+        # `fromJSON` range: workflow_dispatch `choice` inputs are strings
+        # evaluated at trigger time, and the simplest thing that reliably
+        # gives "the first N of a fixed list" without a matrix-generation
+        # job is to always define all 20 and skip past `samples`.
+        n: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
+    runs-on: macos-latest
+    if: fromJSON(inputs.samples) >= matrix.n
+    steps:
+      # Pinned to the dispatch input, not the branch tip: a workflow_dispatch
+      # run with no `ref` checks out whatever the dispatching branch currently
+      # points to, which could drift between samples -- pinning `sha` is what
+      # makes this a same-commit baseline instead of a moving target.
+      - uses: actions/checkout@v7
+        with:
+          ref: ${{ inputs.sha }}
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2.7.8
+        continue-on-error: true
+      - name: Run the exact CI command, but keep going past a failure
+        # --no-fail-fast matters: plain `cargo test --workspace` (what ci.yml
+        # actually runs) stops at the FIRST failing test binary and never
+        # runs the rest -- confirmed empirically in the 2026-09-04 census.
+        # Without it, one sample's results for the other two target tests go
+        # silently missing instead of recording a clean pass.
+        run: cargo test --workspace --no-fail-fast 2>&1 | tee run.log
+      - name: Classify this sample's result for each target test
+        # Three independent checks, not one grep with a single pass/fail
+        # fallback: a combined grep only reports "nothing matched at all", so
+        # if two of the three targets produced a result line and one didn't,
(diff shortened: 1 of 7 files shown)


madmax983:
`Test (ubuntu-latest)` (and the same across the matrix) failed on the head commit this PR opened with — not this PR's fault: `trunk-dev`'s own tip is red with the identical `autumn-macros` `compile_error!{"route macros can only be applied to functions"}` signature (see this PR's own description, and confirmed independently by #2517/#2518, which diagnose the same root cause: PR #2488 merged 4 commits behind PR #2484's base, so the two never got tested together — this PR's own escape finding).

Ported the fix rather than waiting: cherry-picked `5e5c27c2` from #2518 (`fix(macros): restore OpenAPI response-schema recovery for throttle/step_up guards`) onto this branch and pushed. Verified locally before pushing: `cargo test -p autumn-macros --lib` (1073/1073), `cargo fmt --all -- --check` (clean), `cargo clippy -p autumn-macros --all-targets -- -D warnings` (clean; the one pre-existing `unknown lint: clippy::unused_async_trait_impl` warning is unrelated, tracked in #2295/#2298/#2304). This will no-op once `trunk-dev` carries the real fix (whichever of #2517/#2518 merges first) and this branch is rebased/merged past it.

---

madmax983:
`Test (macos-latest)` failed on the current head (`2269ef27`) — not this PR's fault, and not the macro regression from the previous comment (that's fixed now: `ubuntu-latest`/`windows-latest` `Test` progressed past the point it used to abort at). This is `hot-upgrade`'s `live_upgrade::upgrades_in_place_under_load_without_dropping_a_connection_or_the_state`, `assert_eq!(connect_errors, 0)` failing `left: 1, right: 0` at `tests/live_upgrade.rs:268` — the exact macOS timing-cluster flake this PR's own report and `docs/ci-health/quarantine-ledger.md` are tracking under investigation (2026-09-04 census: 3/17 macOS executions, 0/16-17 ubuntu). This PR doesn't touch `hot-upgrade` or anything it depends on.

No verified fix to port: PR #2510 attempts a reclassification of this test's error handling but, per its own description, could not be verified against a real macOS run, so it doesn't meet this repo's bar for a de-flake yet. Per this role's own rules, a raised timeout/retry/tolerance change isn't a valid response to an unconfirmed flake either — so there's nothing safe to port. Re-running once now, per this PR's one-retry allowance for a suspected flake with an existing baseline.

---

madmax983:
🚦 Semaphore follow-up: `manual-macos-contention-check.yml` has never actually registered with GitHub, so as-is it can't be dispatched once this merges.

Evidence (zero new CI spend — no macOS runners touched):
- `GET /repos/.../actions/workflows` reports this workflow's `name` as its own file path (`.github/workflows/manual-macos-contention-check.yml`), not the declared `name: Manual macOS contention check (Semaphore rerun harness)` — the standard symptom of a workflow GitHub couldn't parse. Compare `ci.yml`, whose registered name is `CI`, matching its file exactly.
- All three pushes to this branch that touched the file (`33960142406`, `33962504398`, `33982014341` — including the "trim overexplained comments" edit, so this isn't comment-related) produced a run for it with **zero jobs** and `conclusion: failure`.
- The rendered run page gives the exact parse error: `Invalid workflow file: .github/workflows/manual-macos-contention-check.yml, Line 34, Column 9 — Unrecognized named-value: 'matrix'. Located at position 29 within expression: fromJSON(inputs.samples) >= matrix.n`.

Root cause: `matrix` is only a valid named-value inside `jobs.<job_id>.strategy` and `jobs.<job_id>.steps[*]` — not in `jobs.<job_id>.if`, where this workflow uses it to filter the static 20-wide matrix down to the requested sample count. GitHub rejects the whole file before any job is ever created, so `workflow_dispatch` never even offers it in the UI.

None of this shows up as a PR check (an invalid `workflow_dispatch`-only file produces no check run), which is presumably how it went unnoticed.

Suggested fix — compute the sample list in a small setup job instead of filtering a static matrix, so nothing needs `matrix` inside `if`:

```yaml
jobs:
  setup:
    runs-on: ubuntu-latest
    outputs:
      matrix: ${{ steps.gen.outputs.matrix }}
    steps:
      - id: gen
        run: |
          n=${{ inputs.samples }}
          echo "matrix=$(seq 1 "$n" | jq -R . | jq -s -c .)" >> "$GITHUB_OUTPUT"

  test:
    needs: setup
    strategy:
      fail-fast: false
      matrix:
        n: ${{ fromJSON(needs.setup.outputs.matrix) }}
    runs-on: macos-latest
    steps: ...  # unchanged
```

This also only spins up exactly N macOS VMs instead of 20 matrix slots with `20-N` shown as skipped, which is a nicer side effect for a job that's explicitly ask-before-dispatch due to cost.

Full writeup: `docs/reports/2026-09-06-semaphore-ci-health-followup.md` (#2548). Not opening a competing PR for this since it's your open branch — flagging here so it can be fixed before merge rather than after someone tries to dispatch it and gets nothing.

---

madmax983:
Merge conflict resolved: `trunk-dev` picked up #2518 (the fix I'd ported a copy of), so `CHANGELOG.md`/`step_up.rs`/`throttle.rs` conflicted with my duplicate of the same change. Merged `trunk-dev` in and took its side for all three — verified identical to `trunk-dev`'s versions, `cargo test -p autumn-macros --lib` (1142/1142), `cargo fmt --all -- --check`, `cargo clippy -p autumn-macros --all-targets -- -D warnings` all clean. Pushed.

---

madmax983:
`Test (ubuntu-latest)` failed on `53d919a0` (the merge commit) — not this PR's fault. `integration::request_timeout::timeout_fires_cleanly_during_graceful_drain` panicked: the client task saw `ConnectionReset` mid-request during a graceful-drain timeout scenario. `request_timeout.rs` hasn't been touched in this PR or recently on `trunk-dev` (last change: an unrelated docs commit), and this PR's diff doesn't touch anything in the request/timeout/drain path. Reads as a new, previously untracked flake — not one of the three macOS tests this PR is already tracking.

No fix to port (undiagnosed, no rerun baseline). Will re-run once the rest of this run finishes (macOS/Windows `Test` are still in progress; GitHub won't rerun failed jobs until the whole run completes).

---

madmax983:
Deflaked `timeout_fires_cleanly_during_graceful_drain` directly instead of rerunning (`df3b7d95`).

Mechanism: the test guessed the `/hang` handler was in-flight after a flat 40ms sleep, then triggered `shutdown.cancel()`. Under CI load that guess can fire before the connection is even `accept()`ed by the userspace accept loop (only TCP-handshaked, sitting in the kernel backlog) — the graceful-shutdown teardown drops the listener while that connection is still unaccepted, and the kernel resets it, producing exactly the observed `ConnectionReset` panic. Unrelated to axum's actual timeout/drain composition, and unrelated to this PR's diff.

Fix: the `/hang` handler now signals a `Notify` on entry; the test awaits that instead of sleeping a guessed duration, so shutdown only fires once the request is confirmed in-flight — no more timing guess.

Verified: 15/15 clean on a same-commit rerun (`cargo test -p autumn-web --test integration_tests --features test-support -- --exact ...timeout_fires_cleanly_during_graceful_drain`, run individually each time), plus the full `request_timeout` module (8/8). `cargo fmt --all -- --check` clean.

---

PR #2528, opened 2026-09-05
🛡 Warden: fold the ambient tenant into every #[cached] key (cross-tenant read, #[cached] × tenant_scoped)

madmax983:
## 🎯 Surface

`autumn-macros::cached` — `#[cached]`'s generated cache-key construction. Entry point: any HTTP route (or any other code path) that calls a `#[cached]` function memoizing a `tenant_scoped` `#[repository]` read.

## 🕵️ Threat model

Against an app that follows Autumn's own documented patterns — turns on multi-tenancy (`[tenancy] enabled = true`) and memoizes a `tenant_scoped` repository read with `#[cached]`, exactly as the framework's own SaaS starter does in `examples/saas/src/repositories.rs` — an attacker who is an ordinary authenticated principal of tenant B can obtain **tenant A's cached response** from any such cached read, for the remainder of the entry's TTL, simply by calling the same route/function with the same *non-tenant* arguments tenant A used.

The app author did nothing the documentation told them not to do: the SaaS starter's own `cached_project_count` is the *correct* pattern (thread `tenant_id` explicitly and name it in `key(tenant_id)`), but nothing in the macro, the build-time cache-coherence gate (`autumn cache audit`), or `autumn routes audit` detects or prevents its omission — and every *other* `tenant_scoped` operation in Autumn resolves the tenant from the ambient `CURRENT_TENANT` task-local automatically, with no explicit parameter required. A developer who caches a `tenant_scoped` read keyed on some other legitimate, real parameter (a page number, an export format, a filter) instead of the tenant has followed the framework's own ambient-scoping idiom everywhere else — and been bitten by the one place it silently doesn't apply.

## 🧪 Reproduction

Test: `autumn/tests/integration/cached_tenant_scope.rs::cached_tenant_scoped_read_leaks_across_tenants`

Seeds two tenants' sentinel rows in a `tenant_scoped` model, mirrors the SaaS starter's `cached_project_count` shape but keys the cached read on a legitimate non-tenant parameter (`format`), and drives it through a real HTTP route (`GET /widgets/summary`) via `TestApp` + header-based tenancy — same convention as `docs/security/2026-09-02-idempotency-tenant-scope/`.

The committed test is `#[ignore = "requires Docker (testcontainers)"]` per `CLAUDE.md`'s Integration Test Layout Guidelines, and is picked up automatically by CI's Docker sweep — no workflow edit. This sandbox had no Docker daemon, so the red/green runs in the ledger were captured with an identical scratch harness pointed at a natively-installed PostgreSQL 16 service instead of a testcontainer (only the pool's connection source differs); the fix was verified reverted-and-restored (`git stash`) against the same running Postgres, so the two runs differ by exactly the `cached.rs` change.

Failure output on trunk (fix reverted) — `docs/security/2026-09-05-cached-tenant-key/trunk-failure.txt`:
```
thread 'cached_tenant_scoped_read_leaks_across_tenants' panicked at ...:170:5:
tenant B received tenant A's cached #[cached] response body: "csv:tenant-a-sentinel"
(the cache key never carried the resolved tenant)
test cached_tenant_scoped_read_leaks_across_tenants ... FAILED
```

## 🔎 Root cause

`autumn-macros/src/cached.rs`'s `generate_cache_body` built the key as `make_cache_key(#id_expr, #key_args)`, where `#key_args` comes exclusively from the function's own explicit parameters (`key_tuple`, ~line 489) — every parameter by default, or exactly the parameters named in `key(...)`. It never read `autumn_web::tenancy::CURRENT_TENANT`, the task-local every `tenant_scoped` repository finder filters by (`autumn-macros/src/repository.rs` ~line 1913). The build-time cache-coherence gate (`autumn-cli/src/cache_audit.rs`) proves only *staleness* edges (declared/derived model dependencies vs. a repository's `invalidates(...)`) — it has no notion of tenant-key correctness, so a `#[cached(reads(TenantScopedModel))]` function with no tenant-identifying key parameter compiles, passes `autumn cache audit`, and runs.

## 🩹 Fix

`generate_cache_body` now reads `CURRENT_TENANT` unconditionally and folds it into the key alongside whatever `key(...)` already names:

```rust
let __autumn_tenant_key_component: Option<String> =
    CURRENT_TENANT.try_with(Clone::clone).ok().flatten();
let __autumn_key = make_cache_key(#id_expr, &(__autumn_tenant_key_component, #key_args));
```

Fixed at the enforcing layer — the macro that generates every `#[cached]` function's key — not at any one call site, so every downstream app gets the fix on upgrade with no code change. `CURRENT_TENANT` resolves to `None` for apps without tenancy enabled and for calls outside a request (a background job, a scheduled sweep), so those keys are unchanged.

## ✅ Verification

- Reproduction test: FAILED before the fix, PASSED after.
- `cargo test -p autumn-macros cached::` — all 40 existing `#[cached]` macro-expansion unit tests pass unchanged.
- `cargo test -p autumn-web --test integration_tests --features "db,cache-moka,test-support" cache_coherence` — 16 passed, 0 failed (existing cache-coherence suite unaffected; new test module compiles cleanly into the 1934-test consolidated binary).
- `cargo clippy -p autumn-macros --lib -- -D warnings` — clean (one pre-existing, unrelated `unknown_lints` warning from a clippy/lint-name version mismatch).
- `cargo fmt --all` — no changes.
- Re-attack: confirmed a same-tenant repeat call still hits the cache within the TTL — the fix partitions by tenant, it doesn't disable caching.

## 📡 Blast radius

- Single fix point: every `#[cached]` function shares the same key-construction code path (`generate_cache_body`/`key_tuple`); no other site independently builds a `#[cached]`-style key.
- Covers the `redis`-backed shared cache backend identically — the generated key is built before the `global_cache()` backend selection, so both Moka and Redis backends use the fixed key.
- `autumn cache audit` / cache-coherence gate intentionally left alone (proves a different property — staleness, not tenant isolation); the macro-level fix is strictly stronger since it protects apps that never run the audit.
- `autumn/src/cache/fragment.rs` and `read_through.rs` build keys from caller-supplied strings, not macro-generated argument hashing — out of scope here; an app misusing those directly is the "app held it wrong" shape, not this one.

## 📜 Compatibility

- No macro *input* syntax changed; per `STABILITY.md`, `#[cached]`'s generated code is explicitly not semver-covered.
- Behavior change (called out in `CHANGELOG.md` under `## [Unreleased]` → `### Security`): a `#[cached]` function intentionally serving one shared, cross-tenant value now partitions per resolved tenant too when called from within a tenant's request — a hit-rate regression for that out-of-contract usage, not a correctness change.
- No config default changed; no migration required.

## 🗂 Ledger

`docs/security/2026-09-05-cached-tenant-key/` (README, trunk-failure.txt, after.txt)

---

https://claude.ai/code/session_01TBKpeGM5tkbSnnhhYNVKVq

---
_Generated by [Claude Code](https://claude.ai/code/session_01TBKpeGM5tkbSnnhhYNVKVq)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3215,6 +3215,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   function that quietly delegated to its borrowed twin would fail even though
   its output is correct.
 
+### Fixed
+
+- **macros: `#[throttle]`/`#[secured]`/`#[step_up]` above the route macro no
+  longer fail to compile, or silently drop the OpenAPI response schema, once
+  stacked (#1668 regression):** #1668 moved these three guards' runtime
+  checks out of the handler body into each guard's own handler-unique
+  `FromRequestParts` gate — `struct` + `impl`, emitted as a sibling item
+  ahead of the (still single) handler function. Every route macro's own
+  `item` parser (`parse::parse_async_handler`, plus the three guard macros'
+  own parsers, needed when one guard expands above another) only ever
+  accepted a lone `ItemFn`, so a guard macro receiving that two-item output
+  as `item` — any of `#[throttle]`/`#[secured]`/`#[step_up]`/`#[route]`
+  written *above* another guard already using the #1668 gate shape — hit a
+  spurious `route macros can only be applied to functions` compile error.
+  `parse::parse_async_handler_with_preamble` now accepts zero or more
+  leading item definitions ahead of the trailing function, returning them
+  separately so the route/guard macro re-emits that preamble (the gate type
+  the function's new first parameter names) verbatim instead of choking on
+  it.
+
+  Fixing the parse gap surfaced a second, previously-masked bug: with
+  parsing no longer failing first, `#[throttle]`/`#[step_up]` above
+  `#[route]` still resolved to no OpenAPI response schema, because
+  `api_doc::infer_response_body`'s `__autumn_inner`-binding recovery (#1677)
+  only trusts that binding when one of `RESPONSE_REWRITING_GUARD_MARKERS`'s
+  marker consts sits earlier in the *same* handler-body block — and #1668
+  moved `#[throttle]`'s/`#[step_up]`'s marker consts into their gate's
+  `impl` block, out of the body, while `#[secured]`'s marker consts stayed
+  in-body for exactly this reason. Both guards now also leave a dead-code
+  copy of their marker const in the handler body (mirroring `#[secured]`'s
+  existing `role_scope_consts` pattern), restoring the invariant
+  `infer_response_body` relies on — including through stacked guards, where
+  only the innermost binding carries the handler's real return type.
+
+  Regression-proven at both the macro-expansion level (`route.rs`'s
+  existing `route_macro_infers_response_schema_when_throttle_expands_first`
+  / `..._when_step_up_expands_first` / `..._under_stacked_guards_above_route`
+  tests, which predate this fix but never previously reached the code path
+  they exercise) and end to end: `cargo test -p autumn-macros --lib` (1073
+  passed), `cargo fmt --all -- --check`, `cargo clippy -p autumn-macros
+  --all-targets -- -D warnings` all clean.
+
 ## [0.7.0] - 2026-08-23
 
 For a narrative tour of this release, see the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1601,6 +1601,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   internal-only receiver during development) will see those deliveries start
   failing after upgrade — this is the intended effect of closing the gap. See
   `docs/security/2026-09-03-webhook-ssrf/`.
+- **`#[cached]`'s generated cache key now folds in the ambient resolved
+  tenant:** the key was built exclusively from the function's own explicit
+  arguments (every parameter by default, or exactly the parameters named in
+  `key(...)`) and never consulted the `CURRENT_TENANT` task-local a
+  `tenant_scoped` repository read filters by. An app that turned on Autumn's
+  multi-tenancy (`[tenancy] enabled = true`) and cached a `tenant_scoped`
+  read keyed on any parameter *other* than the tenant itself — a page, an
+  export format, a filter, anything but the literal `tenant_id` the SaaS
+  starter's own `cached_project_count` goes out of its way to thread through
+  `key(tenant_id)` — shared one cache slot across every tenant that called it
+  with the same non-tenant arguments: tenant B received **tenant A's cached
+  response** for the remainder of the entry's TTL. Nothing in the macro, the
+  build-time cache-coherence gate (`autumn cache audit`), or `autumn routes
+  audit` detected or prevented the omission, and Autumn's own tenancy idiom
+  never requires threading `tenant_id` through a function signature for any
+  *other* tenant-scoped operation (`tenant_scoped` finders resolve it from
+  `CURRENT_TENANT` automatically) — so the omission was an easy, natural
+  mistake, not a documented misuse. The generated wrapper now reads
+  `CURRENT_TENANT` (when tenancy is enabled and a tenant has been resolved)
+  and folds it into the key unconditionally, in addition to whatever `key(...)`
+  already names. Apps without tenancy enabled, or calling a `#[cached]`
+  function outside a request (a background job, a scheduled sweep), compute
+  the same key as before — `CURRENT_TENANT` resolves to `None` in both cases.
+  **Compatibility note:** a `#[cached]` function that intentionally serves one
+  shared, cross-tenant value (a genuinely global computation, not a
+  `tenant_scoped` read) now partitions its cache per resolved tenant too when
+  called from within a tenant's request — a harmless drop in hit rate, not a
+  correctness change, since the computed value does not vary by tenant. See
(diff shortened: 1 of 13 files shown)


madmax983:
`Test (ubuntu-latest)` failed on commit `c1bb30c` with 4 `autumn-macros` unit-test failures (`route::tests::route_macro_infers_response_schema_when_{secured,throttle,step_up}_expands_first`, `route_macro_infers_response_schema_under_stacked_guards_above_route`), all with the same panic: `compile_error! { "route macros can only be applied to functions" }`.

This is not caused by this PR — it's the pre-existing #1668 guard-stacking regression already red on `trunk-dev` at this PR's base commit (`66106a5`), tracked by the already-open fix `d6ed1c8` ("Fix #1668 guard-stacking regression: route macros can only be applied to functions") on branch `fix/route-macro-guard-stacking-1668-regression`. This PR never touches `route.rs`/`secured.rs`/`step_up.rs`/`throttle.rs`/`parse.rs`, so it inherited the failure from the base rather than introducing it.

I ported that fix into this PR (cherry-picked `d6ed1c8`, pushed as `d30cdf5`) so this PR is green independent of when the other fix lands — it no-ops harmlessly once `trunk-dev` carries it too. Verified locally: `cargo test -p autumn-macros --lib` — 1073 passed, 0 failed (up from 4 failed before the cherry-pick).

---

madmax983:
`Test (ubuntu-latest)` failed on `d30cdf5` in `autumn/tests/cached_global_backend.rs`: `cached_fn_reads_from_global_on_hit` and `cached_fn_writes_to_global_on_miss` both failed.

This one **is** this PR's — the test hand-computes the exact cache key `#[cached]`'s generated code produces (`make_cache_key(fn_name, &(args,))`) to pre-seed/inspect the global cache backend directly. Since the fix now folds the ambient tenant component in ahead of the named key parameters, the key shape changed to `(Option<String>, &(args,))` and the test's hardcoded key no longer matched what `double_a`/`double_b` actually compute.

Fixed in `0ef6975`: updated both hand-computed keys to `&(Option::<String>::None, &(args,))` — `None` because these plain functions run outside any tenancy-resolved request. The behavior under test (reading/writing through a registered global cache backend) is unaffected; only the internal key shape the test white-box-asserts changed, which is exactly what this PR's fix does. Verified locally: `cargo test -p autumn-web --test cached_global_backend --features "cache-moka,test-support"` (3 passed) and `cargo test -p autumn-web --lib --features "cache-moka,test-support"` (5559 passed, 0 failed).

Also swept the other `make_cache_key` call sites in the repo (`examples/bookmarks-distributed/tests/cache_cross_replica.rs`, `autumn/tests/integration/cache_coherence.rs`) — neither calls the real macro-generated code path (one tests raw Redis replica consistency under an arbitrary shared key, the other tests only the namespace-prefix format), so neither needed a change.

---

madmax983:
Merge conflict resolved (`trunk-dev` had merged its own, more complete fix for the #1668 guard-stacking regression — including `authorize.rs` changes my earlier cherry-pick didn't have — superseding the commit I'd ported in). Merged `trunk-dev` into this branch and took its version of `route.rs`/`secured.rs`/`step_up.rs`/`throttle.rs`/`parse.rs`/`authorize.rs` wholesale, dropping my now-redundant cherry-pick. `cached.rs` (this PR's actual fix) was untouched by trunk-dev, so no further changes needed there.

Verified locally post-merge: `cargo test -p autumn-macros --lib` — 1075 passed, 0 failed; `cargo test -p autumn-web --lib --features "cache-moka,test-support"` — 5559 passed, 0 failed. Pushed as `2e4cec0`.

---

madmax983:
Another merge conflict, this time in `autumn-macros/src/parse.rs` — `trunk-dev` landed yet another iteration of the same #1668/#2516 guard-stacking fix family (this round extending it to `#[static_get]`/`#[ws]`, renaming the shared helper). Same resolution as before: merged `trunk-dev` in and took its version of `parse.rs` wholesale (the only file with a real conflict; everything else auto-merged cleanly) since this PR has no independent changes there. `cached.rs` — this PR's actual fix — was untouched.

Verified locally post-merge: `cargo test -p autumn-macros --lib` — 1142 passed, 0 failed; `cargo test -p autumn-web --lib --features "cache-moka,test-support"` — 5631 passed, 0 failed; `cargo test -p autumn-web --test cached_global_backend --features "cache-moka,test-support"` — 3 passed. Pushed as `f137034`.

---

madmax983:
`Test (macos-latest)` (and the `Test suite` gate it feeds) failed on `f137034` in `integration::cache_stampede::swr_serves_stale_and_refreshes_in_background`: `"background refresh must publish the new value after it finishes computing"`.

This is not this PR's — `autumn/tests/integration/cache_stampede.rs` is untouched by this PR (which only edits `autumn-macros/src/cached.rs`, the tenant-key fix, plus the reproduction test and unrelated merge-conflict resolutions from `trunk-dev`'s own churn). The failing test asserts a background stale-while-revalidate refresh has published its new value by the time the test polls for it — a wall-clock timing assertion, and this repo's own CI-health tooling (`docs/plans/2026-09-04-semaphore-ci-health-census.md`) already documents a macOS-only timing-flake cluster in exactly this shape.

Re-ran the failed jobs once (`gh`/API `rerun_failed_jobs` on run `34068587012`) to confirm before treating it as a flake, per the drive-to-green rules — watching for the result.

---

madmax983:
Re-run confirms the flake: `Test (macos-latest)` and `Test suite` both passed on attempt 2 with no code changes. All 81 checks are now green on `f137034`.

---

